PR #90, opened 2017-06-30
Add splash page

### LINKED ISSUE #75: Add a nice landing page when you first start Lucky

paulcsmith:
Closes #75 

screen shot 2017-06-30 at 3 42 42 pm

Very simple at first. Might have a few links with instructions later.

All CSS and images should be available on the page so there is NOTHING
to delete, and nothing to serve with webpack. The page should just work.

Lato is loaded over the web, but it falls back to a sans-serif font that
works fine if you aren't connected.


src/lucky_web/welcome_page.cr
```diff
@@ -0,0 +1,88 @@
...
+    @view << "<style>"
+    yield
+    @view << "</style>"
+  end
+
+  def raw(string)
+    @view << string
+  end
+
+  def lucky_logo_data_uri
```
paulcsmith: Should use an SVG, but this will work for now

PR #141, opened 2017-10-23
Fix pages breaking `lucky` command

BlakeWilliams:
When trying to run `lucky` an error was thrown since `Tasks::NewPage`
and `Tasks::IndexPage` were trying to inherit from `LuckyWeb::Page`
which is a model and not a class.

This resolves the issue by replacing inheritance with `include
LuckyWeb::Page`

I'm assuming we want to `include` here instead of turning `LuckyWeb::Page` into a class.


BlakeWilliams:
`crystal -v` output if that's helpful

`Crystal 0.23.1 (2017-10-12) LLVM 4.0.1`

PR #142, opened 2017-10-23
Rename p tag to para to prevent method conflict

BlakeWilliams:
https://github.com/luckyframework/web/issues/128

### DIFF
--- a/src/lucky_web/tags/base_tags.cr
+++ b/src/lucky_web/tags/base_tags.cr
@@ -36,11 +36,11 @@ module LuckyWeb::BaseTags
   end
 
   {% for tag in TAGS %}
-    generate_tag_methods({{tag.id}}, {{tag.id}})
+    generate_tag_methods(method_name: {{tag.id}}, tag: {{tag.id}})
   {% end %}
 
   {% for name, tag in RENAMED_TAGS %}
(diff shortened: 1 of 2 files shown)


BlakeWilliams:
I went ahead and followed the same pattern as `TAGS`. My first pass didn't but this ended up being more clear since it follows the existing pattern and also sets us up for more renamed tags if necessary.

src/lucky_web/tags/base_tags.cr
```diff
@@ -35,7 +36,11 @@ module LuckyWeb::BaseTags
   end
 
   {% for tag in TAGS %}
-    generate_tag_methods({{tag.id}})
+    generate_tag_methods({{tag.id}}, {{tag.id}})
+  {% end %}
+
+  {% for name, tag in RENAMED_TAGS %}
+    generate_tag_methods({{name}}, {{tag}})
```
paulcsmith: Could you use named args here and above so it's a bit more clear

```crystal
generate_tag_methods(method_name: {{ name }}, tag: {{ tag }})
```

BlakeWilliams: That's a lot better, good call!

paulcsmith:
Thanks man!

PR #157, opened 2017-10-27
Show helpful error message when using `p`

### LINKED ISSUE #143: Add `p` method to `LuckyWeb::Page` that directs the user to use `pp` or `para`
There isn't a good reason to use `p` in a web page other than to debug something. So we should add a `p` macro to `LuckyWeb::Page` that accepts any number of arguments and let's the user know they should use `para` if they want a paragraph tag and `pp` if they want to debug. `pp` is better anyway because it also outputs the code that it's printing to standard out. It's super handy!

It would look something like this

```crystal
def p(_arg, **args)
  {% raise <<-ERROR
    `p` is not available on Lucky pages. This is because it's not clear whether you want to print something out or use a `p` HTML tag.
    
    Instead try:
      * The `para` method if you want to use an HTML paragraph.
      * The `pp` method to pretty print information for debugging.
    ERROR 
  %}
end
```

5minpause:
There isn't a good reason to use `p` in a web page other than to debug something. So we should add a p macro to `LuckyWeb::Page` that accepts any number of arguments and let's the user know they should use para if they want a paragraph tag and pp if they want to debug. pp is better anyway because it also outputs the code that it's printing to standard out. It's super handy!

Fixes #143

screen shot 2017-10-27 at 16 33 29

(I used the `BasePage` to verify locally that it works, like this:
```
  def render
    p "test"
  end
```
since I yet don't know how to build my own routes/controllers/view things. and excuse the Rails vocabular… that's what I am coming from…)

### DIFF
--- a/src/lucky_web/page.cr
+++ b/src/lucky_web/page.cr
@@ -39,4 +39,14 @@ module LuckyWeb::Page
 
     generate_initializer
   end
+
+  macro p(_arg, **args)
+    {% raise <<-ERROR
+      `p` is not available on Lucky pages. This is because it's not clear whether you want to print something out or use a `p` HTML tag.
+      
+      Instead try:
+        * The `para` method if you want to use an HTML paragraph.
+        * The `pp` method to pretty print information for debugging.
+      ERROR %}
+  end
 end


src/lucky_web/page.cr
```diff
@@ -39,4 +39,16 @@ module LuckyWeb::Page
 
     generate_initializer
   end
+
+  macro p(_arg, **args)
+    def p(_arg, **args)
```
paulcsmith: I *think* this will still work if you remove the `def p` and just have the macro. Could you give that a shot?

5minpause: Sure!

5minpause: @paulcsmith It did work. I don't know why, since https://crystal-lang.org/docs/syntax_and_semantics/macros.html tells me to include the `def p` but what do *I* know, right? 💃

paulcsmith:
Thank you :D

PR #158, opened 2017-10-27
Add infos how to run the tests

5minpause:
If you are new to crystal it might confuse you how to run specs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@
 
 Lucky is still in it's early stages. A guide will be written once the API is more stabilized.
 
+## Testing
+
+You need to make sure to install the Crystal dependencies.
+1. Run `crystal deps`
+1. Run `crystal spec` from the project root.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,12 @@
 
 Lucky is still in it's early stages. A guide will be written once the API is more stabilized.
 
+## Testing
+
+You need to make sure to install the Crystal dependencies.
+1. Run `crystal deps`
+2. Run `crystal spec` from the project root.
```
paulcsmith: Can you change this `2` to `1`. This is a cool feature in markdown where it will correctly number them when it's rendered. This way if we add more steps and need to change the order, the number won't be out of sync

5minpause: Yeah, I noticed this in the opening section. Great idea.

paulcsmith:
Thanks!

Issue #140, opened 2017-10-21
Add type declaration to HTML tags


paulcsmith:
https://github.com/luckyframework/web/blob/master/src/lucky_web/tags/base_tags.cr#L9

That way the errors are a bit better. Right now the error comes from the `text` method instead of the tag that called it. So it's a bit harder to realize what went wrong

tgturner:
I would like to take a look at this one!

PR #160, opened 2017-10-27
Add type declaration to HTML tags

tgturner:
Addresses PR #140 

I'm new to Crystal, so please let me know if you see any issues.

### DIFF
--- a/src/lucky_web/tags/base_tags.cr
+++ b/src/lucky_web/tags/base_tags.cr
@@ -7,7 +7,7 @@ module LuckyWeb::BaseTags
   @view = IO::Memory.new
 
   macro generate_tag_methods(method_name, tag)
-    def {{method_name.id}}(content = "", options = EMPTY_HTML_ATTRS, **other_options)
+    def {{method_name.id}}(content : String? = "", options = EMPTY_HTML_ATTRS, **other_options)
       merged_options = merge_options(other_options, options)
       {{method_name.id}}(merged_options) do
         text content


src/lucky_web/tags/base_tags.cr
```diff
@@ -7,7 +7,7 @@ module LuckyWeb::BaseTags
   @view = IO::Memory.new
 
   macro generate_tag_methods(method_name, tag)
-    def {{method_name.id}}(content = "", options = EMPTY_HTML_ATTRS, **other_options)
+    def {{method_name.id}}(content : String? = "", options = EMPTY_HTML_ATTRS, **other_options)
```
paulcsmith: Since this should always accept a String and not `nil` it should be `String`.

tgturner: I tried that first but was getting a test failure here.

```
Error in line 16: while requiring "./spec/lucky_web/input_helpers_spec.cr"

in spec/lucky_web/input_helpers_spec.cr:140: instantiating 'TestPage#textarea(LuckyRecord::AllowedField(String))'

    view.textarea(form.first_name).to_s.should contain <<-HTML
         ^~~~~~~~

in src/lucky_web/tags/input_helpers.cr:7: wrong number of arguments for 'TestPage#textarea' (given 2, expected 0..1)
Overloads are:
 - LuckyWeb::InputHelpers#textarea(field : LuckyRecord::AllowedField, **html_options)
 - LuckyWeb::BaseTags#textarea(content : String)
 - LuckyWeb::BaseTags#textarea(options = EMPTY_HTML_ATTRS, **other_options, &block)

    textarea field.param.to_s, merge_options(html_options, {
    ^~~~~~~~
```

My understanding here is that there are times, like in the `textarea` helper, where Lucky is passing a `nil` value as the first parameter, which allows the default value to be assigned. Is that a behavior you would like to continue?

tgturner: Also, to be fair, my understanding of Crystal is limited at this point, so there is a good chance I'm missing something entirely.

paulcsmith: Ah I see. That makes sense, thanks for the detailed error message. 

I think the type should be `String` and the textarea helper should pass in a `""` instead of `nil`. Does that make sense?

tgturner: Yep! I'll see what I can do. Thanks for all of the help Paul.

tgturner: @paulcsmith I've looked into this issue a bit more and I'm not quite able to figure out what the problem is with adding the String type declaration. Even once I comment out the two `textarea` tests, I'm seeing other failures with the same `wrong number of arguments for` compiler error. Any suggestions on moving forward?

paulcsmith: @tgturner I pulled this down and made the change to see what you mean. I think what you have here is probably best then because you're right that there is no easy way to require the String. I think what you have here is good and a definite improvement over what's there. Thank you!

tgturner: Thanks Paul!

PR #169, opened 2017-10-29
Add singular resources

### LINKED ISSUE #102: Add singular resources
```crystal
action singular: true
```

Steps to do this one:

1. Add a spec for it. You can probably add it to `spec/lucky_web/action_spec.cr`. Note: Feel free to split these up if you want. That spec is getting pretty big
1. Add an option for `singular: true` to the `action` macro https://github.com/luckyframework/web/blob/ea7b5ac0c09a8ae779f2d2b1d97b10980ecf8c0c/src/lucky_web/routeable.cr#L20. 
1. Pass whether it is singular to not to the route inference code https://github.com/luckyframework/web/blob/ea7b5ac0c09a8ae779f2d2b1d97b10980ecf8c0c/src/lucky_web/routeable.cr#L33
1.  Write code to handle singular routes in the route inference class in `src/run_macros/infer_route.cr`. This class is what's used to figure out the path and method. You'll want to check for singular routes and omit `:id` if it's singular.

### LINKED ISSUE #103: Add singular nested action
`nested_action singular: true`

Similar steps as shown in https://github.com/luckyframework/web/issues/102 but for nested_actions. The code may be almost identical so whichever one is done second may be pretty simple

citizen428:
Closes #102 
Closes #103.

I took the liberty to also refactor some minor things in `InferRoute` and the specs (adding a `RoutesHelper` to share between the different spec files), I hope you don't mind.

### DIFF
--- a/spec/lucky_web/singular_action_spec.cr
+++ b/spec/lucky_web/singular_action_spec.cr
@@ -1,0 +1,52 @@
+require "../../spec_helper"
+
+include RoutesHelper
+
+class Singular::New < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+class Singular::Edit < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+class Singular::Show < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+class Singular::Delete < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+class Singular::Update < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+class Singular::Create < LuckyWeb::Action
+  action singular: true do
+    render_text "test"
+  end
+end
+
+describe LuckyWeb::Action do
+  describe "singular routing" do
+    it "adds singular routes to the router" do
+      assert_route_added? LuckyWeb::Route.new :get, "/singular/new", Singular::New
+      assert_route_added? LuckyWeb::Route.new :get, "/singular/edit", Singular::Edit
+      assert_route_added? LuckyWeb::Route.new :get, "/singular", Singular::Show
+      assert_route_added? LuckyWeb::Route.new :delete, "/singular", Singular::Delete
+      assert_route_added? LuckyWeb::Route.new :put, "/singular", Singular::Update
+      assert_route_added? LuckyWeb::Route.new :post, "/singular", Singular::Create
+    end
+  end
+end
--- a/src/lucky_web/routeable.cr
+++ b/src/lucky_web/routeable.cr
@@ -23,28 +23,28 @@ module LuckyWeb::Routeable
     end
   {% end %}
 
-  macro nested_action
-    infer_nested_route
+  macro nested_action(singular = false)
+    infer_nested_route(singular: {{singular}})
 
     def call : LuckyWeb::Response
       {{yield}}
     end
   end
 
-  macro action
-    infer_route
+  macro action(singular = false)
+    infer_route(singular: {{singular}})
 
     def call : LuckyWeb::Response
       {{yield}}
     end
   end
 
-  macro infer_nested_route
-    infer_route(true)
+  macro infer_nested_route(singular = false)
+    infer_route(has_parent: true, singular: singular)
   end
 
-  macro infer_route(has_parent = false)
-    {{ run "../run_macros/infer_route", has_parent, @type.name }}
+  macro infer_route(has_parent = false, singular = false)
+    {{ run "../run_macros/infer_route", has_parent, @type.name, singular }}
   end
 
   macro add_route(method, path, action)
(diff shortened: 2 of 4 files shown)


citizen428:
@paulcsmith This is ready for review now. 😃

spec/lucky_web/singular_action_spec.cr
```diff
@@ -0,0 +1,52 @@
...
+    it "adds singular routes to the router" do
+      assert_route_added? LuckyWeb::Route.new :get, "/singular/new", Singular::New
+      assert_route_added? LuckyWeb::Route.new :get, "/singular/edit", Singular::Edit
+      assert_route_added? LuckyWeb::Route.new :get, "/singular", Singular::Show
+      assert_route_added? LuckyWeb::Route.new :delete, "/singular", Singular::Delete
+      assert_route_added? LuckyWeb::Route.new :put, "/singular", Singular::Update
+      assert_route_added? LuckyWeb::Route.new :post, "/singular", Singular::Create
+    end
+  end
+end
```
paulcsmith: This spec looks perfect. Nice extraction of the `RoutesHelper` 👍

src/lucky_web/routeable.cr
```diff
@@ -23,28 +23,28 @@ module LuckyWeb::Routeable
...
 
     def call : LuckyWeb::Response
       {{yield}}
     end
   end
 
-  macro action
-    infer_route
+  macro action(singular = false)
+    infer_route(singular: {{singular}})
```
paulcsmith: Could you add some spaces inside the `{{  }}`? I haven't been consistent, but I'd like to do that from now on so it's easier to spot where interpolation is being done in macros

src/run_macros/infer_route.cr
```diff
@@ -51,32 +52,67 @@ class InferRoute
...
+    when "index", "create"
       [resource]
-    elsif action_name == "new"
+    when "new"
       [resource, "new"]
-    elsif action_name == "edit"
+    when "edit"
       [resource, ":id", "edit"]
-    elsif ["show", "update", "delete"].includes? action_name
+    when "show", "update", "delete"
```
paulcsmith: Very nice change 👍

src/run_macros/infer_route.cr
```diff
@@ -51,32 +52,67 @@ class InferRoute
...
-          #{action_class_name} (missing a known resourceful action)
-
-        Expected something like:
-          Users::Index # Index, Show, New, Create, Edit, Update, or Delete
+        Reason: Singular routes do not support an index action"
         ERROR
       )
 
-      raise "Problem inferring route"
+      raise "Invalid index action for singular resource"
```
paulcsmith: One small thing, could you capitalize `Index`?

src/run_macros/infer_route.cr
```diff
@@ -31,7 +31,8 @@ class InferRoute
   end
 
   private def all_pieces
-    (namespace_pieces + parent_resource_pieces + resource_pieces).reject(&.== "")
+    resource_pieces = singular? ? singular_resource_pieces : plural_resource_pieces
```
paulcsmith: Instead of doing this inline, could you extract a `resource_pieces` private method that uses an `if` instead of ternary? I think that might make it a bit easier to read and change in the future

Something like this
```crystal
private def resource_pieces
  if singular?
    singular_pieces
  else
    plural_resources_pieces
  end
end
```

src/lucky_web/routeable.cr
```diff
@@ -23,28 +23,28 @@ module LuckyWeb::Routeable
...
-  macro infer_nested_route
-    infer_route(true)
+  macro infer_nested_route(singular = false)
+    infer_route(has_parent: true, singular: singular)
   end
 
-  macro infer_route(has_parent = false)
-    {{ run "../run_macros/infer_route", has_parent, @type.name }}
+  macro infer_route(has_parent = false, singular = false)
+    {{ run "../run_macros/infer_route", has_parent, @type.name, singular }}
```
paulcsmith: I didn't realize this when I first wrote it, but maybe the action class should go first, then has_parent, then singular. So the main focus (the action) is first, and the options (has_parent, singular) are last. What do you think?

src/lucky_web/routeable.cr
citizen428: Sure thing! I actually prefer the spaces, but most of the other places I checked didn't have them. I changed all the occurrences in `routeable.cr`, hope that's ok.

src/run_macros/infer_route.cr
citizen428: Funny, I had this first, then inlined it. Changed!

src/run_macros/infer_route.cr
citizen428: Sure, makes sense.

src/lucky_web/routeable.cr
citizen428: I thought about that yesterday while looking at the code. It makes a lot of sense, I changed it now.

paulcsmith:
Thank you!

PR #179, opened 2017-11-01
Change routeable's "route" to "with"

### LINKED ISSUE #170: Change Action.route to Action.with
So you call route helpers like this `link "edit", to: MyAction::Edit.with(id: something.id)`

Another alternative is `MyAction::Edit(id: something.id)` which is sort, but maybe a bit magical? I'm not sure about it, but it is nice that it's so short. EDIT: This won't work: https://github.com/luckyframework/web/issues/170#issuecomment-340450926

See discussion in Gitter https://gitter.im/luckyframework/Lobby?at=59f7126fd6c36fca31b4223c

From @edwardloveall
> I was thinking about routing last night, and it feels a little weird to me that it's link "foo", to: SomeClass OR link "foo", to: SomeClass.route(with_params)
from an API, it's seems inconsistent that I'm linking to either a class, or a or whatever route returns
i see how it works on link internally, but I wonder if it would make more sense for the params version to look like link "foo", to: SomeAction, params: with_params
.route was a bit of a head scratcher when I first saw it
probably because it's not clear to me that the route method would take params (as opposed to a route, or location)
actually, changing route to with_params or params would also solve this

My response:
> I was thinking it was confusing too but couldn’t come up with a good name. I think your suggestion to rename is great. I think params might be confusing because it seems like it would return a params object or something. with_params is super clear but a bit long. Maybe with would be a good middle ground? link to: Topics::Show.with(topic) not quite as clear as with_params, but at least it’s not actively confusing people, and this way it’s not so much to type once you do learn it. Thoughts?

edwardloveall:
Fixes #170 

When routing to a particular action, an action class is specified. Some
actions also need parameters passed to them. That used to look like:

```crystal

link "edit", to: MyResource::Index

link "edit", to: MyResource::Edit.route(some_params)
```

The `route` method was a confusing here though. You either pass an
action class or the result of it's route method? Too weird. Several
options were discussed:

* `link "edit", to: MyResource::Edit(id: something.id)` - Won't work
syntactically
* `link "edit", to: MyResource::Edit, id: something.id` - confusing
because it could be an html attribute or params
* `link "edit", to: MyResource::Edit.with_params(id: something.id)` - Too
long given the amount it needs to be typed while using lucky
* `link "edit", to: MyResource::Edit.for(id: something.id)` - Possible.
Some debate was had on if that's the best preposition

In the end, `with` was chosen for its brevity, ease of understanding in
context, and it's ability to convey what needs to be passed to it.

It is a little weird that if you want an action's route, you call `MyResource::Index.with` which makes no sense. I looked into aliasing the method back to route for this case, but it looks like that's [not available](https://github.com/crystal-lang/crystal/issues/1179#issuecomment-131267190) in Crystal anymore.

---

I also ran into a problem in src/lucky_web/redirectable.cr trying to remove the `with` on line 7: https://github.com/edwardloveall/web/blob/735c217a1da407c2c0381bcaccef36a514b40851/src/lucky_web/redirectable.cr#L7

This caused a crash at run time, it compiled okay though. I'm not sure why. Here's the crash if you're curious:

```
Invalid memory access (signal 11) at address 0x7fff5f2a7ff8
[0x10019843b] *CallStack::print_backtrace:Int32 +107
[0x100171e1c] __crystal_sigfault_handler +60
[0x7fff92b75b3a] _sigtramp +26
[0x1002d2655] *RedirectAction@LuckyWeb::Redirectable#redirect<RedirectAction:Class, Int32>:NoReturn +5
[0x1002d2659] *RedirectAction@LuckyWeb::Redirectable#redirect<RedirectAction:Class, Int32>:NoReturn +9 (523934 times)
[0x1002d2645] *RedirectAction@LuckyWeb::Redirectable#redirect:to<RedirectAction:Class>:NoReturn +37
[0x10017a084] ~procProc(Nil)@spec/lucky_web/action_redirect_spec.cr:12 +212
[0x100176e6b] *it<String, String, Int32, Int32, &Proc(Nil)>:(Array(Spec::Result) | Nil) +379
[0x100179f60] ~procProc(Nil)@spec/lucky_web/action_redirect_spec.cr:11 +112
[0x1002154a2] *Spec::RootContext::describe<String, String, Int32, &Proc(Nil)>:Spec::Context+ +354
[0x100179c43] *describe<LuckyWeb::Action:Class, String, Int32, &Proc(Nil)>:Spec::Context+ +51
[0x10015ab9c] __crystal_main +5788
[0x100171cc8] main +40
```

### DIFF
--- a/spec/lucky_web/action_redirect_spec.cr
+++ b/spec/lucky_web/action_redirect_spec.cr
@@ -15,7 +15,7 @@ describe LuckyWeb::Action do
     should_redirect(action, to: "/somewhere", status: 302)
 
     action = RedirectAction.new(build_context, params)
-    action.redirect to: RedirectAction.route
+    action.redirect to: RedirectAction.with
     should_redirect(action, to: RedirectAction.path, status: 302)
 
     action = RedirectAction.new(build_context, params)
--- a/spec/lucky_web/action_redirect_spec.cr
+++ b/spec/lucky_web/action_redirect_spec.cr
@@ -15,7 +15,7 @@ describe LuckyWeb::Action do
     should_redirect(action, to: "/somewhere", status: 302)
 
     action = RedirectAction.new(build_context, params)
-    action.redirect to: RedirectAction.with
+    action.redirect to: RedirectAction.route
     should_redirect(action, to: RedirectAction.path, status: 302)
 
     action = RedirectAction.new(build_context, params)
--- a/spec/lucky_web/button_helpers_spec.cr
+++ b/spec/lucky_web/button_helpers_spec.cr
@@ -15,15 +15,15 @@ private class TestPage
   end
 
   def get_route
-    button "Test", to: ButtonHelpers::Index.route
+    button "Test", to: ButtonHelpers::Index
   end
 
   def non_get_route
-    button "Test", to: ButtonHelpers::Create.route
+    button "Test", to: ButtonHelpers::Create
   end
 
   def non_get_route_with_options
-    button "Test", to: ButtonHelpers::Create.route, something_custom: "foo"
+    button "Test", to: ButtonHelpers::Create, something_custom: "foo"
   end
 
   def string_path
@@ -35,7 +35,7 @@ private class TestPage
   end
 
   def get_route_with_block
-    button to: ButtonHelpers::Index.route do
+    button to: ButtonHelpers::Index do
       text "Hello"
     end
   end
--- a/src/lucky_web/version.cr
+++ b/src/lucky_web/version.cr
@@ -1,3 +1,3 @@
 module LuckyWeb
-  VERSION = "0.4.2"
+  VERSION = "0.4.3"
 end
--- a/src/lucky_web/version.cr
+++ b/src/lucky_web/version.cr
@@ -1,3 +1,3 @@
 module LuckyWeb
-  VERSION = "0.4.3"
+  VERSION = "0.4.2"
 end
--- a/spec/lucky_web/action_spec.cr
+++ b/spec/lucky_web/action_spec.cr
@@ -86,9 +86,9 @@ describe LuckyWeb::Action do
   describe "routing" do
     it "creates URL helpers for the resourceful actions" do
       Tests::Index.path.should eq "/tests"
-      Tests::Index.with.should eq LuckyWeb::RouteHelper.new(:get, "/tests")
+      Tests::Index.route.should eq LuckyWeb::RouteHelper.new(:get, "/tests")
       Tests::New.path.should eq "/tests/new"
-      Tests::New.with.should eq LuckyWeb::RouteHelper.new(:get, "/tests/new")
+      Tests::New.route.should eq LuckyWeb::RouteHelper.new(:get, "/tests/new")
       Tests::Edit.path("test-id").should eq "/tests/test-id/edit"
       Tests::Edit.with("test-id").should eq LuckyWeb::RouteHelper.new(:get, "/tests/test-id/edit")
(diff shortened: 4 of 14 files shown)


src/lucky_web/version.cr
```diff
@@ -1,3 +1,3 @@
 module LuckyWeb
-  VERSION = "0.4.2"
+  VERSION = "0.4.3"
```
edwardloveall: Since this is a breaking change, normally the major version number would change, but since this is pre-release I didn't know what to change this to, if anything. Happy to do whatever.

spec/lucky_web/action_redirect_spec.cr
```diff
@@ -15,7 +15,7 @@ describe LuckyWeb::Action do
     should_redirect(action, to: "/somewhere", status: 302)
 
     action = RedirectAction.new(build_context, params)
-    action.redirect to: RedirectAction.route
+    action.redirect to: RedirectAction.with
```
paulcsmith: Even though there is no built-in `alias_method` in Crystal we could do something like this: https://play.crystal-lang.org/#/r/3076. Because you're right that `with` without params is pretty weird

src/lucky_web/version.cr
paulcsmith: I think you could leave this as-is. I'll bump it to 0.5.0 once a couple more things are merged

paulcsmith:
It makes sense why it crashed without the `with` on https://github.com/edwardloveall/web/blob/735c217a1da407c2c0381bcaccef36a514b40851/src/lucky_web/redirectable.cr#L7, but I'm not sure why it compiled. If I had to guess it's because it got stuck in an infinite loop because it looks that would keep passing the LuckyWeb::Action class to `redirect` over and over

paulcsmith:
Sorry, one more thought. Could you try adding this:

```crystal
# For this to work I think you'll need to generate the `def self.route` method as it is on master
# Then add
def self.with(**args)
  route(**args)
end

# And this should raise if no args are given to `with`
def self.with
  {% raise "Use `route` instead of `with` if the action doesn't need params  %}
end
```

That way people don't think they need to use `Action.with`. They'll get a compile time error that guides them in the right direction instead.

edwardloveall:
To be fair, the compilation was with `crystal spec`. Maybe there are some optimizations or checks the compiler does differently in spec mode.

edwardloveall:
This last commit has a bug in the specs. I outlined it in the [commit message](https://github.com/luckyframework/web/pull/179/commits/740dcbf5a19313b0c2c54cb13a0aa7b18ac93a02)

paulcsmith:
@edwardloveall Oh yeah that was a tricky one. Here's the fixes:

```crystal
    def self.with(**args)
      route(**args)
    end

    # So you can also pass params with just values and no keys
    def self.with(*args)
      route(*args)
    end

    def self.with
      # This needs to be escaped because it's a macro generating a macro
      # If it's not escaped it raises when defining an action
      \{% raise "Use `route` instead of `with` if the action doesn't need params" %}
    end
```

spec/lucky_web/button_helpers_spec.cr
```diff
@@ -15,15 +15,15 @@ private class TestPage
   end
 
   def get_route
-    button "Test", to: ButtonHelpers::Index.route
+    button "Test", to: ButtonHelpers::Index
   end
 
   def non_get_route
-    button "Test", to: ButtonHelpers::Create.route
+    button "Test", to: ButtonHelpers::Create
```
paulcsmith: Nice!

PR #181, opened 2017-11-01
[WIP] Add optional_param macro to actions

### LINKED ISSUE #116: Add optional_param macro to actions
This will make it so you can declare that a query param is optional. This would be used for things like `page` for paginating.

```crystal
class Users::Index < BrowserAction
  optional_param page
end
```

When you declare this it should also created a method for the param

```crystal
def page
  params.get!(:page)
end
```

And it should add an optional argument to the route and path methods that are generated for the action

citizen428:
Closes #116

@paulcsmith This is very WIP, just opening a PR early so you can follow along.

### DIFF
--- a/src/lucky_web/routeable.cr
+++ b/src/lucky_web/routeable.cr
@@ -1,4 +1,20 @@
 module LuckyWeb::Routeable
+  macro included
+    OPTIONAL_PARAMS = [] of Symbol
+
+    macro inherited
+      OPTIONAL_PARAMS = [] of Symbol
+
+      inherit_optional_params
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


paulcsmith:
@citizen428 I'm going to close this up since https://github.com/luckyframework/lucky/pull/285 is being worked on now, but thanks for starting on this. I know you're super busy, so take it easy and we'll take are of this feature in the meantime. Good luck with all the plans coming up!

citizen428:
Yup, sorry for disappearing in the middle of this, real life has a nasty habit of sneaking up on you 😉

PR #182, opened 2017-11-02
Add HTML `select` helper

azbshiri:


src/lucky_web/tags/select_helpers.cr
```diff
@@ -0,0 +1,24 @@
+module LuckyWeb::SelectHelpers
+  alias SelectOption = (String | Int32 | Int64)
+
+  def select(field : LuckyRecord::AllowedField, **html_options)
```
paulcsmith: Could you rename this to `select_input`? I think that will result in less conflicts in the future

spec/lucky_web/select_helpers_spec.cr
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  it "renders select with options" do
+    view.select form.company_id do
+      options_for_select(form.company_id, [{"Volvo", 1}])
+    end.to_s.should contain <<-HTML
+      <select name="company:company_id">
+        <option value="1">Volvo</option>
+      </select>
+    HTML
```
paulcsmith: Could you add a test for when the option is selected? That way we can be sure that functionality works now and in the future

spec/lucky_web/select_helpers_spec.cr
```diff
@@ -0,0 +1,46 @@
...
+describe LuckyWeb::SelectHelpers do
+  it "renders select" do
+    (view.select form.company_id do; end).to_s.should contain <<-HTML
+      <select name="company:company_id"></select>
+    HTML
+  end
+
+  it "renders select with options" do
+    view.select form.company_id do
+      options_for_select(form.company_id, [{"Volvo", 1}])
```
paulcsmith: The problem was that this should be `view.options_for_select`. I think that you could change this spec to not render within a select though and it would work well and be a bit shorter. Something like this:

```crystal
view.options_for_select(form.company_id, [{"Volvo", 1}]).to_s.should contain <<-HTML
<option value="1">Volvo</option>
HTML
```

src/lucky_web/tags/select_helpers.cr
```diff
@@ -0,0 +1,24 @@
...
+    select_tag merge_options(html_options, {"name" => input_name(field)}) do
+      yield
+    end
+  end
+
+  def options_for_select(field : LuckyRecord::AllowedField, select_options : Array(String, SelectOption), **html_options)
+    select_options.each do |select_name, select_value|
+       is_selected = if select_value.to_s == field.value.to_s
```
paulcsmith: Sorry for the barrage of comments, but I just remembered that you should use `field.param` here instead. That's because if a field fails to save, the `param` method will be set, but value will be `nil`.

paulcsmith:
Thanks a ton. This looks great!

azbshiri:
I think we should change `Array(Tuple(String, SelectOption))` to `Tuple(String, SelectOption)` otherwise `options_for_select` only returns one option, IMO.

paulcsmith:
I think you're right that we should change this, but I'm not totally clear how. The return value actually isn't important since when rendering the page you never use the return value. It's only used in the spec. I think the best bet might be to do something more like the https://github.com/luckyframework/web/blob/master/spec/lucky_web/form_helpers_spec.cr where it renders tags in the page instead. But I'm not sure. I think what you have here is fine for now. I plant to go through and clean up the page specs in the future and I'll tackle the select helper one at that point as well

PR #183, opened 2017-11-03
Remove unnecessary return

paulcsmith:
Minor change to the options_for_select method


paulcsmith:
Never mind. I see now that this is to make the test pass by returning the option. It technically works without it on a page though. I'll close this and leave it for now. I'll probably change some of the specs later to make HTML page tests a bit easier

PR #201, opened 2017-11-10
Add checkbox input helper

### LINKED ISSUE #19: Add check box helper
Remember to use hidden field so we get the unchecked value. This will work similar to the `text_input` helpers https://github.com/luckyframework/web/blob/master/src/lucky_web/tags/input_helpers.cr

http://stackoverflow.com/questions/11550941/why-does-the-check-box-form-helper-generate-two-checkboxes-one-hidden

Also: https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/inputs/boolean_input.rb#L93

5minpause:
This is inspired by the Rails and SimpleForm checkbox helpers. There
will always be a hidden field, that transmits the unchecked value.
Otherwise the form would just omit the value, if the checkbox isn't
checked. If the user provides a value for `unchecked_value` we just use
that one. If none given it defaults to "0".

Closes #19

### DIFF
--- a/spec/lucky_web/input_helpers_spec.cr
+++ b/spec/lucky_web/input_helpers_spec.cr
@@ -36,6 +36,20 @@ describe LuckyWeb::InputHelpers do
     HTML
   end
 
+  it "renders checkbox inputs" do
+    view.checkbox(form.first_name, value: "1").to_s.should contain <<-HTML
+    <input type="checkbox" name="user:first_name" value="1"/>
+    HTML
+
+    view.checkbox(form.first_name, value: "1", class: "cool").to_s.should contain <<-HTML
+    <input type="checkbox" name="user:first_name" value="1" class="cool"/>
+    HTML
+
+    view.checkbox(form.first_name, value: "1").to_s.should have_unchecked_value("0")
+
+    view.checkbox(form.first_name, unchecked_value: "not_accepted").to_s.should have_unchecked_value("not_accepted")
+  end
+
   it "renders text inputs" do
     view.text_input(form.first_name).to_s.should contain <<-HTML
     <input type="text" name="user:first_name" value="My name"/>
@@ -153,4 +167,10 @@ end
 
 private def view
   TestPage.new
+end
+
+private def have_unchecked_value(value)
+  contain <<-HTML
+  <input type="hidden" name="user:first_name" value="#{value}" id=""/>
+  HTML
 end
--- a/src/lucky_web/tags/input_helpers.cr
+++ b/src/lucky_web/tags/input_helpers.cr
@@ -9,6 +9,12 @@ module LuckyWeb::InputHelpers
     })
   end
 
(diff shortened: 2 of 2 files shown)


spec/lucky_web/input_helpers_spec.cr
```diff
@@ -36,6 +36,24 @@ describe LuckyWeb::InputHelpers do
...
+  it "renders checkbox inputs" do
+    view.checkbox(form.first_name, value: "1").to_s.should contain <<-HTML
+    <input type="checkbox" name="user:first_name" value="1"/>
+    HTML
+
+    view.checkbox(form.first_name, value: "1", class: "cool").to_s.should contain <<-HTML
+    <input type="checkbox" name="user:first_name" value="1" class="cool"/>
+    HTML
+
+    view.checkbox(form.first_name, value: "1").to_s.should contain <<-HTML
```
paulcsmith: I think extracting a method to explain this matcher might aid in readability. I'm thinking something like this:

```crystal
view.checkbox(form.first_name, value: "1").to_s.should have_unchecked_value("0")

def have_unchecked_value(value)
  contain <<-HTML
  <input type="hidden" name="user:first_name" value="#{value}" id=""/>
  HTML
end
```

What do you think?

src/lucky_web/tags/input_helpers.cr
```diff
@@ -9,6 +9,12 @@ module LuckyWeb::InputHelpers
     })
   end
 
+  def checkbox(field : LuckyRecord::AllowedField, **html_options, unchecked_value : (String | Nil) = nil)
```
paulcsmith: You can do the equivalent of `(String | Nil)` by using `String?`

paulcsmith: Also I wonder if we should also have a `checked_value` option as well?

5minpause: `checked_value` would currently be passed as `value`. This is the same as the other inputs. My thoughts where to have a unified api there for the helpers.

paulcsmith: Oh duh. That makes total sense 👍

5minpause: `String | Nil` => `String?` … the wonders of Crystal. Thanks for pointing that out 👍

spec/lucky_web/input_helpers_spec.cr
5minpause: That makes total sense!

spec/lucky_web/input_helpers_spec.cr
```diff
@@ -154,3 +168,9 @@ end
 private def view
   TestPage.new
 end
+
+def have_unchecked_value(value)
```
paulcsmith: Oops. I forgot the `private` part in my example. Could you mark this is `private` (like on the method above) so it doesn't leak into other specs?

PR #202, opened 2017-11-10
Introduce LuckyWeb::AllowedInTags for HTML tags

### LINKED ISSUE #192: Add module to allow printing other types in HTML text
Right now you can only print strings to HTML. So you have to `to_s` Int32s for example.

Instead, let's do something like this:
* Add a `LuckyWeb::HtmlOutputtable` (hoping to find a better name)
* Change type restriction for HTML tags from `String` to `String | LuckyWeb::HtmlOutputtable`
* Add `LuckyWeb::HtmlOututtable` to `Int32` and `Int64`

Other types could also include the module to be allowed in HTML text, but I think that should be rarely done. Might be helpful for custom types like `Money` or `EmailAddress` though. That way you don't need to `to_s` it first

5minpause:
Until now we only did allow Strings for content in HTML tags. Since
people do want to write Int32s and Int64 (and potentially other types as
well) to HTML, we created a new module called `LuckyWeb::AllowedInTags`.

This type is allowed for HTML tags and will be written to HTML with
`to_s`.

Closes #192

### DIFF
--- a/src/lucky_web/allowed_in_tags.cr
+++ b/src/lucky_web/allowed_in_tags.cr
@@ -1,0 +1,2 @@
+module LuckyWeb::AllowedInTags
+end
--- a/src/lucky_web/tags/base_tags.cr
+++ b/src/lucky_web/tags/base_tags.cr
@@ -14,7 +14,7 @@ module LuckyWeb::BaseTags
       end
     end
 
-    def {{method_name.id}}(content : String)
+    def {{method_name.id}}(content : String | LuckyWeb::AllowedInTags)
       {{method_name.id}}(EMPTY_HTML_ATTRS) do
         text content
       end
@@ -55,8 +55,8 @@ module LuckyWeb::BaseTags
     end
   {% end %}
 
-  def text(content : String)
-    @view << HTML.escape(content)
+  def text(content : String | LuckyWeb::AllowedInTags)
+    @view << HTML.escape(content.to_s)
   end
 
   private def build_tag_attrs(options)
--- a/spec/lucky_web/base_tags_spec.cr
+++ b/spec/lucky_web/base_tags_spec.cr
@@ -7,9 +7,23 @@ private class TestPage
   end
 end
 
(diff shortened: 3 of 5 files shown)


src/lucky_web/support/int32.cr
```diff
@@ -0,0 +1,5 @@
+require "../allowed_in_tags"
+
+struct Int32
```
paulcsmith: This should go in `src/charms/*.cr` That's where all the extensions to build in types go

src/lucky_web/tags/base_tags.cr
```diff
@@ -1,4 +1,6 @@
 module LuckyWeb::BaseTags
+  include LuckyWeb::AllowedInTags
```
paulcsmith: I don't think you need to include this here

src/lucky_web/tags/base_tags.cr
```diff
@@ -55,8 +57,8 @@ module LuckyWeb::BaseTags
     end
   {% end %}
 
-  def text(content : String)
-    @view << HTML.escape(content)
+  def text(content : String | LuckyWeb::AllowedInTags)
```
paulcsmith: Nice! This is exactly what I was thinking 👍

5minpause:
Glad you like it. I made the changes. 😄

src/lucky_web/allowed_in_tags.cr
```diff
@@ -0,0 +1,2 @@
+module LuckyWeb::AllowedInTags
```
paulcsmith: This needs to be included in the Int32 and Int64 structs. That can be done by adding some extensions in `src/charms`

5minpause: Ah thanks. After moving the files to the `/charms` folder I forgot a `git add`. 🕶

paulcsmith:
Perfect! Thanks 🎉

PR #207, opened 2017-11-11
Add better logger

### LINKED ISSUE #206: Make custom logger that allows for optional debug output
Here's how it should look like:

```crystal
puts <<-LOG
GET 200 #{"/".colorize(:green)} (1.3ms) # Later add (X ms in database)
  #{"▸".colorize(:green)} Handled by #{"Home::Index".colorize(:green)}
  #{"▸".colorize(:green)} Rendered #{"User::NewPage".colorize(:green)} within #{"MainLayout".colorize(:green)}
  #{"▸".colorize(:green)} Ran 10 queries - edit config/database.cr for verbose mode
LOG
```

Based on this gist I made: https://gist.github.com/paulcsmith/d7f99bbc8b91cb2d593ed43e3ef595b2

Here's a screenshot
screen shot 2017-11-10 at 3 18 22 pm

The stuff with arrows should only be output if the logger is set to show debug messages, because this would be really noisy (and slow) in production

paulcsmith:
screen shot 2017-11-11 at 1 35 00 pm

Closes #206

### DIFF
--- a/spec/lucky_web/log_handler_spec.cr
+++ b/spec/lucky_web/log_handler_spec.cr
@@ -1,0 +1,52 @@
+require "../spec_helper"
+require "http/server"
+
+include ContextHelper
+
+describe LuckyWeb::LogHandler do
+  it "logs" do
+    io = IO::Memory.new
+    called = false
+    log_io = IO::Memory.new
+    context = build_context_with_io(io)
+
+    call_log_handler_with(log_io, context) { called = true }
+
+    log_output = log_io.to_s
+    log_output.should contain("GET")
+    log_output.should contain("200")
+    log_output.should contain("/")
+    called.should be_true
+  end
+
+  it "logs debug messages" do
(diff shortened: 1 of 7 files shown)


edwardloveall:
Looks really nice, great stuff! The only thing I could think of that would be useful is an absolute time stamp. This would be especially useful for production logs. Also that could be in another PR called "even better logger" 😛

paulcsmith:
Good idea! I'm thinking that it could be a configurable option. In dev it's not really helpful so it could be off, and could be enabled in prod. What do you think?

edwardloveall:
Yeah that sounds good. Maybe something like:

```
LuckyWeb::LogHandler.configure do
  settings.show_absolute_timestamps = Lucky::Env.production?
end
```

paulcsmith:
Yep exactly! I think `show_timestamps` would be a bit more concise, but the idea is spot on

PR #215, opened 2017-11-13
Add ISO 8601 timestamp to request log

### LINKED ISSUE #210: Add an option to show timestamps on same line as status/path etc. in logger
https://github.com/luckyframework/web/pull/207#issuecomment-343697763

This will work similarly to how we do this with the ErrorHandler: https://github.com/luckyframework/web/blob/master/src/lucky_web/error_handler.cr#L4-L6

The options should be called `show_timestamps`

* Use `Habitat.create` to create a `show_timestamps` setting.
* If that setting is true include the timestamp in the `LuckyWeb::LogHandler` output
* If false, do not include it
* See https://github.com/luckyframework/web/blob/master/spec/lucky_web/error_handling_spec.cr#L49 for an example of testing this
* You'll also need to configure the handler in spec_helper: https://github.com/luckyframework/web/blob/master/spec/spec_helper.cr#L21 similar to how the other classes/modules are configured in that file

arr-dev:
Closes #210 

Example when enabled: 
```
GET 200 /? 2017-11-13T15:12:10+0100 (372.0µs)
   ▸ Handled by Home::Index
   ▸ Rendered LuckyWeb::WelcomePage
GET 404 /favicon.ico? 2017-11-13T15:12:10+0100 (137.0µs)
```

### DIFF
--- a/spec/lucky_web/log_handler_spec.cr
+++ b/spec/lucky_web/log_handler_spec.cr
@@ -16,6 +16,7 @@ describe LuckyWeb::LogHandler do
     log_output.should contain("GET")
     log_output.should contain("200")
     log_output.should contain("/")
+    log_output.should_not match(/#{Time.now.to_s("%Y-%m-%d")}/)
     called.should be_true
   end
 
@@ -41,6 +42,7 @@ describe LuckyWeb::LogHandler do
     end
     log_output = log_io.to_s
     log_output.should contain("GET")
+    log_output.should_not match(/#{Time.now.to_s("%Y-%m-%d")}/)
     log_output.should contain("Unhandled exception:")
   end
 
--- a/spec/lucky_web/log_handler_spec.cr
+++ b/spec/lucky_web/log_handler_spec.cr
@@ -43,6 +43,32 @@ describe LuckyWeb::LogHandler do
     log_output.should contain("GET")
     log_output.should contain("Unhandled exception:")
   end
+
+  context "when configured to log timestamps" do
+    it "logs timestamp" do
+      begin
+        LuckyWeb::LogHandler.configure do
+          settings.show_timestamps = true
+        end
+
+        io = IO::Memory.new
+        called = false
+        log_io = IO::Memory.new
+        context = build_context_with_io(io)
+
+        call_log_handler_with(log_io, context) { called = true }
+
+        log_output = log_io.to_s
+        log_output.should contain("GET")
+        log_output.should match(/#{Time.now.to_s("%Y-%m-%d")}/)
+        called.should be_true
+      ensure
+        LuckyWeb::LogHandler.configure do
+          settings.show_timestamps = false
+        end
+      end
+    end
+  end
 end
 
 private def call_log_handler_with(io : IO, context : HTTP::Server::Context, &block)
--- a/src/lucky_web/log_handler.cr
+++ b/src/lucky_web/log_handler.cr
@@ -3,6 +3,10 @@ require "colorize"
 class LuckyWeb::LogHandler
   include HTTP::Handler
 
+  Habitat.create do
+    setting show_timestamps : Bool
+  end
+
   def initialize(@io : IO = STDOUT)
   end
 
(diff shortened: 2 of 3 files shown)


spec/lucky_web/log_handler_spec.cr
```diff
@@ -43,6 +43,32 @@ describe LuckyWeb::LogHandler do
...
+        log_io = IO::Memory.new
+        context = build_context_with_io(io)
+
+        call_log_handler_with(log_io, context) { called = true }
+
+        log_output = log_io.to_s
+        log_output.should contain("GET")
+        log_output.should match(/#{Time.now.to_s("%Y-%m-%d")}/)
+        called.should be_true
+      ensure
```
ugisozols: I'm wondering - isn't context reset between each spec so that the timestamp configured here doesn't leak into other specs? Or it's guarding against some other edge case?

arr-dev: Well, I've implemented as per @paulcsmith's description in #210.
Starting with crystal myself so not really sure about details, 
but it's implemented same as ErrorHandler

paulcsmith: @ugisozols Good question. This is making sure that if an unexpected error happens the configuration is reset. That way it won't break other tests that rely on the timestamp not being shown

src/lucky_web/log_handler.cr
```diff
@@ -12,12 +16,18 @@ class LuckyWeb::LogHandler
...
     elapsed_text = elapsed_text(elapsed)
 
-    @io.puts "#{context.request.method} #{context.response.status_code} #{context.request.resource.colorize(:green)} (#{elapsed_text})"
+    timestamp = if settings.show_timestamps
+                  Time::Format::ISO_8601_DATE_TIME.format(time)
+                else
+                  ""
+                end
+
+    @io.puts "#{context.request.method} #{context.response.status_code} #{context.request.resource.colorize(:green)} #{timestamp} (#{elapsed_text})"
```
paulcsmith: I think this is almost just right, but there will be two spaces if no timestamp is show. I'm thinking you could add the space when formatting the time: `" #{Time::Format::ISO_8601_DATE_TIME.format(Time.now}"`

src/lucky_web/log_handler.cr
```diff
@@ -12,12 +16,18 @@ class LuckyWeb::LogHandler
     elapsed = Time.now - time
     elapsed_text = elapsed_text(elapsed)
 
-    @io.puts "#{context.request.method} #{context.response.status_code} #{context.request.resource.colorize(:green)} (#{elapsed_text})"
+    timestamp = if settings.show_timestamps
```
paulcsmith: This could be it's own private method I think. `private def timestamp` and you could call `Time.new` in it so you don't have to pass anything to it

spec/lucky_web/log_handler_spec.cr
```diff
@@ -43,6 +43,32 @@ describe LuckyWeb::LogHandler do
...
+        io = IO::Memory.new
+        called = false
+        log_io = IO::Memory.new
+        context = build_context_with_io(io)
+
+        call_log_handler_with(log_io, context) { called = true }
+
+        log_output = log_io.to_s
+        log_output.should contain("GET")
+        log_output.should match(/#{Time.now.to_s("%Y-%m-%d")}/)
```
paulcsmith: Could you add a `should_not` expectation to the other test to make sure it's only shown if `show_timestamp` is true?

paulcsmith:
Thanks for working on this. It's looking really close

src/lucky_web/log_handler.cr
arr-dev: wouldn't that record timestamp of request end instead of start?

paulcsmith: Yes you're right. You should pass the time so it records when it started. Good catch!

spec/lucky_web/log_handler_spec.cr
arr-dev: Sure

src/lucky_web/log_handler.cr
arr-dev: Yes, you are right

src/lucky_web/log_handler.cr
```diff
@@ -40,4 +44,12 @@ class LuckyWeb::LogHandler
 
     "#{(millis * 1000).round(2)}µs"
   end
+
+  private def timestamp(time)
+    if settings.show_timestamps && time
```
arr-dev: @paulcsmith compiler complained without checking that `time` is set. Not sure whether best choice is this conditional or `time ||= Time.now` before so it always logs timestamp, even if it's maybe bit off. It would probably help with parsing if timestamp is always present so format is consistent?

paulcsmith: Ah I see. The compiler caught a potential bug. The `time` might be `nil` because it could raise and therefore `time` may not have been set. I think what we should do is something like this:

```crystal
This way if `time` is nil it is set to Time.now
private def timestamps(time : Time = Time.now)
```

Like you said it may be a bit off, but I think that's better than not having any timestamp

arr-dev: Fixed, default value didn't work, but `time || Time.now` did

arr-dev:
Should I add a PR to [luckyframework/cli](github.comluckyframework/cli) to add to app skeleton, 
like https://github.com/luckyframework/cli/blob/master/src/web_app_skeleton/config/error_handler.cr
or set default value for config to false?
Currently, app wouldn't work without manually setting value for `show_timestamps`

paulcsmith:
@arr-dev That would be great! Could you add it to web_app_skeleton/config/log_handler.cr? Thank you so much!

Issue #209, opened 2017-11-12
Add helpers like those found in ActionView::Helpers (Rails)


paulcsmith:
Things like:

- [x] number_to_currency http://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html
- [ ] time_ago_in_words https://apidock.com/rails/ActionView/Helpers/DateHelper/time_ago_in_words
- [x] simple_format https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format
- [x] pluralize https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize unlike Rails, always require specifying the plural version of the word.

I think these can go in  a new module called `LuckyWeb::PageHelpers` and would be included in `LuckyWeb::HTMLPage`. If someone wants to help with this, feel free to tackle just one at a time.

paulcsmith:
If someone decides to start on one of these, please comment here so two people don't accidentally do the same work. Thanks!

actsasflinn:
I'm working on the inflector code now. I'm planning to port portions of ActiveSupport::Inflector and then make them available similar to ActionView::Helpers::TextHelper.

ActiveSupport is useful outside of the web package as inflections are used in various places in Rails. ActiveRecord particularly and various generators in particular. I started into creating some scaffold generators and found a shortcoming in lack of inflectors.

ugisozols:
I'm going to try and tackle the `number_to_currency` 😎

azbshiri:
I'm going to write `pluralize`

paulcsmith:
@azbshiri Just a heads up that @actsasflinn has started on the inflector so that needs to be merged into the `luckyframework/support` repo first. 

Until that's in, maybe a good option would be `time_ago_in_words`? It's a pretty helpful one that I'd *love* to get in soon

azbshiri:
#221

actsasflinn:
I'm working on simple_format now

actsasflinn:
👍

PR #221, opened 2017-11-17
Implement `time_ago_in_words` (RubyOnRails)

azbshiri:


actsasflinn:
Perhaps this should go in DateHelpers instead of PageHelpers?

Both **distance_of_time_in_words** and **distance_of_time_in_words_to_now** would be useful as well.

paulcsmith:
I think we should definitely have a `to_now` option, but I think we could keep it all in `time_ago_in_words` and make the `to` option default to `Time.new`. What do you think?

actsasflinn:
**distance_of_time_in_words** is the core method that both **distance_of_time_in_words_to_now** and **time_ago_in_words** are based.

**distance_of_time_in_words_to_now**  is an alias for  **time_ago_in_words** and both names accurately describes it's relativity to now.

**distance_of_time_in_words** on the other hand is between from and two and not relative to now. They could be the future, two points in the past, etc.

I'd put the core logic into **distance_of_time_in_words** and then implement **time_ago_in_word** with the to: Time.new

azbshiri:
I plan to do so tomorrow. just did a pull request to remind myself.

azbshiri:
@paulcsmith @actsasflinn please review the changes I think it's on a good state for now at least.

actsasflinn:
@azbshiri good stuff. A couple questions/comments...

1. Is db83f6e relevant to the feature?

2. The file date_helpers.cr appears in src/lucky_web and is under the module space LuckyWeb. Looks like you should absorb the move to Lucky and src/lucky.

3. Are you intending a port of the RoR feature? distance_of_time_in_words in the RoR implementation supports a  **:include_seconds** option which isn't present yet in your feature. **:include_seconds** true and false both behave differently in the RoR implementation than in your feature branch.

```crystal
view.distance_of_time_in_words(from, to + 0.seconds, include_seconds: true).should eq "less than 5 seconds"
view.distance_of_time_in_words(from, to + 0.seconds, include_seconds: false).should eq "less than a minute"
```

azbshiri:
1. I run `crystal tool format` after save, so it's just formatting.
3. I think we can add this features one at a time, @paulcsmith what do you think?
2. Don't know, @paulcsmith what do you think?

src/lucky_web/date_helpers.cr
```diff
@@ -0,0 +1,47 @@
+module LuckyWeb::DateHelpers
```
paulcsmith: Can we call this `TimeHelpers` instead? Also this should be in `lucky_web` and under `Lucky` namespace now. Sorry for the hassle! This was recently changed. But better now than later

src/lucky_web/date_helpers.cr
```diff
@@ -0,0 +1,47 @@
...
+    case distance
+    when       1 then "a minute"
+    when 2...45  then "#{distance} minutes"
+    when 45...90 then "about an hour"
+    end
+  end
+
+  private def distance_in_seconds(distance : Int32)
+    distance == 1 ? "a second" : "#{distance} seconds"
+  end
```
paulcsmith: I think this is really great. Thanks for doing this :)

src/lucky_web/date_helpers.cr
```diff
@@ -0,0 +1,47 @@
+module LuckyWeb::DateHelpers
+  def distance_of_time_in_words(from : Time, to : Time)
+    minutes = (to - from).minutes
+    seconds = (to - from).seconds
+    hours = (to - from).hours
+    days = (to - from).days
+
+    return distance_in_days(days) if days != 0
```
paulcsmith: Instead of returning a string, can you call `text` in this method and in `time_ago_in_words`? That way in pages you only need to do:

```crystal
time_ago_in_words Time.new - 1.minute

# instead of
text time_ago_in_words(Time.new - 1.minute)`
```

This would mean you can't use these helpers outside of views, but that is ok for now. We can figure out how to extract them later on. I don't think it will be too tricky

paulcsmith:
Just saw the other comment. 

1. Thanks for running format and catching that!
2. Yeah it should be under `Lucky` now
3. I think this is a good start. We can consider including those other options later on

src/lucky_web/date_helpers.cr
azbshiri: Sure. you mean under `lucky_web` directory but call module `Lucky`. How about spec? I just changed the directory too.

src/lucky_web/date_helpers.cr
azbshiri: No problem I figure it out.

src/lucky_web/date_helpers.cr
paulcsmith: Looks like you got it right. `Lucky` namespace in `src/lucky` directory 👍

azbshiri: It was `src/lucky` but you said `lucky_web` so I changed it again 😭

paulcsmith: Oops! I'll update my comment :P

azbshiri:
I'm having trouble with changing `DateHelpers` to `TimeHelpers` I think I was come up with the same name but I can't run specs and it doesn't find `Lucky::TimeHelpers`
```
in src/lucky/error_action.cr:1: while requiring "./*"

require "./*"
^

in src/lucky/html_page.cr:16: undefined constant Lucky::TimeHelpers(did you mean 'Lucky::LinkHelpers')

  include Lucky::TimeHelpers
```

paulcsmith:
@azbshiri Require works alphabetically, so `TimeHelpers` is not required yet. You should be able to do something like `require "./time_helpers` to tell Crystal to require that file first. LMK if that does the trick

azbshiri:
@paulcsmith would you merge it?

paulcsmith:
@azbshiri It is still missing changes for this comment: https://github.com/luckyframework/lucky/pull/221#discussion_r153024576 LMK if you need any help with that change. I'm happy to help :)

azbshiri:
oh, forgot that 😄

paulcsmith:
Hey @azbshiri I'm going to finish off the converstion to use `text` so #209 can be closed. I hope that's ok! Thanks for getting it this far. It looks great

PR #223, opened 2017-11-17
Fine tune hash "get!" specs

hanneskaeufler:
Fix copy-pasted it-description and remove unneeded expectation inside expectation.
I didn't actually run this (shame on me) but the `.should be_nil` inside the `expect_raises` seems unneeded and confused me for a second.


paulcsmith:
Nice catch 👍

PR #226, opened 2017-11-18
add .DS_Store to .gitignore

actsasflinn:
Resolves issue #225 

Probably needed across the suite not just web


paulcsmith:
Good catch. Thank you

PR #232, opened 2017-11-19
Typed status  codes for render methods

hanneskaeufler:
Fiddling with this :) Do you still want to keep around the versions of redirect which take an int? Probably keep them around, but advertise the typed with Status variants 👍

To be honest I'm not 100% sold on the `Status` name. What do you think of `StatusCode`, `HttpStatus`?

The status codes are lifted from rack.


hanneskaeufler:
Did you comment something, I don't see any review comments ❓

src/lucky_web/status.cr
```diff
@@ -0,0 +1,63 @@
+module LuckyWeb
+  enum Status
```
paulcsmith: Hmm for some reason it did not save my comment. Here's roughly what I said:

I agree that `Status` is probably a bit ambiguous in this context, but if it was moved to `LuckyWeb::Action::Status` then I think it would make more sense, and you'd be able to use it like this in actions: `redirect to: "/foo", Status::MovedPermanently`

hanneskaeufler:
Oh yeah moving `Status` into `Action` made sense. No idea if there is a better way to include the enum there, other than pasting it into `action.cr`.

paulcsmith:
I think copy pasting the enum is ok for now. We can consider extracting it later I think.

paulcsmith:
Awesome work :) thanks!

PR #234, opened 2017-11-19
Add number_to_currency

ugisozols:
For https://github.com/luckyframework/web/issues/209.

I'm ready for my first PR teardown. 😎

### DIFF
--- a/spec/lucky_web/number_to_currency_spec.cr
+++ b/spec/lucky_web/number_to_currency_spec.cr
@@ -21,10 +21,6 @@ describe LuckyWeb::NumberToCurrency do
       view.number_to_currency(92).should eq "$92.00"
     end
 
-    it "accepts Nil" do
-      view.number_to_currency(nil).should eq nil
-    end
-
     describe "options" do
       it "accepts precision" do
         view.number_to_currency(29.929123, precision: 3).should eq "$29.929"
--- a/spec/lucky_web/number_to_currency_spec.cr
+++ b/spec/lucky_web/number_to_currency_spec.cr
@@ -7,7 +7,7 @@ private class TestPage
   end
 end
 
-describe LuckyWeb::PageHelpers do
+describe LuckyWeb::NumberToCurrency do
   describe "number_to_currency" do
     it "accepts Float" do
       view.number_to_currency(29.92).should eq "$29.92"
--- a/src/lucky_web/html_page.cr
+++ b/src/lucky_web/html_page.cr
@@ -1,4 +1,5 @@
 require "./tags/**"
+require "./page_helpers"
 
 module LuckyWeb::HTMLPage
   include LuckyWeb::BaseTags
@@ -11,6 +12,7 @@ module LuckyWeb::HTMLPage
   include LuckyWeb::SpecialtyTags
   include LuckyWeb::Assignable
   include LuckyWeb::AssetHelpers
+  include LuckyWeb::PageHelpers
 
   macro setup_initializer_hook
     macro finished
--- a/src/lucky_web/html_page.cr
+++ b/src/lucky_web/html_page.cr
@@ -1,5 +1,5 @@
 require "./tags/**"
-require "./page_helpers"
+require "./page_helpers/**"
 
 module LuckyWeb::HTMLPage
   include LuckyWeb::BaseTags
@@ -12,7 +12,7 @@ module LuckyWeb::HTMLPage
   include LuckyWeb::SpecialtyTags
   include LuckyWeb::Assignable
   include LuckyWeb::AssetHelpers
-  include LuckyWeb::PageHelpers
+  include LuckyWeb::NumberToCurrency
 
   macro setup_initializer_hook
     macro finished
--- a/src/lucky_web/page_helpers/number_to_currency.cr
+++ b/src/lucky_web/page_helpers/number_to_currency.cr
@@ -6,7 +6,14 @@ module LuckyWeb::NumberToCurrency
   DEFAULT_DELIMITER_REGEX = /(\d)(?=(\d\d\d)+(?!\d))/
   DEFAULT_FORMAT          = "%u%n"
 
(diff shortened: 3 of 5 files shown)


src/lucky_web/page_helpers/number_to_currency.cr
```diff
@@ -0,0 +1,33 @@
...
+  DEFAULT_UNIT            = "$"
+  DEFAULT_SEPARATOR       = "."
+  DEFAULT_DELIMITER       = ","
+  DEFAULT_DELIMITER_REGEX = /(\d)(?=(\d\d\d)+(?!\d))/
+  DEFAULT_FORMAT          = "%u%n"
+
+  def number_to_currency(value : Nil, **options)
+  end
+
+  def number_to_currency(value : Float | Int32, precision : Int32 = DEFAULT_PRECISION, unit : String = DEFAULT_UNIT, separator : String = DEFAULT_SEPARATOR, delimiter : String = DEFAULT_DELIMITER, delimiter_pattern : Regex = DEFAULT_DELIMITER_REGEX, format : String = DEFAULT_FORMAT, negative_format : String = DEFAULT_FORMAT)
```
ugisozols: Is there a common style for such long argument lists?

spec/lucky_web/number_to_currency_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+
+private class TestPage
+  include LuckyWeb::HTMLPage
+
+  def render
+  end
+end
+
+describe LuckyWeb::NumberToCurrency do
+  describe "number_to_currency" do
```
ugisozols: Is spec format (separate test cases) OK or it'd be better to do them [like Rails with everything put under one test case](https://github.com/rails/rails/blob/19fc41992561c901f66064042aa23f4910101d98/activesupport/test/number_helper_test.rb#L64-L79)?

src/lucky_web/html_page.cr
```diff
@@ -1,4 +1,5 @@
 require "./tags/**"
+require "./page_helpers/**"
```
ugisozols: There was `luckyframework/support` repo which was renamed to inflector. Does it still make sense to put this and other helpers alike in support?

spec/lucky_web/number_to_currency_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+
+    it "accepts String" do
+      view.number_to_currency("92.29").should eq "$92.29"
+    end
+
+    it "accepts Integer" do
+      view.number_to_currency(92).should eq "$92.00"
+    end
+
+    it "accepts Nil" do
```
ugisozols: Should this be allowed?

spec/lucky_web/number_to_currency_spec.cr
paulcsmith: I think what you have here is great :) If it gets much bigger then it might be better to combine things, but for now it's 👍

spec/lucky_web/number_to_currency_spec.cr
paulcsmith: Good question. I think it should not be allowed

spec/lucky_web/number_to_currency_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      it "accepts format" do
+        view.number_to_currency("1234567890.50", format: "<b>%n</b> %u").should eq "<b>1,234,567,890.50</b> $"
+      end
+
+      it "accepts negative format" do
+        view.number_to_currency("-1234567890.50", negative_format: "<b>%n</b> - %u").should eq "<b>1,234,567,890.50</b> - $"
+      end
+    end
```
paulcsmith: Very nice test descriptions and test cases 👍

src/lucky_web/html_page.cr
paulcsmith: That's also a really good question. I will be creating a `luckyframework/utils` repo where we can put stuff like these number helpers, but for now we can keep them in web. I'll split them up later when I have a bit more time. For now it's easier to keep them here.

src/lucky_web/page_helpers/number_to_currency.cr
paulcsmith: Here's what I *think* you can do. Put the `value` argument on a new line and maybe precision too. Then run `crystal tool format` I think that it will put the rest of them on newlines and format them. LMK if that does the trick!

src/lucky_web/page_helpers/number_to_currency.cr
```diff
@@ -0,0 +1,33 @@
...
+    number_to_currency(value.to_s, precision: precision, unit: unit, separator: separator, delimiter: delimiter, delimiter_pattern: delimiter_pattern)
+  end
+
+  def number_to_currency(value : String, precision : Int32 = DEFAULT_PRECISION, unit : String = DEFAULT_UNIT, separator : String = DEFAULT_SEPARATOR, delimiter : String = DEFAULT_DELIMITER, delimiter_pattern : Regex = DEFAULT_DELIMITER_REGEX, format : String = DEFAULT_FORMAT, negative_format : String = DEFAULT_FORMAT)
```
paulcsmith: Rather than having two methods, you could do this:

```crystal
def number_to_currency(value : String | Int32 | Float, ...)
  value = value.to_s # Force it into a string. This way you don't need two different methods
```

ugisozols:
Took care of the suggestions.

src/lucky_web/html_page.cr
actsasflinn: If we use Rails as a model then there's both ActionView/Helpers/NumberHelper and ActiveSupport::NumberHelper. One module is exposed to the view and one more general purpose. 

I'm not sure which this is based on but ActionView/Helpers/NumberHelper and ActiveSupport::NumberHelper have a number of other useful methods.

PR #237, opened 2017-11-20
Issue 209 text helper

actsasflinn:
implements truncate, highlight, excerpt, pluralize, word_wrap, cycle TextHelper methods as requested by issue #209


spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+    view.excerpt("my very\nvery\nvery long\nstring", "long", separator: "\n", radius: 0).should eq "...very long..."
+    view.excerpt("my very\nvery\nvery long\nstring", "long", separator: "\n", radius: 1).should eq "...very\nvery long\nstring"
+    view.excerpt("This is a beautiful morning", "a", separator: nil).should eq view.excerpt("This is a beautiful morning", "a")
+  end
+
+  it "should word wrap" do
+    view.word_wrap("my very very very long string", line_width: 15).should eq "my very very\nvery long\nstring"
+  end
+
+  it "test_word_wrap_with_extra_newlines" do
```
hanneskaeufler: `it "wraps with extra newlines"`?

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+    view.word_wrap("my very very very long string\n\nwith another line", line_width: 15).should eq "my very very\nvery long\nstring\n\nwith another\nline"
+  end
+
+  it "test_word_wrap_with_custom_break_sequence" do
```
hanneskaeufler: test name could be improved as well

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+  end
+
+  def test_truncate
+    truncate "Hello World", length: 8 do
+      link "Continue", "#"
+    end
+  end
+
+  def test_highlight
+    highlight "This is a beautiful morning, but also a beautiful day", "beautiful" do |word|
```
paulcsmith: Same question with this one as well

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+    view.truncate("Hello World!", length: 12).should eq "Hello World!"
+    view.truncate("Hello World!!", length: 12).should eq "Hello Wor..."
+  end
+
+  it "truncates with default length of 30" do
+    str = "This is a string that will go longer then the default truncate length of 30"
+    view.truncate(str).should eq str[0...27] + "..."
+  end
+
+  it "truncates with options hash" do
```
paulcsmith: It's not technically a has so I think this could be just "with options"

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+  end
+
+  it "highlights" do
+    view.highlight("This is a beautiful morning", "beautiful").should eq "This is a <mark>beautiful</mark> morning"
+    view.highlight("This is a beautiful morning, but also a beautiful day", "beautiful").should eq "This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day"
+    view.highlight("This is a beautiful morning, but also a beautiful day", "beautiful", highlighter: "<b>\\1</b>").should eq "This is a <b>beautiful</b> morning, but also a <b>beautiful</b> day"
+    view.highlight("This text is not changed because we supplied an empty phrase", nil).should eq "This text is not changed because we supplied an empty phrase"
+  end
+
+  it "highlights pending" do
```
paulcsmith: Maybe "does not highlight empty text"?

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+    view.highlight("<p class=\"beautiful\">This is a beautiful morning, but also a beautiful day</p>", "beautiful").should eq "<p class=\"beautiful\">This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day</p>"
+    view.highlight("<p>This is a beautiful <a href=\"http://example.com/beautiful\#top?what=beautiful%20morning&when=now+then\">morning</a>, but also a beautiful day</p>", "beautiful").should eq "<p>This is a <mark>beautiful</mark> <a href=\"http://example.com/beautiful\#top?what=beautiful%20morning&when=now+then\">morning</a>, but also a <mark>beautiful</mark> day</p>"
+    view.highlight("<div>abc div</div>", "div", highlighter: "<b>\\1</b>").should eq "<div>abc <b>div</b></div>"
+  end
+
+  it "highlights with block" do
+    view.highlight("one two three", ["one", "two", "three"]) { |word| "<b>#{word}</b>" }.should eq "<b>one</b> <b>two</b> <b>three</b>"
+    view.test_highlight.should eq "This is a <a href=\"#beautiful\">beautiful</a> morning, but also a <a href=\"#beautiful\">beautiful</a> day"
+  end
```
paulcsmith: Could you group the different type of methods in a `describe` block? Like `describe "highlighting/truncating/cycling/etc."? I think that might aid readability

src/charms/string_extensions.cr
```diff
@@ -1,12 +1,4 @@
 class String
-  def dasherize
-    self.gsub("_", "-")
-  end
-
-  def humanize
-    self.downcase.gsub("_", " ").capitalize
-  end
-
```
paulcsmith: Nice 👍

src/lucky_web/tags/text_helpers.cr
```diff
@@ -0,0 +1,217 @@
...
+  def truncate(text, length = 30, omission = "...", separator = nil, escape = false, blk : Nil | Proc = nil)
+    if text
+      content = truncate_text(text, length, omission, separator)
+      content = escape ? HTML.escape(content) : content
+      content += blk.call.to_s if blk.is_a?(Proc) && text.size > length
+      content
+    end
+  end
+
+  def truncate(text, length = 30, omission = "...", separator = nil, escape = true, &block : -> _)
```
paulcsmith: Can you add type specific types for all these `text : String, length : Int32 = 30`, etc.? It makes the docs read a bit better because people can more easily see the types

src/lucky_web/tags/text_helpers.cr
```diff
@@ -0,0 +1,217 @@
+module LuckyWeb::TextHelpers
+  @@_cycles = Hash(String, Cycle).new
+
+  def truncate(text, length = 30, omission = "...", separator = nil, escape = false, blk : Nil | Proc = nil)
+    if text
```
paulcsmith: I think we should disallow nil text and only allow String instead.

src/lucky_web/tags/text_helpers.cr
```diff
@@ -0,0 +1,217 @@
...
+      if separator
+        text.rindex(separator, length_with_room_for_omission) || length_with_room_for_omission
+      else
+        length_with_room_for_omission
+      end
+
+    "#{text[0, stop]}#{omission}"
+  end
+
+  def highlight(text, phrases : Nil | String | Regex | Array(String | Regex), highlighter : Proc | String = "<mark>\\1</mark>")
```
paulcsmith: I think we should disallow a `nil` phrase as well. It seems like that would probably be a bug if someone tried to do that

src/lucky_web/tags/text_helpers.cr
```diff
@@ -0,0 +1,217 @@
...
+  end
+
+  def highlight(text, phrases : Nil | String | Regex | Array(String | Regex), &block : String -> _)
+    highlight(text, phrases, highlighter: block)
+  end
+
+  def excerpt(text, phrase : Nil | Regex | String, separator = "", radius = 100, omission = "...")
+    return "" if text.to_s.blank?
+
+    separator = "" if separator.nil?
```
paulcsmith: I think separator should not ever be `nil` since the default is `""`. I think that would be a bit safer

src/lucky_web/tags/text_helpers.cr
```diff
@@ -0,0 +1,217 @@
...
+      end
+
+    "#{text[0, stop]}#{omission}"
+  end
+
+  def highlight(text, phrases : Nil | String | Regex | Array(String | Regex), highlighter : Proc | String = "<mark>\\1</mark>")
+    if text.to_s.blank? || phrases.to_s.blank?
+      text.nil? ? "" : text
+    else
+      unless phrases.is_a?(Array)
```
paulcsmith: Instead of allow non-arrays, I think we can require an array. That way this code is unnecessary and we have less tests. I also think it might read a bit better that way: `phrases: "a phrase"` seems a bit strange because `phrases` is plural, but you're passing in just one object.

spec/lucky_web/text_helpers_spec.cr
```diff
@@ -0,0 +1,294 @@
...
+
+private class TestPage
+  include LuckyWeb::HTMLPage
+
+  def render
+  end
+
+  def test_truncate
+    truncate "Hello World", length: 8 do
+      link "Continue", "#"
```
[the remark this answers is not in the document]
actsasflinn: This is again based on Rails ActionView implementation of the same. truncate_text is not actually implemented in the TextHelper but in a String extension and does not escape HTML by default. The helper implementation is useful for views specifically, escaping HTML in text or allowing access to other helper methods.

In this case the block is used as the "more" link

"Hello World" becomes
"Hello...<a href=\"#\">Continue</a>"

The omission is only used if truncation occurs.

    view.truncate("Hello World", 12){ view.link "Continue", "#more" }.should eq "Hello World"

spec/lucky_web/text_helpers_spec.cr
actsasflinn: This block is slightly different than how truncate acts but similar in purpose. In the case of highlight, the block is passed to gsub as the substitution. The block is definitely a matter of preference since you can accomplish the replace another way (as with gsub).

Beauty is in the eye of the beholder but perhaps this contrived example illustrates the utility of the block form a bit more.

```crystal
text = "This is a beautiful morning, but also a beautiful day"
highlight text, "beautiful" do |word|
  span word, "data-highlight-word": word, "data-color": "yellow", "data-español": "bello"
end
```
vs.
```crystal
text = "This is a beautiful morning, but also a beautiful day"
highlight text, "beautiful", highlighter: span(word, "data-highlight-word": word, "data-color": "yellow", "data-español": "bello").to_s
```

Admittedly contrived but the latter is a bit of a word jumble on one line to me.

src/lucky_web/tags/text_helpers.cr
actsasflinn: I'm not attached to the nil check. There are specific test cases in Rails for a reason and my guess is nilable text from DB records.

src/lucky_web/tags/text_helpers.cr
actsasflinn: Yup, good catch

actsasflinn:
Ok, 485dfde has suggested changes from PR and includes simple_format as well. Thx -- Flinn

actsasflinn:
I just rebased so it should merge cleanish.

spec/lucky/tag_helpers_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+    content_tag "p" do
+      text "Foo bar baz"
+    end
+  end
+
+  def test_content_tag_without_block
+    content_tag "p", "Foo bar baz"
+  end
+
+  def test_content_tag_with_block_and_attrubute
```
olimart: typo in `attribute`

spec/lucky/tag_helpers_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+    content_tag "a", href: "#foo" do
+      text "Foo bar baz"
+    end
+  end
+
+  def test_content_tag_without_block_and_extra_attrubute
```
olimart: typo in `attribute`

spec/lucky/tag_helpers_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+    content_tag "a", "Foo bar baz", href: "#foo", bar: "baz"
+  end
+
+  def test_content_tag_with_block_and_attrubutes_as_hash
```
olimart: typo in `attribute`

spec/lucky/tag_helpers_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+    content_tag "a", { href: "#foo", bar: "baz" } do
+      text "Foo bar baz"
+    end
+  end
+
+  def test_content_tag_with_symbol_and_attrubutes
```
olimart: typo in `attribute`

spec/lucky/tag_helpers_spec.cr
```diff
@@ -0,0 +1,84 @@
...
+    content_tag :a, "Foo bar baz", href: "#foo", bar: "baz"
+  end
+
+  def test_content_tag_with_symbol_block_and_attrubutes_as_hash
```
olimart: typo in `attribute`

spec/lucky/text_helpers_spec.cr
```diff
@@ -0,0 +1,343 @@
...
+      view.cycle("red", "blue", name: "colors")
+      view.current_cycle("colors").should eq "red"
+    end
+
+    it "gets current cycle with no exceptions" do
+      view.current_cycle.should be_nil
+      view.current_cycle("colors").should be_nil
+    end
+
+    it "gets current cycle with more than two names" do
```
paulcsmith: I've been thinking about this a bit, and it sees this test is pretty big. What do you think about splitting it into separate `highlight_spec`, `truncate_spec`, etc?

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,233 @@
+module Lucky::TextHelpers
+  @@_cycles = Hash(String, Cycle).new
+
+  def truncate(text : String, length : Int32 = 30, omission : String = "...", separator : String | Nil = nil, escape : Bool = false, blk : Nil | Proc = nil)
+    if text
+      content = truncate_text(text, length, omission, separator)
+      content = escape ? HTML.escape(content) : content
+      content += blk.call.to_s if blk.is_a?(Proc) && text.size > length
```
paulcsmith: It's more common to check that something is not `nil` like this `if !blk.nil?`, rather than checking it's type.

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,233 @@
...
+      content
+    end
+  end
+
+  def truncate(text : String, length : Int32 = 30, omission : String = "...", separator : String | Nil = nil, escape : Bool = true, &block : -> _)
+    truncate(text, length, omission, separator, escape, blk: block)
+  end
+
+  # This could go in a String extension
```
paulcsmith: I agree. This should probably go in utils or in an optional string extension at some point. Can you remove the comment for now though?

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,233 @@
...
+      if separator
+        text.rindex(separator, length_with_room_for_omission) || length_with_room_for_omission
+      else
+        length_with_room_for_omission
+      end
+
+    "#{text[0, stop]}#{omission}"
+  end
+
+  def highlight(text : String, phrases : String | Regex | Array(String | Regex), highlighter : Proc | String = "<mark>\\1</mark>")
```
paulcsmith: Could we change this so `phrases` is `phrases : Array(String | Regex)` so you must always pass an array? That'll simplify the tests and code and I think will look cleaner: `phrases: ["some phrase"]`

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,233 @@
...
+  end
+
+  def word_wrap(text : String, line_width : Int32 = 80, break_sequence : String = "\n")
+    text = text.split("\n").map do |line|
+      line.size > line_width ? line.gsub(/(.{1,#{line_width}})(\s+|$)/, "\\1#{break_sequence}").strip : line
+    end
+    text.join(break_sequence)
+  end
+
+  def simple_format(text : String, **html_options, wrapper_tag : String | Symbol = "p")
```
paulcsmith: I was thinking about this a lot and I think we can remove `content_tag` and the capture helper by passing in an optional block here instead. I think it would be more powerful too:

```crystal
simple_format("my cool test") do |formatted_text|
  p formatted_text, class: "this-is-a-custom-class"
end

# or just `simple_format` if you don't need anything fancy
simple_format("my cool test")
```

paulcsmith:
If I didn't make sense with any of those comments, LMK and I'll try to clarify what I mean :)

src/lucky/tags/text_helpers.cr
actsasflinn: Interesting. This works in this scenario where blk is either Nil or Proc.

Side note, I struggle with the optional Block/Proc a bit with this in Crystal. The Ruby way might be to check if the object respond_to? but the Crystal convention seems to force the object to conform to a type or interface.

src/lucky/tags/text_helpers.cr
actsasflinn: thoughts on doing that plus an overload?

```crystal
  def highlight(text : String, phrases : Array(String | Regex), highlighter : Proc | String = "<mark>\\1</mark>")
  # ...
  def highlight(text : String, phrase : String | Regex, highlighter : Proc | String = "<mark>\\1</mark>")
    phrases = [phrase]
    highlight(text, phrases, highlighter)
```

spec/lucky/text_helpers_spec.cr
actsasflinn: On returning text vs. string. There's probably cases where ```raw``` makes sense as well. The Rails implementation has a escape option on many of the methods to allow for this. Night sure the right answer but we should think about that a bit.

src/lucky/tags/text_helpers.cr
actsasflinn: I like the block syntax a lot. It simplifies the implementation a lot.

I deliberated on this a lot honestly. I didn't like how ugly ```capture``` was then I realized Rails capture implementation is almost verbatim ugly. ```capture``` is useful for a number of things within Rails, but Lucky may be solving for those in other ways.

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,244 @@
...
+    first_part, second_part = text.split(phrase, 2)
+
+    prefix, first_part   = cut_excerpt_part(:first, first_part, separator, radius, omission)
+    postfix, second_part = cut_excerpt_part(:second, second_part, separator, radius, omission)
+
+    affix = [first_part, separator, phrase, separator, second_part].join.strip
+    [prefix, affix, postfix].join
+  end
+
+  def pluralize(count : Int32 | String | Nil, singular : String, plural = nil)
```
paulcsmith: Sorry I didn't see this earlier. I think that the count should only be `Int32`. If people want to use a string they can call `to_i` on it. I think this is important because most of the time you want to use an Int32, and if you accidentally pass String, you'd want to know. If you really did mean to pass a String, calling `to_i` in those rare instances should be ok

src/lucky/tags/text_helpers.cr
```diff
@@ -0,0 +1,244 @@
...
+  end
+
+  def simple_format(text : String, **html_options, wrapper_tag : String | Symbol = "p")
+    paragraphs = split_paragraphs(text)
+
+    paragraphs = [""] if paragraphs.empty?
+
+    paragraphs.map { |paragraph|
+      content_tag(wrapper_tag, paragraph, **html_options)
+    }.join("\n\n")
```
paulcsmith: The version without a block could do something like this to remove duplicate code:

```crystal
def simple_format(text : String)
  simple_format(text) do |formatted_text|
    para formatted_text
  end
end
```

What do you think?

src/lucky/tags/text_helpers.cr
actsasflinn: It's a feature in it's current form not an accident enabler, I deliberated on this one a bit when adding the types. The spec shows some of the flexibility you get in it's current form.

```crystal
view.pluralize("1,066", "count").should eq "1,066 counts"
view.pluralize("1.25", "count").should eq "1.25 counts"
view.pluralize("1.0", "count").should eq "1.0 count"
view.pluralize("1.00", "count").should eq "1.00 count"
view.pluralize(2, "count", "counters").should eq "2 counters"
view.pluralize(nil, "count", "counters").should eq "0 counters"
```

paulcsmith: I think we could cover most of those cases by also allowing `Float`. Then people could call `to_f` on the string. I'm not totally sure about the `nil` case, but I'd rather remove it because it seems odd to have something be `nil` like that in practice. Usually you would use an array of items and call `size`, so you shouldn't have `nil` in most cases. I'd rather be a bit more strict early on and then add more types. That way we know for sure that people need them and know the exact use-case of it

src/lucky/tags/text_helpers.cr
actsasflinn: Fails a few tests

```crystal
view.simple_format("We want to put a wrapper...\n\n...right there."){ |txt| view.div txt }.should eq "<div>We want to put a wrapper...</div>\n\n<div>...right there.</div>"
```
```
       Expected: "<div>We want to put a wrapper...</div>\n\n<div>...right there.</div>"
            got: "\n\n"
```

```crystal
view.simple_format(""){ |txt| view.div txt }.should eq "<div></div>"
```
```
       Expected: "<div></div>"
            got: ""
```

```crystal
view.simple_format("crazy\r\n cross\r platform linebreaks").should eq "<p>crazy\n<br /> cross\n<br /> platform linebreaks</p>"
```
```
       Expected: "<p>crazy\n<br /> cross\n<br /> platform linebreaks</p>"
            got: "<p>crazy\n&lt;br /&gt; cross\n&lt;br /&gt; platform linebreaks</p>"
```

paulcsmith: For the first spec I think you need to call `to_s` on the result since it returns an `IO::Memory` now and not a String. The second on I think should be an empty string since there is no text, so there shouldn't be a wrapper for the paragraphs. I think for the third one it would be a bit trickier. I'm not totally sure how to handle that one. 

I think for now we should remove the option to use a custom block. Instead let's do just `para` as the default, remove the capture and content_tag stuff and see if we can figure out a way to do it with the block later. That way it doesn't hold up the rest of this PR for too long. What do you think?

actsasflinn: I prefer the working software ;-) You really that against content_tag and capture? Can we make them private to contain their use? I have a sneaking suspicion they are useful for other purposes.

If we don't want to hold up the other helper I can break them into their own files.

spec/lucky/text_helpers/pluralize_spec.cr
```diff
@@ -0,0 +1,24 @@
+require "./text_helpers_spec"
+
+describe Lucky::TextHelpers do
+  describe "pluralize" do
+    it "pluralizes words" do
```
paulcsmith: Thanks for separating these specs. Great idea on putting them in `text_helpers` 👍

src/lucky/tags/text_helpers.cr
paulcsmith: Yeah working software is good haha. I am pretty against content_tag for now (but maybe add it in the future as a separate PR), but I'm pretty against capture at this point. It may be useful, but I'd rather wait to add it. I'm fairly certain this feature can be done without it, but I don't have time to dive into it right now :( 

I'd be ok with separating the simple_format stuff into their own PR. This one is a bit big so that would make it easier to review as well 👍

src/lucky/tags/text_helpers.cr
paulcsmith: Actually you're right about keeping strings. I can see why that would be useful so you can pass a formatted/localized number. In the future we may want to try to do it with a block though so that you pass in an int or float and can format it in a block or something: 

```crystal
pluralize(1_300, "thing") do |number|
  humanize_number(number) # returns "1,300"
end
```

If that's not too much work we could do it in this PR, but might be better to do separately so it doesn't hold this one up too long. I still think we should remove the `nil` option though. I could see that causing bugs

spec/lucky/text_helpers/truncate_spec.cr
```diff
@@ -6,33 +6,44 @@ class TextHelperTestPage
       link "Continue", "#"
     end
   end
+
+  def text_truncate_with_block_invoked
+    truncate("Here is a long test and I need a continue to read link", length: 27) do
```
paulcsmith: I love this 👍

PR #239, opened 2017-11-21
[WIP] Accept multipart forms

### LINKED ISSUE #204: Add multipart form upload support

jdbann:
Accept `multipart/mixed` requests, extract params from `application/x-www-form-urlencoded` parts and extract IO objects from file parts.

Closes #204

### DIFF
--- a/src/lucky_web/params.cr
+++ b/src/lucky_web/params.cr
@@ -59,6 +59,8 @@ class LuckyWeb::Params
   private def body_param(key : String)
     if json?
       parsed_json.as_h[key]?.try(&.to_s)
+    elsif multipart?
+      multipart_params[key]?
     else
       form_params[key]?
     end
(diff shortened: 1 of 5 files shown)


src/lucky_web/params.cr
```diff
@@ -70,10 +72,35 @@ class LuckyWeb::Params
...
+    if @_multipart_body
+      @_multipart_body
+    else
+      HTTP::Multipart.parse(request) do |headers, io|
+        if headers["Content-Type"] == "application/x-www-form-urlencoded"
+          @_multipart_body = io.gets_to_end
+        end
+      end
+      @_multipart_body
+    end
```
jdbann: This is a mess but the `parse` doesn't return in a useful way.

src/lucky_web/params.cr
```diff
@@ -70,10 +72,35 @@ class LuckyWeb::Params
...
+  end
+
   private def json?
     content_type == "application/json"
   end
 
+  private def multipart?
+    content_type.try { |ct| ct.match /^multipart\// }
```
jdbann: Crystal question - Is there a better way to handle nil-able queries like this?

paulcsmith: Good question. You can use the block shorthand https://crystal-lang.org/docs/syntax_and_semantics/blocks_and_procs.html to make it more concise: `content_type.try(&.match(multipleart regex))`

You can also call `not_nil!` on it but you should only do that if you expect it to ALWAYS be not nil because it raises at runtime. I consider not_nil to be a code smell and try to avoid it if possible.  I hope that helps :)

jdbann:
This isn't how forms work! Will start a new PR using `HTTP::FormData` methods for building and parsing requests.

PR #240, opened 2017-11-22
Add anchor support

### LINKED ISSUE #238: Add "anchor" option to all action route helpers
```
comment_id = 12345
Posts::Show.with(id: "123", anchor: "comment-#{comment_Id}") # "/posts/1#comment-12345
```

actsasflinn:
Adds support for anchors to Action#with, Action#with, and LinkHelpers#link

Closes #238

### DIFF
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -99,6 +99,10 @@ describe Lucky::Action do
       Tests::Update.with("test-id").should eq Lucky::RouteHelper.new(:put, "/tests/test-id")
       Tests::Create.path.should eq "/tests"
       Tests::Create.route.should eq Lucky::RouteHelper.new(:post, "/tests")
+      Tests::Show.path("test-id", anchor: "comment-12345").should eq "/tests/test-id#comment-12345"
+      Tests::Show.with("test-id", anchor: "comment-12345").should eq Lucky::RouteHelper.new(:get, "/tests/test-id#comment-12345")
+      Tests::Show.path("test-id", anchor: "#comment-12345").should eq "/tests/test-id#comment-12345"
+      Tests::Show.with("test-id", anchor: "#comment-12345").should eq Lucky::RouteHelper.new(:get, "/tests/test-id#comment-12345")
     end
 
     it "adds routes to the router" do
--- a/spec/lucky/link_helpers_spec.cr
+++ b/spec/lucky/link_helpers_spec.cr
@@ -45,6 +45,36 @@ private class TestPage
       text "Hello"
     end
   end
+
+  def string_path_with_anchor
+    link "Hello", to: "/foo", anchor: "bar"
+  end
+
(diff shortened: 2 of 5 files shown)


actsasflinn:
Rebased

spec/lucky/link_helpers_spec.cr
```diff
@@ -45,6 +45,36 @@ private class TestPage
...
+
+  def non_get_route_with_options_with_anchor
+    link "Test", to: LinkHelpers::Create, something_custom: "foo", anchor: "bar"
+  end
+
+  def string_path_with_options_with_anchor
+    link "Test", to: "/foos", data_method: "post", anchor: "bar"
+  end
+
+  def get_route_with_block_with_anchor
```
paulcsmith: I don't the link helper need to be modified because you can do `link "Test", to: LinkHelpers::Create.with(anchor: "bar")`. This will keep the link helper simpler and makes it so there is just one way to add an anchor

spec/lucky/action_spec.cr
```diff
@@ -99,6 +99,10 @@ describe Lucky::Action do
       Tests::Update.with("test-id").should eq Lucky::RouteHelper.new(:put, "/tests/test-id")
       Tests::Create.path.should eq "/tests"
       Tests::Create.route.should eq Lucky::RouteHelper.new(:post, "/tests")
```
paulcsmith: Could you add  test for `Tests::Index.with(anchor: "foo")` to test an action that only uses an anchor and no other params?

src/lucky/routeable.cr
```diff
@@ -100,15 +103,18 @@ module Lucky::Routeable
 
       is_root_path = path == ""
       path = "/" if is_root_path
+      if anchor = args[:anchor]?
+        path = "#{path}##{anchor.sub(/^#/,"")}"
```
paulcsmith: I don't think we should sub out `#` because it could be confusing to users. I think they should sub it out themselves if they need. For example:

```crystal
link "Something", to: Posts::Show.with(the_post_id, anchor: "#foo" # Would strip out the #

# In the Posts::ShowPage
div id: "#foo" # The # is not stripped out
```

So I think we should leave the string as-is so there isn't any confusion about how things work.

paulcsmith:
@actsasflinn Just making sure you saw these notes. No rush though :)

actsasflinn:
@paulcsmith definitely saw it but am without my laptop for several days since your comment and will likely be without for a few more days. I’ll take care of it once able.

src/lucky/routeable.cr
actsasflinn: I definitely think they are different but I’m cool to remove it and make the use more clear.

paulcsmith:
Thanks for starting on this @actsasflinn. Just FYI, I started a new PR at https://github.com/luckyframework/lucky/pull/339 since this is something I'll need soon for a project.

PR #246, opened 2017-11-23
Do not run app process in a subshell

### LINKED ISSUE #245: `lucky watch` task starts server in a process's sub shell that doesn't terminate on recompile
I'm using:
- OS Ubuntu 16.04 with bash
- lucky 0.6.2.
- lucky-web ~> 0.5.0

The problem is that the `lucky watch` task wasn't recompiling the app because the port was still in use:
```bash
compiling...
Listening on http://127.0.0.1:8080...
bind: Address already in use (Errno)
0x4e5a37: *CallStack::unwind:Array(Pointer(Void)) at ??
0x6008ba: initialize at /opt/crystal/src/socket/tcp_server.cr 71:15
0x5fee72: initialize:reuse_port at /opt/crystal/src/socket/tcp_server.cr 30:3
0x5fee15: *TCPServer::new:reuse_port<String, Int32, Bool>:TCP Server at ??
0x5f6258: bind at /opt/crystal/src/http/server.cr 154:5
0x5f6182: listen at /opt/crystal/src/http/server.cr 161:5
0x5f612f: listen at /opt/crystal/src/http/server.cr 160:3
```
The cause is this method in the Sentry module:

```crystal
private def create_app_processes
  @app_processes.clear
  result = @run_commands.each do |command|
    @app_processes << Process.new(command, shell: true, output: true, error: true)
  end
  ...
end
```
The `shell: true` option starts a sub shell process and adds it to `@app_processes`. The sub shell in turn starts the server in a new process by running  `./server`. This is not added to `@app_processes`.

So when the processes are killed:
```crystal
def start_app
  @app_processes.each do |process|
    process.kill unless process.terminated?
  end
  ...
end
```
only the shell is killed, not the server. I'm guessing the shell is supposed to kill it's child processes when it's terminated, but for some reason it's not doing so on my machine.

The solution for me was to set `shell: false`.

mikeeus:
Closes #245

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -27,7 +27,7 @@ module Sentry
     private def create_app_processes
       @app_processes.clear
       result = @run_commands.each do |command|
-        @app_processes << Process.new(command, shell: true, output: true, error: true)
+        @app_processes << Process.new(command, shell: false, output: true, error: true)
       end
 
       @successful_compilations += 1


paulcsmith:
I tried this out on a couple projects and it seems to work ok. Thanks for catching this and finding the fix!

mikeeus:
I'm glad it's working because it made the dev server unusable for me. I'm surprised it was only me that was effected though, am I the only one on Ubuntu?

PR #248, opened 2017-11-24
Add 'style' tag to HTMLPage. Closes #91

### LINKED ISSUE #91: Add `style` tag helper

hanneskaeufler:
This allows you to specify raw css which will be output
into a style tag.

```crystal
style "body { background: blue; }
```

I am opening this early. Still debating whether to add a `block`ed version of this to do

```crystal
style do |css|
  css <<-CSS
  body { background: blue; }
  CSS
end
```

### DIFF
--- a/spec/lucky/base_tags_spec.cr
+++ b/spec/lucky/base_tags_spec.cr
@@ -25,6 +25,14 @@ describe Lucky::BaseTags do
     view.para(MySpecialClass.new).to_s.should contain "<p>it works</p>"
     view.para(1_i64).to_s.should contain "<p>1</p>"
   end
+
+  describe "#style" do
+    it "renders a style tag" do
+      view.style("body { font-size: 2em; }").to_s.should contain <<-HTML
+      <style>body { font-size: 2em; }</style>
+      HTML
+    end
+  end
 end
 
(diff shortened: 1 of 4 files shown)


hanneskaeufler:
I'm trying to :) But it also gets me thinking, when would you _not_ want raw output? Seeing as each css declaration is inside `{}` the curlies I don't see the point in calling `raw`?

paulcsmith:
@hanneskaeufler After thinking about it...you're totally right. And you can accomplish the same thing with what you have already:

```crystal
style <<-CSS
  .all-the-css {}
CSS
```

So let's actually leave off the block version until someone comes up with a valid use case. Does that sound good to you?

hanneskaeufler:
Sounds good to me. Been looking for the syntax you just posted all along, I really like that. Btw I don't think circle runs on this PR?

I changed the `WelcomePage` to use this helper. Didn't really run this, couldn't figure out how to use the local version of lucky to create a new test project to render the WelcomePage.

paulcsmith:
Hmm. Not sure why CI isn't running. I pulled it locally and specs pass so I'll merge and investigate CI after. This looks great

PR #251, opened 2017-11-25
Escape paramters passed to actions via route helpers

### LINKED ISSUE #81: Escape params passed to `path` and `route`

hanneskaeufler:
This would close #81. Not quite sure if this is the best place to solve that.


spec/lucky/action_spec.cr
```diff
@@ -101,6 +101,17 @@ describe Lucky::Action do
       Tests::Create.route.should eq Lucky::RouteHelper.new(:post, "/tests")
     end
 
+    it "escapes path params" do
+      Tests::Edit.path("test/id").should eq "/tests/test%2Fid/edit"
+      Tests::Edit.with("test/id").should eq Lucky::RouteHelper.new(:get, "/tests/test%2Fid/edit")
```
paulcsmith: I think this test needs just these two lines. The rest can be deleted

hanneskaeufler: True, rest was repetitive 🏁

PR #255, opened 2017-11-26
Accept multipart forms

### LINKED ISSUE #204: Add multipart form upload support

jdbann:
Accept `multipart/form-data` requests, extracts params and Tempfile objects.

Also adds `file_input` helper.

Demo of posting and logging params and files from multipart requests here: https://github.com/jdbann/lucky_file_demo

Closes #204


spec/lucky/params_spec.cr
```diff
@@ -148,4 +169,85 @@ describe Lucky::Params do
       params.nested(:missing).should eq({} of String => String)
     end
   end
+
+  describe "get_file" do
+    it "gets files" do
+      request = build_multipart_request file_parts: {
+        "welcome" => "hello paul"
```
paulcsmith: Could we make it more clear what each of these represent by doing something like "welcome_file" => "hello-paul.png" or something? What do you think about that?

spec/lucky/params_spec.cr
```diff
@@ -148,4 +169,85 @@ describe Lucky::Params do
...
+
+      params.get_file(:missing).should be_nil
+    end
+  end
+
+  describe "nested_file" do
+    it "gets multipart nested params" do
+      request = build_multipart_request file_parts: {
+        "user" => {
+          "avatar" => "an image"
```
paulcsmith: Same here. Maybe "an-image.png"?

spec/support/context_helper.cr
```diff
@@ -15,6 +15,56 @@ module ContextHelper
...
+    end
+  end
+
+  private def multipart_file_part(
+    formdata : HTTP::FormData::Builder, name : String, value : String
+  )
+    file_io = IO::Memory.new(value)
+    metadata = HTTP::FormData::FileMetadata.new(filename: name)
+    headers = HTTP::Headers{"Content-Type" => "text/plain"}
+    formdata.file(name, file_io, metadata, headers)
```
paulcsmith: You can run `crystal tool format` to make sure these are all formatted correctly. They may already be fine, but might be worth running just in case

spec/support/context_helper.cr
```diff
@@ -15,6 +15,56 @@ module ContextHelper
...
+  end
+
+  private def multipart_file_part(
+    formdata : HTTP::FormData::Builder, name : String, value : Hash(String, String)
+  )
+    value.each do |key, nested_value|
+      nested_name = name + ":" + key
+      multipart_file_part(formdata, nested_name, nested_value)
+    end
+  end
```
paulcsmith: I like all the test helpers you wrote 👍

src/lucky/params.cr
```diff
@@ -70,10 +107,43 @@ class Lucky::Params
     @_form_params ||= HTTP::Params.parse(body)
   end
 
+  @_multipart_params : Hash(String, String)?
+
+  private def multipart_params : Hash(String, String)
+    @_multipart_params ||= parse_multipart_request[0]
```
paulcsmith: You could use `.first` here instead.

src/lucky/params.cr
```diff
@@ -70,10 +107,43 @@ class Lucky::Params
...
+  end
+
+  @_multipart_files : Hash(String, Tempfile)?
+
+  private def multipart_files : Hash(String, Tempfile)
+    @_multipart_files ||=  parse_multipart_request[1]
```
paulcsmith: You *might* be able to use `second`. I'm not sure if Crystal has that method, but worth a try

src/lucky/params.cr
```diff
@@ -70,10 +107,43 @@ class Lucky::Params
...
+  end
+
+  private def parse_multipart_request
+    multipart_params = {} of String => String
```
paulcsmith: Might be worth memoizing this as well:

```crystal
@_multipart_reqeuest ||= begin
  # all the code for parsing the multipart request
end
```

spec/lucky/params_spec.cr
jdbann: Yeah, that makes a lot of sense. Will rename to be clearer. Could also create `Tempfile` objects to pass in here but thought that might clutter up the specs.

spec/support/context_helper.cr
jdbann: I think vim was running it automatically built I'll give it a try before pushing again.

spec/support/context_helper.cr
jdbann: Worth extracting to a `multipart_helper.cr`?

src/lucky/params.cr
jdbann: Worth a try indeed.

jdbann:
Really appreciate the comments @paulcsmith! I'm trying to get the form helper to support `multipart: true` but something isn't quite right and I'm struggling to debug the requests at runtime.

jdbann:
Can now extract params and files from requests. Have tested in an app and all works well.

Still needs wiring into the rest of Lucky but that could be a separate PR if you'd prefer smaller merges.

jdbann:
Hold fire - tempfiles might not be the right approach.

jdbann:
I think the final step is also extracting metadata for the files. Then handling multipart uploads can be ticked off and I'll start on a proposal for how Lucky handles file storage.

spec/lucky/input_helpers_spec.cr
```diff
@@ -70,6 +70,16 @@ describe Lucky::InputHelpers do
     HTML
   end
 
+  it "renders file inputs" do
+    view.file_input(form.first_name).to_s.should contain <<-HTML
+    <input type="file" id="user_first_name" name="user:first_name" value="My name"/>
+    HTML
+
+    view.file_input(form.first_name, class: "cool").to_s.should contain <<-HTML
```
paulcsmith: One there is support in `LuckyRecord` this helper should probably restrict the field to `LuckyRecord::Field(Tempfile)` or something. But for now I think this is good 👍

spec/lucky/params_spec.cr
```diff
@@ -14,6 +14,21 @@ describe Lucky::Params do
...
+        "name" => "Paul",
+        "age"  => "28",
+      },
+    }
+
+    params = Lucky::Params.new(request)
+    dup_params = Lucky::Params.new(request)
+
+    params.nested(:user).should eq({"name" => "Paul", "age" => "28"})
+    dup_params.nested(:user).should eq({"name" => "Paul", "age" => "28"})
```
paulcsmith: I don't think you need to check the contents. You can probably just call `prams.nested(:users)` and `dup_params.nested(:user)` to make it clear that that is all it takes to test that multipart param parsing works

spec/lucky/params_spec.cr
```diff
@@ -79,6 +94,14 @@ describe Lucky::Params do
       params.get(:id).should eq "1"
     end
 
+    it "parses multipart params" do
```
paulcsmith: Maybe "it parses params in multipart requests"?

spec/lucky/params_spec.cr
```diff
@@ -148,4 +184,85 @@ describe Lucky::Params do
...
+      file = params.get_file!(:with)
+      File.read(file.path).should eq "a file"
+    end
+
+    it "raises if missing a param and using get_file! version" do
+      request = build_multipart_request form_parts: {"this" => "that"}
+
+      params = Lucky::Params.new(request)
+
+      expect_raises do
```
paulcsmith: In Crystal 0.24 I believe you need to pass an exception type. Can you add an exception type to this? Something like `expect_raises WhateverTheExceptionIs`

spec/support/context_helper.cr
paulcsmith: Yeah I think that's a great idea!

src/lucky/params.cr
```diff
@@ -70,10 +107,50 @@ class Lucky::Params
...
+  @_parsed_multipart_request : Tuple(Hash(String, String), Hash(String, Tempfile))?
+
+  private def parse_multipart_request
+    @_parsed_multipart_request ||= begin
+      multipart_params = {} of String => String
+      multipart_files = {} of String => Tempfile
+      body_io = IO::Memory.new(body)
+      boundary =
+        HTTP::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
+      HTTP::FormData.parse(body_io, boundary.to_s) do |part|
```
paulcsmith: Maybe this could be extracted to a private method called `parse_form_data` so that it is easier to see what this method does?

src/lucky/params.cr
```diff
@@ -70,10 +107,50 @@ class Lucky::Params
...
+        case part.headers
+        when .includes_word?("Content-Disposition", "filename")
+          part_file = Tempfile.open(part.name) do |tempfile|
+            IO.copy(part.body, tempfile)
+          end
+          multipart_files[part.name] = part_file
```
jdbann: @paulcsmith I've been thinking about handling metadata. I could, at this point, extract all of the available metadata to `model:#{file_param_name}_content_type` etc. params. This would then automatically populate the model appropriately.

spec/lucky/input_helpers_spec.cr
jdbann: Noted

spec/lucky/params_spec.cr
jdbann: Fair enough, I've also removed the contents checks from https://github.com/luckyframework/lucky/pull/255/files/1f0c4c289a0507b6f075db59d8cf915813d331b0#diff-8cbcd6e8b8829525283cc450f05a0166R6

Only concern is if the second attempt ever returned a blank string? If I understand things correctly, `gets_to_end` on an IO stream will initially return the contents of the IO, but a second call will continue from where it left off giving a blank string.

spec/lucky/params_spec.cr
jdbann: 👍

spec/lucky/params_spec.cr
jdbann: They're just generic `Exception` at the moment but I've added that and the error messages.

src/lucky/params.cr
jdbann: I've had a go at this but I might have misunderstood.

Xosmond:
What is the state of this PR?

jdbann:
I'll update the PR to address the conflicts, but otherwise it's ready for @paulcsmith. I think he's been focusing on v1 requirements.

Xosmond:
@jdbann processing, backgrounding, direct upload and storage are not covered right? Its only local storage?

jdbann:
@Xosmond Yep. This PR is just to parse multipart forms. It doesn't even cover local storage yet, just creates tempfiles.

Next step is to work on `LuckyRecord` to actually persist the files. I'll aim to cover local storage and S3 but with a nice interface for others to add extra storage adapters.

There isn't really a job system in Lucky to my knowledge so processing and backgrounding might be blocked until that's sorted. But I know @paulcsmith prefers direct upload so that will probably be discussed soon as well.

paulcsmith:
Thanks for rebasing and covering the feedback! I'll have some time Friday to check this out more thoroughly. I'm excited to get this in!

And yeah, I'm focusing a lot on getting some stuff into LuckyRecord and LuckyMigrator that we've needed in our projects, so haven't had much time to focus on file upload, but it's definitely something that is needed soon 👍 

For background processing, there is sidekiq.cr and faktory.cr. I haven't used either, but I've heard sidekiq.cr is pretty good.

Xosmond:
Great!

Issue #83, opened 2017-06-30
Add `lucky gen.action` for generating basic action


paulcsmith:
`lucky gen.action Users::Index`

paulcsmith:
Closed by https://github.com/luckyframework/lucky/pull/264

PR #264, opened 2017-12-01
Add `lucky gen.action` to generate a basic action

ianagne:
Add a task to generate an action given the name of the action (e.g.
Users::Index).  Of note, this task generates the class skeleton but does
not generate methods within the class.

See issue #83

### DIFF
--- a/spec/tasks/gen/action_spec.cr
+++ b/spec/tasks/gen/action_spec.cr
@@ -33,5 +33,18 @@ describe Gen::Action do
 
     io.to_s.strip.should eq(message)
   end
-end
 
+  it "displays an error if given only one class" do
+    begin
+      io = IO::Memory.new
+      ARGV.push("Users")
+
+      Gen::Action.new.call(io)
+      message = "\e[31mThat's not a valid Action.  Example: lucky gen.action Users::Index\e[0m"
+
+      io.to_s.strip.should eq(message)
+    ensure
+      ARGV.pop
+    end
+  end
+end
--- a/spec/tasks/gen/action_spec.cr
+++ b/spec/tasks/gen/action_spec.cr
@@ -1,0 +1,37 @@
+require "../../spec_helper"
+
+describe Gen::Action do
+  it "generates a basic action" do
+    begin
+      ARGV.push("Users::Index")
+      Gen::Action.new.call
+
+      File.exists?("src/actions/users/index.cr").should be_true
+    ensure
+      ARGV.pop
+      FileUtils.rm_rf "src/actions/users/index.cr"
+    end
+  end
+
+  it "generates a nested action" do
+    begin
+      ARGV.push("Users::Announcements::Index")
+      Gen::Action.new.call
+
+      File.exists?("src/actions/users/announcements/index.cr").should be_true
+    ensure
+      ARGV.pop
+      FileUtils.rm_rf "src/actions/users/announcements/index.cr"
+    end
+  end
+
+  it "displays an error if given no arguments" do
+    io = IO::Memory.new
+
+    Gen::Action.new.call(io)
+    message = "\e[31mAction name is required. Example: lucky gen.action Users::Index\e[0m"
+
+    io.to_s.strip.should eq(message)
+  end
+end
+
--- a/tasks/gen/action.cr
+++ b/tasks/gen/action.cr
@@ -1,4 +1,3 @@
-require "lucky_cli"
 require "colorize"
 require "ecr"
 require "file_utils"
--- a/tasks/gen/action.cr
+++ b/tasks/gen/action.cr
@@ -1,0 +1,54 @@
+require "lucky_cli"
+require "colorize"
+require "ecr"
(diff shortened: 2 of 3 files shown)


tasks/gen/action.ecr
```diff
@@ -0,0 +1,2 @@
+class <%= name %> < BrowserAction
+end
```
paulcsmith: I think it would be good to add the `action do/end` here. I think that's the most common thing to use, and it's easy to add `nested_` if you want a nested action instead.

```crystal
class <%= name %> < BrowserAction
  action do
  end
end
```

spec/tasks/gen/action_spec.cr
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  it "displays an error if given no arguments" do
+    io = IO::Memory.new
+
+    Gen::Action.new.call(io)
+    message = "\e[31mAction name is required. Example: lucky gen.action Users::Index\e[0m"
+
+    io.to_s.strip.should eq(message)
+  end
```
paulcsmith: It's unlikely to happen, but it might be worth throwing an error if there is only one class too: `lucky gen.action Users`

tasks/gen/action.cr
```diff
@@ -0,0 +1,53 @@
...
+class Lucky::ActionGenerator
+  getter :name
+
+  ECR.def_to_s "#{__DIR__}/action.ecr"
+
+  def initialize(@name : String)
+  end
+
+  def generate
+    make_folders_if_missing
```
paulcsmith: I love this method name!

tasks/gen/action.cr
```diff
@@ -0,0 +1,53 @@
...
+  private def path
+    path_args[0..-2].join("/")
+  end
+
+  private def action
+    path_args.last
+  end
+
+  private def filename
+    Dir.current + "/src/actions/#{path}/#{action}.cr"
```
paulcsmith: I like how you handled this 👍

tasks/gen/action.cr
```diff
@@ -0,0 +1,53 @@
+require "colorize"
+require "ecr"
+require "file_utils"
+
+class Lucky::ActionGenerator
+  getter :name
+
+  ECR.def_to_s "#{__DIR__}/action.ecr"
```
paulcsmith: Instead of using ECR, could you switch this to using teeplate? It's included with LuckyCLI and it's what is used elsewhere to generate files. 

* https://github.com/mosop/teeplate
* Example in LuckyCLI: https://github.com/luckyframework/lucky_cli/blob/master/src/lucky_cli/src_template.cr

tasks/gen/action.cr
```diff
@@ -0,0 +1,53 @@
...
+  private def contents
+    to_s
+  end
+end
+
+class Gen::Action < LuckyCli::Task
+  banner "Generate a new action"
+
+  def call(io : IO = STDOUT)
+    if ARGV.first? == nil
```
paulcsmith: You can use `.nil?` to shorten this up a bit

tasks/gen/action.ecr
ianagne: Totally agree.  I misread the docs and thought you needed to explicitly enter the path in the action (e.g. `def "users/:user_id/announcements"`).  It's great that Lucky figures that out for you.

tasks/gen/action.cr
ianagne: Great idea!

tasks/gen/action.cr
ianagne: I stole it from your migration generator 😛

paulcsmith: Lol :P

spec/tasks/gen/action_spec.cr
ianagne: I personally love when CLI's throw helpful errors when I'm learning a new framework.  I'll add it in.

tasks/gen/action.cr
ianagne: Excellent.  I saw that Amber is using teeplate and it looks pretty cool.

tasks/gen/action.ecr
paulcsmith: Yep! As long as it is a resourceful action :) The generator could probably be smarter in the future and check for a resourceful action name, and if it isn't, then it would leave a comment in the template asking the user to declare their own custom route:

```crystal
class Users::Pricing < BrowserAction
  # This is not a RESTful action, so you'll need to declare your own route
  # https://luckyframework.org/guides/actions-and-routing/#routing
end
```

But that can come later

tasks/gen/action.cr
paulcsmith: Yeah it's super nice. I didn't want to add it as a dependency on LuckyMigrator, but since Lucky CLI generates a lot of files, it made sense to include it :)

paulcsmith:
Awesome :D

PR #268, opened 2017-12-03
Generate Action contents with Teeplate

ianagne:
The Action generator was halfway between using ECR and Teeplate to
render the contents of the Action file.   As such, when generating a new
Action, the file path and name were correct, but file itself just
contained the stringified ActionGenerator object (e.g.
<ActionGenerator:312>).

Switch to using the Teeplate `#render` method on ActionGenerator (now
ActionTemplate) to create a new Action.

### DIFF
--- a/spec/tasks/gen/action_spec.cr
+++ b/spec/tasks/gen/action_spec.cr
@@ -3,25 +3,29 @@ require "../../spec_helper"
 describe Gen::Action do
   it "generates a basic action" do
     begin
-      ARGV.push("Users::Index")
+      valid_action_name = "Users::Index"
+      ARGV.push(valid_action_name)
+
       Gen::Action.new.call
 
-      File.exists?("src/actions/users/index.cr").should be_true
+      File.read("./src/actions/users/index.cr").
+        should contain(valid_action_name)
     ensure
-      ARGV.pop
-      FileUtils.rm_rf "src/actions/users/index.cr"
+      cleanup
     end
   end
 
   it "generates a nested action" do
     begin
-      ARGV.push("Users::Announcements::Index")
+      valid_nested_action_name = "Users::Announcements::Index"
+      ARGV.push(valid_nested_action_name)
+
       Gen::Action.new.call
 
-      File.exists?("src/actions/users/announcements/index.cr").should be_true
+      File.read("src/actions/users/announcements/index.cr").
+        should contain(valid_nested_action_name)
     ensure
-      ARGV.pop
-      FileUtils.rm_rf "src/actions/users/announcements/index.cr"
+      cleanup
     end
   end
 
@@ -37,14 +41,20 @@ describe Gen::Action do
   it "displays an error if given only one class" do
     begin
       io = IO::Memory.new
-      ARGV.push("Users")
+      invalid_action_name = "Users"
+      ARGV.push(invalid_action_name)
 
       Gen::Action.new.call(io)
       message = "\e[31mThat's not a valid Action.  Example: lucky gen.action Users::Index\e[0m"
 
       io.to_s.strip.should eq(message)
     ensure
-      ARGV.pop
+      ARGV.clear
     end
   end
+end
+
+def cleanup
+  ARGV.clear
+  FileUtils.rm_rf("./src/actions")
 end
--- a/spec/tasks/gen/action_spec.cr
+++ b/spec/tasks/gen/action_spec.cr
@@ -3,25 +3,37 @@ require "../../spec_helper"
 describe Gen::Action do
   it "generates a basic action" do
     with_cleanup do
+      io = IO::Memory.new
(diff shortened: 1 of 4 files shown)


tasks/gen/action.cr
```diff
@@ -63,4 +35,24 @@ class Gen::Action < LuckyCli::Task
     @error = "That's not a valid Action.  Example: lucky gen.action Users::Index"
```
paulcsmith: I didn't notice this before, but there are 2 spaces between Action and Example

paulcsmith:
One more thought, it might be nice to say where the action was generated:

```
Done generating Users::Index in src/actions/users/index.cr
```

ianagne:
I also totally thought the last PR was ready to merge.  I didn't realize I used Teeplate wrong until I started working on other generators.

spec/tasks/gen/action_spec.cr
```diff
@@ -37,14 +41,20 @@ describe Gen::Action do
...
 
       io.to_s.strip.should eq(message)
     ensure
-      ARGV.pop
+      ARGV.clear
     end
   end
 end
+
+def cleanup
```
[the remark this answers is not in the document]
ianagne: Nice, I like it.  It looks a lot cleaner than using begin/ensure inside of the tests.

tasks/gen/action.cr
ianagne: Good catch.

ianagne:
I like the idea of adding a success message, especially using colorize to point out the important parts. 👍

spec/tasks/gen/action_spec.cr
```diff
@@ -2,26 +2,38 @@ require "../../spec_helper"
...
+    with_cleanup do
+      io = IO::Memory.new
+      valid_action_name = "Users::Index"
+      ARGV.push(valid_action_name)
+
+      Gen::Action.new.call(io)
+      expected_path = "/src/actions/users"
+      expected_message =
+        "Done generating #{valid_action_name.colorize(:green)}" \
+        " in #{expected_path.colorize(:green)}"
```
paulcsmith: Maybe instead of testing the exact message you can use `contain` to test that is has the right path and action:

```crystal
io.to_s.should contain(valid_action_name)
io.to_s.should contain("/src/actions/users")
```

That way we can change the message without breaking the test

spec/tasks/gen/action_spec.cr
```diff
@@ -2,26 +2,38 @@ require "../../spec_helper"
...
+
+      Gen::Action.new.call(io)
+      expected_path = "/src/actions/users/announcements"
+      expected_message =
+        "Done generating #{valid_nested_action_name.colorize(:green)}" \
+        " in #{expected_path.colorize(:green)}"
+
+      File.read("src/actions/users/announcements/index.cr").
+        should contain(valid_nested_action_name)
+      io.to_s.strip.should eq(expected_message)
```
paulcsmith: Same thing here about using `contain` so the test isn't as coupled to the message

spec/tasks/gen/action_spec.cr
```diff
@@ -35,16 +47,28 @@ describe Gen::Action do
...
       io = IO::Memory.new
-      ARGV.push("Users")
+      invalid_action_name = "Users"
+      ARGV.push(invalid_action_name)
 
       Gen::Action.new.call(io)
-      message = "\e[31mThat's not a valid Action.  Example: lucky gen.action Users::Index\e[0m"
+      message = "\e[31mThat's not a valid Action. Example: lucky gen.action Users::Index\e[0m"
 
       io.to_s.strip.should eq(message)
```
paulcsmith: You could use `contain` here as well so you don't need the color stuff at the beginning and end. Maybe something like: `io.to_s.should contain "That's not a valid Action"`

tasks/gen/action.cr
```diff
@@ -60,7 +33,35 @@ class Gen::Action < LuckyCli::Task
...
+  private def path
+    path_args[0..-2].join("/")
+  end
+
+  private def path_args
+    action_name.split("::").map(&.downcase)
+  end
+
+  private def success_message
+    "Done generating #{action_name.colorize(:green)} in #{app_directory_path.colorize(:green)}"
```
paulcsmith: Can you add a `.` before the `app_directory_path` so it's clear it's not an absolute path?

spec/tasks/gen/action_spec.cr
ianagne: I'm a huge fan of using contain throughout the specs for these messages. 👍

tasks/gen/action.cr
ianagne: 💯

paulcsmith:
Looks fantastic. Thanks!

PR #279, opened 2017-12-06
Extract #with_cleanup into helper

ianagne:
Each of the generator tasks output files to a different directory.
Extract the #with_cleanup method into a spec support helper and
generalize it to handle outputting to different directories.


ianagne:
I'm currently working on a PR for the model generator and some sort of cleanup helper would be userful since each of the generators output to a different directory.  Thoughts?

paulcsmith:
I love it! Nicely done

Issue #84, opened 2017-06-30
Add `lucky gen.model User`


paulcsmith:
Generates a base query, form, user model and migration

```crystal
lucky gen.model User

# generates
# src/models/user.cr
class User < BaseModel
  # Use inflector to get the pluralized table name
  table :users do
  end
end

# and a new migration called CreateUsers
# Can use the Gen::Migration class: https://github.com/luckyframework/lucky_migrator/blob/master/src/lucky_migrator/tasks/gen/migration.cr
```

Later on add support for creating a model and migration with fields: `lucky gen.model User name:String age:Int32`

paulcsmith:
Closed by https://github.com/luckyframework/lucky/pull/281

PR #281, opened 2017-12-07
Add `lucky gen.model` to generate a model

ianagne:
Add a task to generate a model, form, and query when given a model name
(e.g. User).  The task will generate the initial class scaffolds, but
does not yet accept attributes as arguments.

See issue #84


tasks/gen/model.cr
```diff
@@ -0,0 +1,18 @@
+class Gen::Model < LuckyCli::Task
+  banner "Generate a model, query, and form"
+
+  def call(io : IO = STDOUT)
+    if ARGV.first?
+      Lucky::ModelTemplate.new(model_name).render("./src/models/")
```
ianagne: I'll be adding a few success messages here similar to the `gen.action` message.

tasks/gen/templates/model/{{name}}.cr.ecr
```diff
@@ -0,0 +1,4 @@
+class <%= @name.capitalize %> < BaseModel
+  table :<%= @pluralized_name %> do
```
ianagne: TIL Crystal doesn't provide an option to convert a String to a Symbol, hence the weird interpolation here.

ianagne:
I was going to call the migration generator as well, but it doesn't seem to be available in this project.    I could port over the migration generator using Teeplate if that would be desirable.  I could also add lucky_migrator to the shard.yml.

spec/tasks/gen/model_spec.cr
```diff
@@ -0,0 +1,36 @@
+require "../../spec_helper"
+
+include CleanupHelper
+
+describe Gen::Model do
+  it "generates a model" do
+    with_cleanup do
+      io = IO::Memory.new
+      model_name = "User"
```
paulcsmith: Could you change the model name to `ContactInfo`? That way we can test that it creates the correct folder name `contact_info`.

tasks/gen/model.cr
```diff
@@ -0,0 +1,29 @@
...
+      io.puts "Model name is required. Example: lucky gen.model User".colorize(:red)
+    end
+  end
+
+  private def render_templates
+    Lucky::ModelTemplate.new(model_name).render("./src/models/")
+    io.puts success_message("#{model_name}.cr", "./src/models/")
+    Lucky::FormTemplate.new(model_name).render("./src/forms/")
+    io.puts success_message("#{model_name}_form.cr", "./src/forms/")
+    Lucky::QueryTemplate.new(model_name).render("./src/queries/")
```
paulcsmith: Instead of having 3 separate template classes, I think you could have one `ModelTemplate` class that has all 3 files:

```
templates/
  model/
    src/models/{{model_name}}.cr
    src/forms/{{model_name}}_form.cr
    src/forms/{{model_name}}_query.cr
```

ianagne: It should definitely be possible with Teeplate.  My original intention was to have different templates that could be re-used for individual generator methods (e.g. `gen.form`), but I think that may have been a bit premature.  Using a single template here should make it easier to parse attribute arguments (e.g. `name:string` `email:string`) in a future PR.

spec/tasks/gen/model_spec.cr
ianagne: Great idea 👍

spec/tasks/gen/model_spec.cr
```diff
@@ -0,0 +1,36 @@
...
+      io = IO::Memory.new
+      model_name = "ContactInfo"
+      ARGV.push(model_name)
+
+      Gen::Model.new.call(io)
+
+      File.read("./src/models/contact_info.cr").
+        should contain("class ContactInfo < BaseModel")
+      io.to_s.should contain("contact_info.cr")
+      io.to_s.should contain("./src/models")
```
paulcsmith: Could these expectations be combined to “./src/models/comtact_info.cr”?

spec/tasks/gen/model_spec.cr
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  it "displays an error if given no arguments" do
+    io = IO::Memory.new
+
+    Gen::Model.new.call(io)
+
+    io.to_s.should contain("Model name is required.")
+  end
+end
```
paulcsmith: What do you think about adding an error and a suggestion if they don’t have a camelcased name?

https://github.com/luckyframework/lucky_migrator/commit/722fb00b44f6490e94c0d37be6cfff48b41374da

tasks/gen/model.cr
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private def display_success_messages
+    io.puts success_message("#{underscored_name}.cr", "./src/models/")
+    io.puts success_message("#{underscored_name}_form.cr", "./src/forms/")
+    io.puts success_message("#{underscored_name}_query.cr", "./src/queries/")
+  end
+
+  private def success_message(file, output_path)
+    "Done generating #{file.colorize(:green)} in #{output_path.colorize(:green)}"
```
paulcsmith: What do you think about shortening this to “Generated...”

ianagne: 👍

spec/tasks/gen/model_spec.cr
ianagne: Yeah, I think we can combine the expectation if we change the success message to be something like:
"Generated ./src/models/contact_info.cr"

spec/tasks/gen/model_spec.cr
ianagne: Good idea.  I'll add a regex matcher that ensures a capital letter and the absence of dashes and underscores.

paulcsmith: You could do what I did in migrator as well. Might be simpler/more accurate `model.camel_case != model`

spec/tasks/gen/model_spec.cr
paulcsmith: Oh I thought it would work without changing the success messages since it container the whole path, but maybe I misread the success message?

ianagne: It currently contains the path and filename separately.  The message would be something like "Done generating contact_info.cr in ./src/models/"

spec/tasks/gen/model_spec.cr
ianagne: 👍 .  I am continually impressed by what Crystal has in the standard library.

spec/tasks/gen/model_spec.cr
paulcsmith: Oh I see! How about, "Generated ContactInfo in ./src/model/contact_info.cr"?

spec/tasks/gen/model_spec.cr
paulcsmith: Me too! It's a nice language :)

spec/tasks/gen/model_spec.cr
ianagne: Sounds like a plan. Getting the model name in there too will be a nice touch.

tasks/gen/model.cr
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  private def model_name_is_present
+    @error = "Model name is required. Example: lucky gen.model User"
+    ARGV.first?
+  end
+
+  private def model_name_is_camelcase
+    @error = "Model name should be camel case. Example: lucky gen.model ContactInfo"
+    model_name.camelcase == model_name
```
paulcsmith: Could you use the model name in the error message, “lucky gen.model #{model_name.camelcase}”

spec/tasks/gen/model_spec.cr
```diff
@@ -0,0 +1,44 @@
...
+
+  it "displays an error if given no arguments" do
+    io = IO::Memory.new
+
+    Gen::Model.new.call(io)
+
+    io.to_s.should contain("Model name is required.")
+  end
+
+  it "displays an error if arugment is not camelcase" do
```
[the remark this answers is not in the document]
ianagne: 👍

tasks/gen/model.cr
ianagne: Great idea

paulcsmith:
Looks great. Thanks!

PR #285, opened 2017-12-07
Add optional params

hanneskaeufler:
Fleshing this out step by step, hope no one else is currently working on this #116, but I could use this for my blog lol.

Currently this already works:
```crystal
class Blog::Index < BrowserAction
  optional_param page, default: "1"

  get "/" do
    posts = PostQuery.new.latest(page: page.to_i).results
    render Blog::IndexPage, posts: posts
  end
end
```

### DIFF
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -85,9 +85,10 @@ end
 class OptionalParams::Index < Lucky::Action
   param page : Int32?
   param with_default : String? = "default"
+  param with_int_default : Int32? = 1
 
   action do
-    render_text "optional param: #{page}"
+    render_text "optional param: #{page} #{with_int_default}"
   end
 end
 
@@ -164,12 +165,12 @@ describe Lucky::Action do
 
     it "is fetched if present" do
       action = OptionalParams::Index.new(build_context(path: "/?page=3"), params)
-      action.page.should eq "3"
+      action.page.should eq 3
     end
 
     it "can be used within the action" do
       response = OptionalParams::Index.new(build_context(path: "/?page=3"), params).call
-      response.body.to_s.should eq "optional param: 3"
+      response.body.to_s.should eq "optional param: 3 1"
     end
 
     it "can specify a default value" do
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -82,6 +82,14 @@ class PlainText::Index < Lucky::Action
   end
 end
 
+class OptionalParams::Index < Lucky::Action
+  optional_param page
+
+  action do
+    render_text "optional param: #{page}"
+  end
+end
+
 describe Lucky::Action do
   describe "routing" do
     it "creates URL helpers for the resourceful actions" do
@@ -144,6 +152,23 @@ describe Lucky::Action do
     it "can get params" do
       action = PlainText::Index.new(build_context(path: "/?q=test"), params)
       action.params.get(:q).should eq "test"
+    end
+  end
+
+  describe "optional params" do
+    it "is initialized to nil" do
+      action = OptionalParams::Index.new(build_context(path: ""), params)
+      action.page.should eq nil
+    end
+
+    it "is fetched if present" do
+      action = OptionalParams::Index.new(build_context(path: "/?page=3"), params)
+      action.page.should eq "3"
+    end
+
+    it "can be used within the action" do
+      response = OptionalParams::Index.new(build_context(path: "/?page=3"), params).call
+      response.body.to_s.should eq "optional param: 3"
     end
   end
 end
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -84,6 +84,7 @@ end
 
 class OptionalParams::Index < Lucky::Action
   optional_param page
+  optional_param with_default, default: "default"
 
   action do
     render_text "optional param: #{page}"
@@ -169,6 +170,11 @@ describe Lucky::Action do
     it "can be used within the action" do
       response = OptionalParams::Index.new(build_context(path: "/?page=3"), params).call
       response.body.to_s.should eq "optional param: 3"
+    end
+
+    it "can specify a default value" do
+      action = OptionalParams::Index.new(build_context(path: ""), params)
+      action.with_default.should eq "default"
     end
   end
 end
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -82,6 +82,14 @@ class PlainText::Index < Lucky::Action
   end
 end
 
+class RequiredParams::Index < Lucky::Action
+  param required_page : Int32
+
+  action do
+    render_text "required param: #{required_page}"
+  end
+end
+
 class OptionalParams::Index < Lucky::Action
   param page : Int32?
   param with_default : String? = "default"
@@ -156,6 +164,16 @@ describe Lucky::Action do
       action = PlainText::Index.new(build_context(path: "/?q=test"), params)
       action.params.get(:q).should eq "test"
     end
+
+    it "can get manually defined required params" do
+      action = RequiredParams::Index.new(build_context(path: "/?required_page=1"), params)
+      action.required_page.should eq 1
+    end
+
+    it "raises for missing required params" do
+      action = RequiredParams::Index.new(build_context(path: ""), params)
+      expect_raises(Lucky::Exceptions::MissingParam) { action.required_page }
+    end
   end
 
   describe "optional params" do
@@ -177,6 +195,11 @@ describe Lucky::Action do
     it "can specify a default value" do
       action = OptionalParams::Index.new(build_context(path: ""), params)
       action.with_default.should eq "default"
+    end
+
+    it "overrides the default if present" do
+      action = OptionalParams::Index.new(build_context(path: "/?with_int_never_nil=42"), params)
+      action.with_int_never_nil.should eq 42
     end
 
     it "is added as optional argument to the path" do
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -179,14 +179,12 @@ describe Lucky::Action do
 
     it "is added as optional argument to the path" do
       OptionalParams::Index.path(7).should eq "/optional_params?page=7"
-      OptionalParams::Index.path(7).should eq "/optional_params?page=%2F7"
-      OptionalParams::Index.path(7, "other").should eq "/optional_params?page=7&with_default=other"
+      OptionalParams::Index.path(7, "/other").should eq "/optional_params?page=7&with_default=%2Fother"
     end
 
     it "is added as optional argument to the route" do
       OptionalParams::Index.route(7).should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7")
-      OptionalParams::Index.route(7).should eq Lucky::RouteHelper.new(:get, "/optional_params?page=%2F7")
-      OptionalParams::Index.route(7, "other").should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7&with_default=other")
+      OptionalParams::Index.route(7, "/other").should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7&with_default=%2Fother")
     end
   end
 end
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -218,7 +218,7 @@ describe Lucky::Action do
     end
 
     it "raises when we cannot parse the non-optional param into the desired type" do
-      expect_raises Lucky::Exceptions::InvalidParam, "Non-optional param \"with_int_never_nil\" with value \"no_int\" couldn't be parsed to a \"Int32\"" do
+      expect_raises Lucky::Exceptions::InvalidParam, "Required param \"with_int_never_nil\" with value \"no_int\" couldn't be parsed to a \"Int32\"" do
         OptionalParams::Index.new(build_context(path: "/?with_int_never_nil=no_int"), params()).call
       end
     end
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -195,7 +195,7 @@ describe Lucky::Action do
     end
 
     it "raises when we cannot parse the non-optional param into the desired type" do
-      expect_raises Exception, "Non-optional param \"with_int_never_nil\" couldn't be parsed to a \"Int32\"" do
+      expect_raises Lucky::Exceptions::InvalidParam, "Non-optional param \"with_int_never_nil\" with value \"no_int\" couldn't be parsed to a \"Int32\"" do
         OptionalParams::Index.new(build_context(path: "/?with_int_never_nil=no_int"), params).call
       end
     end
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -220,9 +220,10 @@ describe Lucky::Action do
       OptionalParams::Index.route(page: 7, with_default: "/other").should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7&with_default=%2Fother")
     end
 
-    it "doesnt raise when we cannot parse the optional param into the desired type" do
-      response = OptionalParams::Index.new(build_context(path: "/?page=no_int"), params).call
-      response.body.to_s.should eq "optional param:  1 1337"
+    it "raises when the optional param cannot be parsed into the desired type" do
+      expect_raises Lucky::Exceptions::InvalidParam do
+        response = OptionalParams::Index.new(build_context(path: "/?page=no_int"), params).call
+      end
     end
 
     it "raises when we cannot parse the non-optional param into the desired type" do
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -83,8 +83,8 @@ class PlainText::Index < Lucky::Action
 end
 
 class OptionalParams::Index < Lucky::Action
-  optional_param page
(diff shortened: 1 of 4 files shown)


spec/lucky/action_spec.cr
```diff
@@ -146,6 +155,28 @@ describe Lucky::Action do
...
+      action = OptionalParams::Index.new(build_context(path: ""), params)
+      action.page.should eq nil
+    end
+
+    it "is fetched if present" do
+      action = OptionalParams::Index.new(build_context(path: "/?page=3"), params)
+      action.page.should eq "3"
+    end
+
+    it "can be used within the action" do
```
hanneskaeufler: I dont think this adds anything actually ...

src/lucky/routeable.cr
```diff
@@ -115,4 +115,16 @@ module Lucky::Routeable
       \{% raise "Use `route` instead of `with` if the action doesn't need params" %}
     end
   end
+
+  macro included
```
hanneskaeufler: Is this explained anywhere in the crystal docs? I THINK it means "always run this when 'include Routable' is called?"  so we can use it to initialize macro consts etc., but not really sure on the specifics.

src/lucky/routeable.cr
```diff
@@ -100,6 +117,17 @@ module Lucky::Routeable
...
+
+      query_params = {} of String => String
+      {% for opt_param in OPTIONAL_PARAMS %}
+        query_params["{{ opt_param }}"] = {{ opt_param }} unless {{ opt_param }}.nil?
+      {% end %}
+
+      unless query_params.empty?
+        path += "?"
+        path += HTTP::Params.encode(query_params)
+      end
```
hanneskaeufler: This duplication could possibly be removed by macro generating macros? 🗡

hanneskaeufler:
If anyone wants to take a look, I think this should just work (™)

hanneskaeufler:
Should we be handling the case when someone does the following?
```crystal
optional_param page

get "/:page" do
end
```

src/lucky/routeable.cr
```diff
@@ -79,12 +82,25 @@ module Lucky::Routeable
...
       is_root_path = path == ""
       path = "/" if is_root_path
+
+      query_params = {} of String => String
+      {% for opt_param in OPTIONAL_PARAMS %}
+        query_params["{{ opt_param }}"] = {{ opt_param }} unless {{ opt_param }}.nil?
+      {% end %}
+
+      unless query_params.empty?
+        path += "?#{HTTP::Params.encode(query_params)}"
```
hanneskaeufler: Should we really escape here?
I am seeing what looks like double escaping when I do
```crystal
      link "Next page", to: Blog::Index.with("2")
```

hanneskaeufler: Hmm no that might be weird `link` escaping. This fixes it lol:
```crystal
      raw "<a href=\"#{Blog::Index.path("2")}\">Next page</a>"
```

spec/lucky/action_spec.cr
```diff
@@ -82,6 +82,15 @@ class PlainText::Index < Lucky::Action
   end
 end
 
+class OptionalParams::Index < Lucky::Action
+  optional_param page
+  optional_param with_default, default: "default"
```
paulcsmith: Sorry to tack this on so late, but I think this can be changed to make it more flexible and be a bit more like other parts of Lucky/Crystal. Here's what I'm thinking:

```crystal
param page : Int32 # A required param
param page : Int32? # Not required, defaults to `nil` and is not added to query params if `nil`
param page : Int32 = 1 # Not required because if it is empty, defaults to `1`
```

This will be a bit trickier to implement, but is a bit more intuitive and powerful. It will also allow only the correct types and will parse them for you so you don't have to manually call `to_i`.

### Here is how it works:

LuckyRecord extends a number of built in types so that they know how to parse strings. You can see that in action here: https://github.com/luckyframework/lucky_record/tree/master/src/lucky_record/charms and specifically: https://github.com/luckyframework/lucky_record/blob/bfd25d77a24bae98a321c50eb5601bd77ce299b7/src/lucky_record/charms/int32_extensions.cr#L10-L14

Since Lucky depends on LuckyRecord we can use those extensions to parse string params into types like `Bool`, `Int32`, `String` or `Time`. 

It is a bit tricky to figure out how to do this, so here is a rough example that should show how this might work:

```crystal
# https://crystal-lang.org/api/0.23.1/Crystal/Macros/TypeDeclaration.html
macro page(type_declaration)
  {% PARAM_DECLARATIONS << type_declaration %}
end

param page : Int32?
{% for type_declaration in PARAM_DECLARATIONS %}
  {{ type_declaration.var }} # would be `page`
  {{ type_declaration.var }} # would be a type union. See example: https://github.com/luckyframework/lucky_migrator/blob/be44188ec6da68a0de9e9fc0c9494f8fd8b899d1/src/lucky_migrator/alter_table_statement.cr#L56
  {{ type_declaration.default }} # would be `nil` in this case
  # Parse a param like this
  # Note: type.first is for Union types. If you did `param page : Int32` you would do just `type`
  {{ type_declaration.type.first }}::Lucky.parse!(params.get(:{{ type_declaration.var }}))
{% end %}
```

Feel free to ping me in Gitter or comment here if this is unclear.

src/lucky/routeable.cr
paulcsmith: Here is a nice guide on it. They're called Macro Hooks: https://crystal-lang.org/docs/syntax_and_semantics/macros/hooks.html

It basically copies what's in that macro into the class that is including/inheriting it

spec/lucky/action_spec.cr
hanneskaeufler: Oooh thats interesting. Yep I definetely like your proposed syntax. Moving this into the type system makes so much sense. Can you tell I'm more fluent in dynamic langs? :D :D I will hack on this a bit over the coming days, lets see what we can make of it! Thanks for all your pointers.

src/lucky/routeable.cr
```diff
@@ -115,4 +145,81 @@ module Lucky::Routeable
...
+          else
+            raise "We're screwed here"
+          end
+      {% end %}
+      end
+    end
+  end
+end
+
+struct Int32
```
hanneskaeufler: All of the code below was lifted from `lucky_record` because I couldn't manage to use the code from there. 🙅‍♂️

spec/lucky/action_spec.cr
```diff
@@ -146,6 +157,49 @@ describe Lucky::Action do
...
+      OptionalParams::Index.route(7).should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7")
+      OptionalParams::Index.route(7, "/other").should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7&with_default=%2Fother")
+    end
+
+    it "doesnt raise when we cannot parse the optional param into the desired type" do
+      response = OptionalParams::Index.new(build_context(path: "/?page=no_int"), params).call
+      response.body.to_s.should eq "optional param:  1 1337"
+    end
+
+    it "raises when we cannot parse the non-optional param into the desired type" do
```
hanneskaeufler: This is bollocks actually I don't think we ever want to raise/runtime error when something inside the query params cannot be parsed. We could just return the default value instead duh.

src/lucky/routeable.cr
```diff
@@ -115,4 +145,81 @@ module Lucky::Routeable
...
+
+  macro param(type_declaration)
+    {% PARAM_DECLARATIONS << type_declaration %}
+
+    def {{ type_declaration.var }} : {{ type_declaration.type }}
+      val = params.get(:{{ type_declaration.var.id }})
+      if val.nil?
+        nil || {{ type_declaration.value || nil }}
+      else
+        {% if type_declaration.type.class_name == "Union" %}
```
hanneskaeufler: This is the laymans way of doing a type check for `Union` types lol. There must be a better way? `union?` I can't seem to call even though I see it in the docs.

src/lucky/routeable.cr
```diff
@@ -115,4 +145,81 @@ module Lucky::Routeable
...
+          result = {{ type_declaration.type.types.first }}::LuckyHack.parse(val)
+          if result.is_a? {{ type_declaration.type.types.first }}::LuckyHack::SuccessfulCast
+            result.value
+          else
+            nil
+          end
+        {% else %}
+          result = {{ type_declaration.type }}::LuckyHack.parse(val)
+          if result.is_a? {{ type_declaration.type }}::LuckyHack::SuccessfulCast
```
hanneskaeufler: We could remove that check if `FailedCast` would have `def value; nil; end` right?

paulcsmith:
I'll try to take a closer look soon, but here is the commit I did to get it passing without hacks: https://github.com/luckyframework/lucky/commit/18c64adfc5fe5d4ed2862d8e9ecb1381f40e78ba

I hope that's helpful :D

hanneskaeufler:
Thanks, it pointed me to the fact that I had to `shards update`. Works after that

spec/lucky/action_spec.cr
```diff
@@ -82,6 +82,17 @@ class PlainText::Index < Lucky::Action
   end
 end
 
+class OptionalParams::Index < Lucky::Action
+  param page : Int32?
+  param with_default : String? = "default"
+  param with_int_default : Int32? = 1
+  param with_int_never_nil : Int32 = 1337
```
paulcsmith: Could you add a case for `param required_page : Int32`?

spec/lucky/action_spec.cr
paulcsmith: I think that if a parameter can’t be parsed then that is an error that should be raised. I think it would be a good idea to make a custom error that we can raise so it can be handled if you want. 

Something like `Lucky::InvalidParamError`. It should probably accept (in the initializer) the param name, and the value that it tried to parse, and the type it tried to parse it to

spec/lucky/action_spec.cr
hanneskaeufler: ✅

spec/lucky/action_spec.cr
hanneskaeufler: Yep agreed I extracted two custom Exceptions. Feel free to comment on naming/positioning ✅

spec/lucky/action_spec.cr
```diff
@@ -145,6 +164,64 @@ describe Lucky::Action do
...
+
+    it "can get manually defined required params" do
+      action = RequiredParams::Index.new(build_context(path: "/?required_page=1"), params)
+      action.required_page.should eq 1
+    end
+
+    it "raises for missing required params" do
+      action = RequiredParams::Index.new(build_context(path: ""), params)
+      expect_raises(Lucky::Exceptions::MissingParam) { action.required_page }
+    end
```
paulcsmith: I think it would be good to also add a test for `RequiredParams::Index.with(required_page: 1)` to make sure the argument is added correctly

spec/lucky/action_spec.cr
```diff
@@ -145,6 +164,64 @@ describe Lucky::Action do
...
+
+    it "overrides the default if present" do
+      action = OptionalParams::Index.new(build_context(path: "/?with_int_never_nil=42"), params)
+      action.with_int_never_nil.should eq 42
+    end
+
+    it "is added as optional argument to the path" do
+      OptionalParams::Index.path(7).should eq "/optional_params?page=7"
+      OptionalParams::Index.path(7, "/other").should eq "/optional_params?page=7&with_default=%2Fother"
```
paulcsmith: Could you use the named arg version so it's clear what is going on `path(page: 7, with_default: "/other"`?

spec/lucky/action_spec.cr
```diff
@@ -145,6 +164,64 @@ describe Lucky::Action do
...
+    end
+
+    it "is added as optional argument to the route" do
+      OptionalParams::Index.route(7).should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7")
+      OptionalParams::Index.route(7, "/other").should eq Lucky::RouteHelper.new(:get, "/optional_params?page=7&with_default=%2Fother")
```
paulcsmith: Same with this one

spec/lucky/action_spec.cr
```diff
@@ -145,6 +164,64 @@ describe Lucky::Action do
...
+    end
+
+    it "doesnt raise when we cannot parse the optional param into the desired type" do
```
paulcsmith: I think it should raise if it can't be parsed, just like with the required params. This would keep it consistent and would make it clear when something goes wrong. What do you think?

src/lucky/exceptions.cr
```diff
@@ -12,5 +12,23 @@ module Lucky
 
     class InvalidMessage < Base
     end
+
+    class InvalidParam < Base
+      def initialize(@param_name : String, @param_value : String, @param_type : String)
+      end
+
+      def message : String?
+        "Non-optional param \"#{@param_name}\" with value \"#{@param_value}\" couldn't be parsed to a \"#{@param_type}\""
```
paulcsmith: How about "Non-optional" -> "Required"?

src/lucky/params.cr
```diff
@@ -11,7 +11,7 @@ class Lucky::Params
   end
 
   def get!(key) : String
-    get(key) || raise "Missing parameter: #{key}"
+    get(key) || raise Lucky::Exceptions::MissingParam.new(key.to_s)
```
paulcsmith: Much better! I really like this 👍

src/lucky/exceptions.cr
```diff
@@ -12,5 +12,23 @@ module Lucky
...
+      end
+    end
+
+    class MissingParam < Base
```
paulcsmith: Could you add a getter for `param_name`?

src/lucky/exceptions.cr
```diff
@@ -12,5 +12,23 @@ module Lucky
 
     class InvalidMessage < Base
     end
+
+    class InvalidParam < Base
+      def initialize(@param_name : String, @param_value : String, @param_type : String)
```
paulcsmith: Could you add getters for all these? That way people can use them in custom error message if they want to

src/lucky/routeable.cr
```diff
@@ -100,6 +118,18 @@ module Lucky::Routeable
 
       is_root_path = path == ""
       path = "/" if is_root_path
+
+      query_params = {} of String => String
+      {% for param in PARAM_DECLARATIONS %}
+        unless {{ param.var }} == {{ param.value || nil }}
```
paulcsmith: What do you think about extracting an intermediate variable to make it clearer what is going on with this? It took me a bit to get it. Maybe something like this:

```crystal
param_is_default_or_nil = {{ param.var }} == {{ param.value || nil }}
unless param_is_default_or_nil
```

src/lucky/routeable.cr
```diff
@@ -100,6 +118,18 @@ module Lucky::Routeable
...
+          query_params["{{ param.var }}"] = {{ param.var }}.to_s
+        end
+      {% end %}
+
+      unless query_params.empty?
+        path += "?#{HTTP::Params.encode(query_params)}"
+      end
```
paulcsmith: Just FYI, there is a lot of duplication, but I think that's because of how these methods are written. I'm going to take a look at these later to see how they can be cleaned up. I think I might remove `path` and instead just have `with/route`. Then if you want the path you could do `User::Show.with(user.id).path` I'm not sure though

src/lucky/routeable.cr
```diff
@@ -115,4 +145,49 @@ module Lucky::Routeable
...
+          {{ type_declaration.value || nil }}
+        else
+          result = {{ type_declaration.type.types.first }}::Lucky.parse(val)
+          if result.is_a? {{ type_declaration.type.types.first }}::Lucky::SuccessfulCast
+            result.value
+          else
+            nil
+          end
+        end
+      {% end %}
```
paulcsmith: Instead of having an if/else you could use some macro variables to keep a lot of the code the same:

```crystal
{% optional = type_declaration.type.is_a?(Union) %}
# Uses get! if required. This will raise MissingParam if missing
val = params.get{% unless optional %}!{% end %}(:{{ type_declaration.var.id }})

result = {{ type_declaration.type }}::Lucky.parse(val)

# Use if/else instead of early return to make things clear
if result.is_a? {{ type_declaration.type }}::Lucky::SuccessfulCast
  result.value
else
  # Use named args to make it clear what each arg is
  raise Lucky::Exceptions::InvalidParam.new(
    param_name: "{{ type_declaration.var.id }}",
    param_value: val.to_s,
    param_type: "{{ type_declaration.type }}"
  )
end
```

This also incorporates a few other changes I suggested earlier, like always raising if a cast fails

spec/lucky/action_spec.cr
hanneskaeufler: ✅

hanneskaeufler: Also, `with` doesnt take any named arguments,
```crystal
    def self.with(**args)
      route(**args)
    end

    def self.with(*args)
      route(*args)
    end
```

src/lucky/routeable.cr
hanneskaeufler: I took a lot of inspiration from your comment, but it's not quite that simple because we have to do a default check for required params as well. So we either do `get!` and catch the exception, or do a nil check. I opted for the second one for now, not that I particularly like it^^

hanneskaeufler:
Thank god for those tests, that macro refactoring is mind-bending haha. I think I tackled the issues, fire away if you have more comments @paulcsmith . Thanks!

paulcsmith:
Excellent work! Thank so much @hanneskaeufler

hanneskaeufler:
Thanks for the learning experience! 🎉 🚢 🚀

Issue #288, opened 2017-12-09
options_for_select sets selected on every option


cpytel:
`options_for_select` outputs the following HTML:

```
<option value="" selected="false"></option>
<option value="2" selected="false">New York City</option>
<option value="1" selected="true">Boston</option>
<option value="3" selected="false">San Francisco</option>
```

Unfortunately, having `selected="true"` or `selected="false"` is not how this works in HTML. The mere presence of the selected attribute causes browsers to recognize that option as selected. We should not output the selected option at all for unselected options. The markup should be like this:

```
<option value="" ></option>
<option value="2">New York City</option>
<option value="1" selected="true">Boston</option>
<option value="3">San Francisco</option>
```

Issue #286, opened 2017-12-08
options_for_select only ever returns one option


cpytel:
While building an app, we discovered that there was only ever one option in our selects! Upon investigation, we identified that there was an early `return` in `options_for_select`:

https://github.com/luckyframework/lucky/blob/9e8ffcb0dba626996daa1bd59d428e2de415cbf2/src/lucky/tags/select_helpers.cr#L13

PR #289, opened 2017-12-09
Hk fix html select

hanneskaeufler:
As shown in #288 and #286, the select/options is not quite working right.

- [x] Fix weird bug. Why is Volvo showing up twice???
- [x] We are only `returning` the html in order to assert in the tests because we have no way to test what `@view` contains. Is that nice?

### DIFF
--- a/spec/lucky/select_helpers_spec.cr
+++ b/spec/lucky/select_helpers_spec.cr
@@ -26,9 +26,10 @@ describe Lucky::SelectHelpers do
     HTML
   end
 
-  it "renders option" do
-    (view.options_for_select form.company_id, [{"Volvo", 2}]).to_s.should contain <<-HTML
+  it "renders options" do
+    (view.options_for_select form.company_id, [{"Volvo", 2}, {"BMW", 3}]).to_s.should contain <<-HTML
     <option value="2">Volvo</option>
+    <option value="3">BMW</option>
     HTML
   end
 
--- a/spec/lucky/select_helpers_spec.cr
+++ b/spec/lucky/select_helpers_spec.cr
@@ -5,6 +5,19 @@ private class TestPage
 
   def render
   end
+
+  def select(field)
+    select_input field do
+    end
+  end
+
+  def options(field, options)
+    options_for_select(field, options)
+  end
+
+  def html
+    @view
+  end
 end
 
 class SomeFormWithCompany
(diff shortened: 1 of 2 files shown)


spec/lucky/select_helpers_spec.cr
```diff
@@ -5,6 +5,21 @@ private class TestPage
 
   def render
   end
+
+  def select(field)
```
paulcsmith: Maybe `render_select`?

spec/lucky/select_helpers_spec.cr
```diff
@@ -5,6 +5,21 @@ private class TestPage
...
+    select_input field do
+    end
+    self
+  end
+
+  def options(field, options)
```
paulcsmith: And `render_options`? That way it's clear that it isn't using the `options` and `select` tags

spec/lucky/select_helpers_spec.cr
```diff
@@ -21,20 +36,22 @@ end
...
+    <option value="2">Volvo</option><option value="3">BMW</option>
     HTML
   end
 
   it "renders selected option" do
-    (view.options_for_select form.company_id, [{"Volvo", 1}]).to_s.should contain <<-HTML
-    <option value="1" selected="true">Volvo</option>
+    view.options(form.company_id, [{"Volvo", 1}, {"BMW", 3}])
+      .html.to_s.should contain <<-HTML
+    <option value="1" selected="true">Volvo</option><option value="3">BMW</option>
```
paulcsmith: I think this should also have another assertion that checks that `BMW` does not contain a `selected` attribute

hanneskaeufler: That should be covered by asserting the `<option value="3">BMW</option>` shouldn't it? Nevertheless I changed the assertions from `containt` to `eq` as I dont think we want anything else in there anyway.

hanneskaeufler:
Thanks paul, I pushed one more commit concerning your notes.

spec/lucky/select_helpers_spec.cr
paulcsmith: Oh you're right. I didn't see that last option. This looks great. And I like the change to `eq` 👍

paulcsmith:
Closes #288 
Closes #286

Issue #129, opened 2017-09-27
Make it more clear what you need to do if you forget to render/redirect at the end of an action


paulcsmith:
https://gist.github.com/paulcsmith/f3d066fc696f516fbbac932c6bbd97a5

5minpause:
I played around with this but couldn't get very far. I thought it could fit in `routable.cr`

```
  macro setup_call_method(body)
    def call : LuckyWeb::Response
      callback_result = run_before_callbacks

      response = if callback_result.is_a?(LuckyWeb::Response)
        callback_result
      else
        {{ body }}
      end

      callback_result = run_after_callbacks

      if callback_result.is_a?(LuckyWeb::Response)
        callback_result
      else
        if response.is_a?(LuckyWeb::Response)
          response
        else
          {% raise "Must return a response. You probably forgot to use `render`" %}
        end
      end
    end
  end

```

But then I got an error for my `Home::Index`

```
class Home::Index < BrowserAction
  get "/" do
    redirect to: Pages::Index
  end
end
```

```
± lucky dev                                                                                                                                             2.3.3
[WARN] No ENV file found
18:13:49 assets.1   |  yarn run v1.3.2
18:13:49 web.1      |  Error in tasks.cr:2: while requiring "./src/app"
18:13:49 web.1      |  require "./src/app"
18:13:49 web.1      |  ^
18:13:49 web.1      |  in src/app.cr:10: while requiring "./actions/**"
18:13:49 web.1      |  require "./actions/**"
18:13:49 web.1      |  ^
18:13:49 web.1      |  in src/actions/home/index.cr:2: Must return a response. You probably forgot to use `render`
18:13:49 web.1      |    get "/" do
18:13:49 web.1      |    ^
18:13:49 web.1      Exited Successfully
18:13:49 assets.1   |  $ /Users/holger/projects/ifeellucky/lucky-blog/node_modules/.bin/brunch watch
18:13:51 assets.1   |  18:13:51 - info: compiled 6 files into 2 files in 1.5 sec
^C[WARN] Interrupted by User
[DONE] Killing all processes with signal  SIGINT
```

Would do I need to do better here?

paulcsmith:
Yeah that actually wasn't the right way to do it. I'll remove the link to that gist. Here's something that should work better: https://gist.github.com/paulcsmith/f3d066fc696f516fbbac932c6bbd97a5

5minpause:
Ah that makes sense. I'll be on it.

PR #291, opened 2017-12-10
Raise friendly error with invalid action

ianagne:
An action must return a response object to be valid. If a developer
fails to do so, the current error message isn't very clear on what needs
to be fixed.  Change the error message to explicitly state that one should
return a response from the action, and suggest methods to do so.

See issue #129

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -37,8 +37,14 @@ module Lucky::Renderable
     response.print
   end
 
-  private def handle_response(_response)
-    {% raise "You must return a Lucky::Response from #{@type}. You can do that by using methods such as `render`, `redirect`, `json`" %}
+  private def handle_response(_response : T) forall T
+    {% raise <<-ERROR
+
+     #{@type} returned #{T}, but it must return a Lucky::Response.
+
+     Try this...
+     ▸ Make sure to use a method like `render`, `redirect`, `json` at the end of your action.
+    ERROR %}
   end
 
   private def render_text(body)
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -43,7 +43,8 @@ module Lucky::Renderable
      #{@type} returned #{T}, but it must return a Lucky::Response.
 
      Try this...
-     ▸ Make sure to use a method like `render`, `redirect`, `json` at the end of your action.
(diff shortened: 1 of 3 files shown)


src/lucky/renderable.cr
```diff
@@ -37,6 +37,10 @@ module Lucky::Renderable
     response.print
   end
 
+  private def handle_response(_response)
```
paulcsmith: I just learned this nice trick so that we can get the type that someone used. This will make it easier for people to debug:

```crystal
private def handle_response(_response : T) forall T
  {% raise <<-ERROR
  You must return a Lucky::Response from #{@type}. Instead, you returned #{T}.

  Try this...
    ▸ Make sure you use `render`, `redirect`, `json` at the end of your action.
  ERROR %}
end
```

I also modified the error a bit to match some of the newer error messages in the other Lucky libs. What do you think?

paulcsmith: Actually, you're right that it might be lost. How about we switch it to something like this:

`Your action returned #{T}, but it must return a Lucky::Response.`

I think it's a bit shorter, and this way the type they returned won't be buried. What do you think?

ianagne: Here's a screenshot of the message in action. I think we should keep `@type` in the message to provide context.  I'm not sure if `T` is helpful or confusing, as it will include Lucky::Response in output along with any other types that were incorrectly used.

screen shot 2017-12-11 at 10 57 06 pm

paulcsmith: I think that looks good. I think there are just a couple more things:

* Add the word `or` before `json`
* Indent arrow 2 spaces
* Consider adding another tip: "If you are using a conditional, make sure all branches return a Lucky::Response". Maybe that's obvious though. I'm not sure. What do you think?

ianagne:
Here's the latest screenshot of the error:
screen shot 2017-12-12 at 5 18 52 pm

src/lucky/renderable.cr
ianagne: It's probably obvious to more experienced developers, but the extra nudge about conditionals totally would have been helpful to me when I was first learning Rails.

ianagne: I can definitely see a case where the extra heads up about conditionals would unstick someone.  I added it in the commit.

PR #303, opened 2017-12-15
Allow empty text for `link`

### LINKED ISSUE #104: Allow empty text for `link`
Sometimes the content is purely style with CSS so you don't need anything inside the link tag

ianagne:
Some links do not need text (e.g. links using a background image ).  Add
link helpers that accept empty text without a block.

Closes #104

### DIFF
--- a/spec/lucky/link_helpers_spec.cr
+++ b/spec/lucky/link_helpers_spec.cr
@@ -45,6 +45,14 @@ private class TestPage
       text "Hello"
     end
   end
+
+  def get_route_without_text
+    link to: LinkHelpers::Index
+  end
+
+  def string_path_without_text
+    link to: "/foo"
+  end
 end
 
 describe Lucky::LinkHelpers do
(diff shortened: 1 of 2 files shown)


spec/lucky/link_helpers_spec.cr
```diff
@@ -45,6 +45,14 @@ private class TestPage
       text "Hello"
     end
   end
+
+  def get_route_without_text
+    link to: LinkHelpers::Index
+  end
+
+  def string_path_without_text
+    link to: "/foo"
```
paulcsmith: I think instead of allowing an empty `link` it should accept a block so you can render images, text, etc. inside of the link

ianagne:
I think I misunderstood.  I thought the goal was to create a link helper that doesn't require text or a block.  I think what you're describing is already possible.

screen shot 2017-12-15 at 6 37 25 pm

ianagne:
I think what I was trying to do is also already possible.

paulcsmith:
Oh you are right. I misremembered the issue. You’re right that sometimes you want a link with nothing inside it because it can be styled with css using abackgournd image or something

Thanks for doing this. It looks good

spec/lucky/link_helpers_spec.cr
paulcsmith: Never mind :)

PR #304, opened 2017-12-15
More consistent flash handling

### LINKED ISSUE #277: Flash notice doesn't clear (Turbolinks?)
My understanding is, that flash messages should disappear on page change (I might be wrong?). So when I have an action like this:
```crystal
  post "/posts/create" do
    PostForm.create(params) do |form, post|
      if post
        flash.notice = "Post successfully published."
        redirect Blog::Index
      else
        render Blog::Posts::NewPage, post_form: form
      end
    end
  end
```

I would expect to see this message only for one page.
However I am seeing it for two page reloads. Page reloads being relative of course due to turbolinks.
Am I misunderstanding or is this something we need to improve? Sorry for the lengthy gif.

dez -05-2017 22-08-55

paulcsmith:
Fixes #277

The previous implementation tracked which keys were read/not read. This
ended up being problematic because sometimes keys weren't marked as read
and so they would stay in the flash forever.

Instead, the new implementation only passes the flashes to the next
pages and then removes the rest.

You can use `now` to only set a message for the current request.

All the other hash methods are available because they are forwarded to
`all_messages`.

The `[]` and `[]=` are overridden to work only with symbols and strings.


src/lucky/flash.cr
```diff
@@ -16,78 +16,113 @@ module Lucky
...
+
+      def [](_not_accepted : T) forall T
+        {% raise "Flash key must be a Symbol or String, got #{T}" %}
+      end
+
+      def [](key : Symbol | String)
+        all_messages[key.to_s]
+      end
+
+      def []?(_not_accepted : T) forall T
```
paulcsmith: These make it so anything other than a Symbol or String raises a helpful message

src/lucky/page_helpers/text_helpers.cr
```diff
@@ -221,11 +221,11 @@ module Lucky::TextHelpers
...
-      drop_index = [part.size - radius, 0].max
-      part[drop_index..-1]
-    else
-      part.first(radius)
-    end
+             drop_index = [part.size - radius, 0].max
+             part[drop_index..-1]
+           else
+             part.first(radius)
+           end
```
paulcsmith: This was caught by `crystal tool format`

src/lucky/flash.cr
```diff
@@ -16,78 +16,113 @@ module Lucky
...
-    class Store < Hash(String, String)
-      def self.from_session(json)
-        flash = new
-        values = JSON.parse(json)
-        values.each { |k, v| flash[k.to_s] = v.to_s }
-        flash
+    class Store
+      @next_key_is_for_now : Bool = false
+
+      forward_missing_to all_messages
```
paulcsmith: You can call all the regular Hash methods like `reject_if`, etc.

src/lucky/flash.cr
```diff
@@ -16,78 +16,113 @@ module Lucky
...
-      def keep(key = nil)
-        @read.delete key
-        @now.delete key
+      def danger=(message : String)
+        self["danger"] = message
       end
 
-      def alert
-        self["alert"]
+      def info
```
mxie: Looks like you've changed the API here (renamed some flash methods and added a new one). Can you document this too in your commit message and release?

src/lucky/tags/select_helpers.cr
```diff
@@ -9,7 +9,7 @@ module Lucky::SelectHelpers
 
   def options_for_select(field : LuckyRecord::AllowedField, select_options : Array(Tuple(String, SelectOption)), **html_options)
     select_options.each do |option_name, option_value|
-      attributes = { "value" => option_value.to_s }
+      attributes = {"value" => option_value.to_s}
```
mxie: Is it convention to not use padded hashes?

tasks/gen/templates/model_template.cr
```diff
@@ -8,7 +8,7 @@ class Lucky::ModelTemplate < Teeplate::FileTree
   directory "#{__DIR__}/model/"
 
   def initialize(@name : String)
-    @pluralized_name  = LuckyInflector::Inflector.pluralize(@name)
+    @pluralized_name = LuckyInflector::Inflector.pluralize(@name)
```
mxie: I think we can do without this change.

src/lucky/tags/select_helpers.cr
paulcsmith: Yep! All the style changes are done by `crystal tool format` automatically. So yeah it is the Crystal convention

tasks/gen/templates/model_template.cr
paulcsmith: This was done automatically by `crystal tool format`

src/lucky/flash.cr
paulcsmith: Yep 'll add a commit and some new docs for them :D Thanks!

paulcsmith:
Thanks for the review @mxie

PR #311, opened 2017-12-19
Output render log message only once

### LINKED ISSUE #270: Log what page was rendered only if it was returned at the end of the action
Right now if you do this:

```crystal
render TestPage
render AnotherPage
```

There will be a log message saying TestPage was rendered and another saying AnotherPage was rendered. Only the one returned should get a log message

ianagne:
If multiple pages were rendered in an action, there would be a log
message for each render.  Attach the log message to the Lucky::Response
and move the logging mechanism to #handle_response, to remove the
possibility of redundant messages.

Closes #270

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -53,9 +53,8 @@ describe Lucky::Action do
   describe "rendering HTML pages" do
     it "render assigns" do
       response = Rendering::Index.new(build_context, params).call
-      body = response.body
 
-      body.should contain "Anything"
+      response.body.should contain "Anything"
       response.debug_message.to_s.should contain "Rendering::Index"
     end
   end
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -55,7 +55,7 @@ describe Lucky::Action do
       response = Rendering::Index.new(build_context, params).call
 
       response.body.should contain "Anything"
-      response.debug_message.to_s.should contain "Rendering::Index"
+      response.debug_message.to_s.should contain "Rendering::IndexPage"
     end
   end
 
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -52,9 +52,11 @@ end
 describe Lucky::Action do
   describe "rendering HTML pages" do
     it "render assigns" do
-      body = Rendering::Index.new(build_context, params).call.body
+      response = Rendering::Index.new(build_context, params).call
+      body = response.body
 
       body.should contain "Anything"
+      response.debug_message.to_s.should contain "Rendering::Index"
     end
   end
 
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -55,7 +55,7 @@ module Lucky::Renderable
     %}
   end
 
-  private def log_response(response)
+  private def log_response(response : Lucky::Response)
     debug_message = response.debug_message
     if !debug_message.nil?
       context.add_debug_message(debug_message)
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -19,13 +19,14 @@ module Lucky::Renderable
         {{ key }}: {{ key }},
       {% end %}
     )
-    log_html_render(context, view)
+    debug_message = log_message(view)
     body = view.perform_render.to_s
-    Lucky::Response.new(context, "text/html", body)
+    Lucky::Response.new(context, "text/html", body,
+                        debug_message: debug_message)
   end
 
-  private def log_html_render(context, view)
-    context.add_debug_message("Rendered #{view.class.colorize(HTTP::Server::Context::DEBUG_COLOR)}")
+  private def log_message(view)
+    "Rendered #{view.class.colorize(HTTP::Server::Context::DEBUG_COLOR)}"
   end
 
   def perform_action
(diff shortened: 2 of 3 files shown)


src/lucky/response.cr
```diff
@@ -1,9 +1,9 @@
 class Lucky::Response
   DEFAULT_STATUS = 200
 
-  getter context, content_type, body
+  getter context, content_type, body, debug_message
 
-  def initialize(@context : HTTP::Server::Context, @content_type : String, @body : String, @status : Int32? = nil)
+  def initialize(@context : HTTP::Server::Context, @content_type : String, @body : String, @status : Int32? = nil, @debug_message : String? = nil)
```
ianagne: These arguments are getting a little log, but I'm not sure how best to handle this in Crystal.  In Ruby, I'd probably move the optional args into an options Hash.  Is there something similar to be done here?

spec/lucky/action_rendering_spec.cr
```diff
@@ -52,9 +52,11 @@ end
 describe Lucky::Action do
   describe "rendering HTML pages" do
     it "render assigns" do
-      body = Rendering::Index.new(build_context, params).call.body
+      response = Rendering::Index.new(build_context, params).call
+      body = response.body
```
paulcsmith: Instead of assigning to a variable, you could do `response.body.should...` on the line below

spec/lucky/action_rendering_spec.cr
```diff
@@ -52,9 +52,11 @@ end
...
 
       body.should contain "Anything"
+      response.debug_message.to_s.should contain "Rendering::Index"
```
paulcsmith: I think this should test `"Rendering::IndexPage"` right? I may be misreading this though

src/lucky/renderable.cr
```diff
@@ -19,13 +19,14 @@ module Lucky::Renderable
         {{ key }}: {{ key }},
       {% end %}
     )
-    log_html_render(context, view)
+    debug_message = log_message(view)
     body = view.perform_render.to_s
-    Lucky::Response.new(context, "text/html", body)
+    Lucky::Response.new(context, "text/html", body,
+                        debug_message: debug_message)
```
paulcsmith: I think you could remove the intermediate variable and call `log_message(view)` here directly.

Also can you run `crystal tool format` to make sure this is formatted? You may need to put "text/html" on a newline for the formatter to put them all on newlines. I'm not sure.

src/lucky/response.cr
paulcsmith: I would put @context on a newline and then run `crystal tool format` and it should put each of them on a newline so it's a bit easier to read and isn't so wide. LMK if that does the trick :)

src/lucky/renderable.cr
```diff
@@ -34,18 +35,28 @@ module Lucky::Renderable
...
+      Try this...
+        ▸ Make sure to use a method like `render`, `redirect`, or `json` at the end of your action.
+        ▸ If you are using a conditional, make sure all branches return a Lucky::Response.
+      ERROR
+    %}
+  end
+
+  private def log_response(response)
+    debug_message = response.debug_message
+    if !debug_message.nil?
```
paulcsmith: You can do `if debug_message` to keep this a bit more concise, or `try` might make it even cleaner:

```crystal
response.debug_message.try do |message|
  context.add_debug_message(message)
end
```

src/lucky/renderable.cr
```diff
@@ -34,18 +35,28 @@ module Lucky::Renderable
...
-       ▸ If you are using a conditional, make sure all branches return a Lucky::Response.
-    ERROR %}
+      Try this...
+        ▸ Make sure to use a method like `render`, `redirect`, or `json` at the end of your action.
+        ▸ If you are using a conditional, make sure all branches return a Lucky::Response.
+      ERROR
+    %}
+  end
+
+  private def log_response(response)
```
paulcsmith: I think adding a type restriction here might make it more clear what the response is: `response : Lucky::Response`. So people know it's not an `HTTP::Server::Response`

spec/lucky/action_rendering_spec.cr
ianagne: Definitely right.

src/lucky/renderable.cr
ianagne: Good call.

spec/lucky/action_rendering_spec.cr
ianagne: 👍

src/lucky/renderable.cr
ianagne: I'm always down to remove temp variables.

src/lucky/response.cr
ianagne: The formatter only seems to enforce proper indentation in this case.
screen shot 2017-12-19 at 7 08 05 pm

The formatter also insists that the closing paren be on the same line as the last argument.  That would normally mean that the opening paren should be on the same line as the first argument, but the formatter doesn't enforce that.  

It does still let me manually make the change though:
screen shot 2017-12-19 at 7 14 02 pm

paulcsmith: I just ran into this myself. Strange that it doesn't automatically put them on newlines. I like what you have in that second screenshot though 👍

src/lucky/renderable.cr
ianagne: Of note, the formatter doesn't seem to have an opinion on how to indent the arguments here.

Issue #328, opened 2017-12-28
CSRF Protection fails


hanneskaeufler:
After upgrading lucky my forms don't create any records anymore because they are `403`ing due to csrf protection.

I did add the csrf tags to the `<head>` and the form also generates a hidden input with the same token:
<img width="663" alt="bildschirmfoto 2017-12-28 um 15 21 58" src="https://user-images.githubusercontent.com/68024/34413343-161b03e0-ebe3-11e7-9ec8-ccafe6ef4fb1.png">
It is posted correctly it seems:
<img width="367" alt="bildschirmfoto 2017-12-28 um 15 22 22" src="https://user-images.githubusercontent.com/68024/34413349-24b6b00c-ebe3-11e7-8277-471be7305f09.png">

However, when `pp`ing the session provided token shows a different token:
<img width="492" alt="bildschirmfoto 2017-12-28 um 15 24 33" src="https://user-images.githubusercontent.com/68024/34413366-46491bc4-ebe3-11e7-8309-ea2604c08e69.png">
So it understandably doesn't match up with the user provided token, which is as shown in the html above
<img width="491" alt="bildschirmfoto 2017-12-28 um 15 25 17" src="https://user-images.githubusercontent.com/68024/34413379-66776aae-ebe3-11e7-8889-5e6e0330bcde.png">

PR #329, opened 2017-12-28
Only generate one session CSRF token

paulcsmith:
The problem with generating a new CSRF token after every request is that
you can open two tabs and one of the tabs will have the old CSRF token
and fail when submitting the form.

After doing some research it seems that it is fine to have just one CSRF
token:

* https://github.com/elixir-plug/plug/issues/504
* http://security.stackexchange.com/questions/22903/why-refresh-csrf-token-per-form-request
* http://stackoverflow.com/questions/8655817/csrf-protection-do-we-have-to-generate-a-token-for-every-form

An additional improvement for the future would be to mask the CSRF
token: https://medium.com/rubyinside/a-deep-dive-into-csrf-protection-in-rails-19fa0a42c0ef

Also closes: #328

### DIFF
--- a/spec/lucky/protect_from_forgery_spec.cr
+++ b/spec/lucky/protect_from_forgery_spec.cr
@@ -9,14 +9,12 @@ class ProtectedAction::Index < Lucky::Action
 end
 
 describe Lucky::ProtectFromForgery do
-  it "sets a new CSRF token after a successful request" do
+  it "sets a CSRF token if none was set" do
     original_token = "my_token"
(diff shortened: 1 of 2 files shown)


paulcsmith:
@hanneskaeufler Would you mind trying this out and seeing if it fixes your issue? I'm pretty sure it will, but I'd love to double check just in case :)

hanneskaeufler:
It does fix the issue 👍

paulcsmith:
Thanks for confirming @hanneskaeufler!

PR #335, opened 2017-12-30
Add `brew update` to update listings before upgrade

westoque:


edwardloveall:
I believe homebrew will automatically update itself before running `brew install`, `brew upgrade` or `brew tap`. If for some reason it isn't, then yeah I totally agree this is a good idea.

paulcsmith:
I think it is auto-updated as @edwardloveall mentioned, but I'm fine adding it as it is usually a quick step and I don't see much harm in doing it just in case. We can do some more research later to see if it is 100% necessary.

paulcsmith:
@westoque did you need to run `brew update` or did it auto update?

P.S. Thanks for the PR

westoque:
@paulcsmith yes. i needed to run `brew update` before doing the upgrade since running just the upgrade did not detect latest version.

and thanks for the awesome framework ❤️

PR #336, opened 2018-01-03
Color 5xx in red and 4xx in yellow, closes #330

### LINKED ISSUE #330: Log status code as red if in the 400-599 range
It should be more obvious that something went wrong. Highlighting the status code as red would help

https://github.com/luckyframework/lucky/blob/1bde3a7a324a0ec60d4c4a123b5a16b4dcf190b3/src/lucky/log_handler.cr#L19-L21

hanneskaeufler:
I totally agree, this has confused me once or twice. I opted to make 4xx codes yellow, 5xx red.

bildschirmfoto 2018-01-03 um 18 09 38
bildschirmfoto 2018-01-03 um 18 10 22

### DIFF
--- a/src/lucky/log_handler.cr
+++ b/src/lucky/log_handler.cr
@@ -17,7 +17,7 @@ class Lucky::LogHandler
     elapsed_text = elapsed_text(elapsed)
 
     if !context.hide_from_logs?
-      @io.puts "#{context.request.method} #{context.response.status_code} #{context.request.resource.colorize(:green)}#{timestamp(time)} (#{elapsed_text})"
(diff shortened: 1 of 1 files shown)


hanneskaeufler:
Mhm I dig it like this (4xx still yellow):
bildschirmfoto 2018-01-04 um 18 42 24

src/lucky/log_handler.cr
```diff
@@ -14,18 +14,38 @@ class Lucky::LogHandler
     time = Time.now
     call_next(context)
     elapsed = Time.now - time
-    elapsed_text = elapsed_text(elapsed)
 
     if !context.hide_from_logs?
-      @io.puts "#{context.request.method} #{context.response.status_code} #{context.request.resource.colorize(:green)}#{timestamp(time)} (#{elapsed_text})"
+      log_request(context, time, elapsed)
```
paulcsmith: I like this method extraction 👍

PR #341, opened 2018-01-04
Add pre, code and address tags

hanneskaeufler:
Tags are now sorted alphabetically, I added `pre`, `code`, `address`. We still don't include everything from https://www.w3schools.com/tags/ btw.


paulcsmith:
Nice! Thanks, Hannes

Issue #347, opened 2018-01-06
Add to_sentence helper


paulcsmith:

tommyschaefer:
Hi there! I was wondering if it might be okay for me to give this a shot? No problem at all if you or anyone else would prefer to take it on yourselves 😄

paulcsmith:
@tommyschaefer That would be great! It would go here: https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr

For the first pass it doesn't need any locale/i18n options. It should work pretty much identically to the Rails version of to_sentence, but it will use the `text` method to output the result to the page.

LMK if you have any questions. I'm happy to help :)

Issue #275, opened 2017-12-04
Add inline documentation


paulcsmith:
Inline documentation was purposely left off when things were changing, but now the API is starting to stabilize. Adding documentation to the public facing methods would be fantastic.

If you'd like to help, here is how Crystal docs work: https://crystal-lang.org/docs/conventions/documenting_code.html

Feel free to make small PRs so it's easy to review and merge them in :D

Thanks so much for the help!

paulcsmith:
I'll go ahead and close this. We of course can always use more, but we'll also never have everything covered so would rather not have this open forever :P

PR #348, opened 2018-01-07
Add to_sentence helper

tommyschaefer:
Hi there!

This is my first pass at tackling #347 and implementing the `#to_sentence` text helper!

I added tests, the implementation, and some documentation (based on #275). I'm running into an error when I try and build the documentation to verify that I didn't mess up the markdown, though.

I ran `crystal docs` and got:

```
Unhandled type in `name`: Lucky::Action+:Class
Crystal::Doc::Type#name:String
Crystal::Doc::Type#full_name_without_type_vars<String::Builder>:String::Builder
Crystal::Doc::Type#full_name<String::Builder>:(String::Builder | Nil)
Crystal::Doc::Method#args_to_html<String::Builder, Bool>:(String::Builder | Nil)
Crystal::Doc::Method#id:String
Crystal::Doc::MethodSummaryTemplate#to_s<File>:File
Crystal::Doc::Generator#generate_types_docs<Array(Crystal::Doc::Type), String, Array(Crystal::Doc::Type)>:Nil
Crystal::Doc::Generator#generate_types_docs<Array(Crystal::Doc::Type), String, Array(Crystal::Doc::Type)>:Nil
Crystal::Command#run:(Bool | Crystal::Compiler::Result | IO::FileDescriptor | Nil)
main

Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
```

I'm not sure whether I should actually open an issue for this, or if I did something wrong 😄

Once everything is looking good to you all, I can squash the commits!

Thank you so much for your time and feedback!! 😄

### DIFF
--- a/src/lucky/page_helpers/text_helpers.cr
+++ b/src/lucky/page_helpers/text_helpers.cr
@@ -126,6 +126,54 @@ module Lucky::TextHelpers
     end
   end
 
+  # Creates a comma-separated sentence from the provided `Enumerable` *list*
+  # and appends it to the view.
+  #
+  # #### Options:
+  #
+  # The following options allow you to specify how the sentence is constructed:
+  #   - *word_connector* - A string used to join the elements in *list*s
+  # containing three or more elements (Default is ", ")
+  #   - *two_word_connector* - A string used to join the elements in *list*s
+  # containing exactly two elements (Default is " and ")
+  #   - *last_word_connector* - A string used to join the last element in
+  # *list*s containing three or more elements (Default is ", and ")
+  #
+  # #### Examples:
+  #
+  #     to_sentence([] of String)            # => ""
+  #     to_sentence([1])                     # => "1"
+  #     to_sentence(["one", "two"])          # => "one and two"
+  #     to_sentence({"one", "two", "three"}) # => "one, two, and three"
+  #
+  #     to_sentence(["one", "two", "three"], word_connector: " + ")
+  #     # => one + two, and three
+  #
+  #     to_sentence(Set{"a", "z"}, two_word_connector: " to ")
+  #     # => a to z
+  #
+  #     to_sentence(1..3, last_word_connector: ", or ")
+  #     # => 1, 2, or 3
+  #
+  # NOTE: By default `#to_sentence` will include a
+  # [serial comma](https://en.wikipedia.org/wiki/Serial_comma). This can be
+  # overriden like so:
+  #
+  #     to_sentence(["one", "two", "three"], last_word_connector: " and ")
+  #     # => one, two and thre
+  def to_sentence(list : Enumerable,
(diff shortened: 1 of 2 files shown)


hanneskaeufler:
Jepp `crystal docs` is broken (as in doesn't run on lucky right now), you're not doing anything wrong. To see your changes you can do something like `crystal docs path/to/something.cr` to just see docs for your one file.

tommyschaefer:
@hanneskaeufler Awesome! Thanks for the help with docs and nice spot on the typo! I appreciate the help!!

I've updated the method documentation to be a bit more readable after seeing what it looked like when rendered and squashed my commits. Let me know if there are any other improvements you all would like to see, and I'll be happy to make them! 😄

Thank you again for the feedback and help @hanneskaeufler!

src/lucky/page_helpers/text_helpers.cr
```diff
@@ -126,6 +126,54 @@ module Lucky::TextHelpers
...
+  #
+  #     to_sentence(Set{"a", "z"}, two_word_connector: " to ")
+  #     # => a to z
+  #
+  #     to_sentence(1..3, last_word_connector: ", or ")
+  #     # => 1, 2, or 3
+  #
+  # NOTE: By default `#to_sentence` will include a
+  # [serial comma](https://en.wikipedia.org/wiki/Serial_comma). This can be
+  # overriden like so:
```
hanneskaeufler: I like the trivia about the oxford comma :) Not sure it warrants an extra `NOTE`, but also happy to see it here. `overriden => overridden`.

tommyschaefer: Ah! Thanks for the catch of the misspelling! Sorry that it snuck through!

As far as the note is concerned, I noticed there was a [comment on the Rails documentation](https://apidock.com/rails/Array/to_sentence#1497-Grammatical-error) expressing some concern about the Oxford comma, so I figured it might be helpful to mention it. Do you think this is worth including? 😄 If so, do you think it should it be under a `NOTE` (I saw that `NOTE`s get handled specially by `crystal docs`)?

hanneskaeufler: No worries about the typos, not much of a problem to have one sneak through :) I think the information per se is worth including, it makes the docs a bit more lively and interesting than just plain stating method documentation / arguments etc. Just my 2 cents :) Looking at the crystal api docs, the notes are more technically relevant I guess, so we should probably not overuse them, but 🤷‍♂️ I like it the more I think about it haha

paulcsmith: I like it a lot! It's a nice thing to know and it is helpful to know how to override it when necessary

PR #350, opened 2018-01-07
Hk button helper generates button

### LINKED ISSUE #346: `button` helper outputs a html link, which is confusing
If `button "Test", to: ButtonHelpers::Index` renders `<a href="/button_helpers">Test</a>` I am not sure I understand. How can I render an actual html button then? I think we should stick to representing the html tags more or less directly (sure, rename `p` to `para` if its a crystal keyword).

hanneskaeufler:
At first I just wanted to fix #346, naively (first commit). However, it occured to me that the behaviour makes no sense for `<button/>` tags. They actually don't concern themselves with hrefs/links. Instead a button is either of submit, reset or button type. The validation of the type is also the only thing warranting a separate `ButtonHelper` imho. I'd be just as happy to delete all that and add the `button` to the `base_tags`. What do you think?


paulcsmith:
@hanneskaeufler I think you're absolutely right about removing it and adding it to BaseTags instead. Nice research and good call 👍

Can you do that? If so, I'll leave this open. If not, LMK and I'll close this PR and the issue and create a new one.

Thanks, Hannes!

hanneskaeufler:
Yeah, I think that's the way to go. Removed it.

hanneskaeufler:
Btw I see the spec failure locally as well, seems unrelated to my changes though?
```
  1) Lucky::TimeHelpers distance_of_time_in_words reports the approximate distance in time between two Time
     Failure/Error: view.distance_of_time_in_words(from_time, from_time + 1.month).to_s.should eq "about a month"
       Expected: "about a month"
            got: "28 days"
```

PR #380, opened 2018-02-03
Add note about save_succeeded? -> saved? to upgrade notes

hanneskaeufler:
Sorry for the many tiny PRs.

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -119,6 +119,8 @@ UserForm.create!(name: "Jane")
 
 More info at https://luckyframework.org/guides/saving-with-forms/#passing-data-without-route-params
 
+* Change calls to `form.save_succeeded?` to `form.saved?`
+
 * Trap int in src/server.cr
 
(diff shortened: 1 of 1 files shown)


UPGRADE_NOTES.md
```diff
@@ -119,6 +119,8 @@ UserForm.create!(name: "Jane")
 
 More info at https://luckyframework.org/guides/saving-with-forms/#passing-data-without-route-params
 
+* Change calls to `form.save_succeeded?` to `form.saved?`
```
edwardloveall: I'd suggest saying `from` first here. So its:

> Change calls **from** `form.save_succeeded?` to `form.saved?`

paulcsmith: I agree. I think that reads a bit better

paulcsmith:
Thanks for these PRs. Small is great because they’re easy to review. So it’s no problem at all!

UPGRADE_NOTES.md
hanneskaeufler: ✅

PR #384, opened 2018-02-04
Add HR tag helper

andyw8:


hanneskaeufler:
Nice, do you see a use-case for a hr with content? If not it might be sufficient to add it to `EMPTY_TAGS` exclusively?

andyw8:
@hanneskaeufler good point, I'll update the PR.

andyw8:
@hanneskaeufler now updated.

paulcsmith:
Looks good! Thank you :)

PR #386, opened 2018-02-07
Show helpful error when using disallowed fields

paulcsmith:


src/lucky/tags/input_helpers.cr
```diff
@@ -1,15 +1,41 @@
...
+      You can't use this field because it is not allowed to be filled out.
+
+      Try this...
+
+        ▸ Allow people to fill out this field by adding 'allow {field_name}' to your form object.
+
+      Q. Why do I need to manually allow fields?
+      A. If Lucky allowed all fields, malicious users could submit any field they want.
+         For example: you might have an 'admin' flag on a User. If all fields were
+         allowed, a malicious user could set the 'admin' flag to 'true' for any form.
```
edwardloveall: I like this example, but the wording gets muddled. It's hard to distinguish HTML input field from a Lucky field. What do you think about:

> If all fields were allowed, a malicious user could add data that sets the 'admin' flag to 'true' for any form.

PR #395, opened 2018-03-02
Add documentation for Lucky::ContentTypeHelpers

edwardloveall:
Getting my feet wet with some documentation (#275). These are pretty simple methods but I think the expanded context of what these helpers are looking for could help.

I put two blank lines above each method's description so that the documentation would only show up on in the details method docs, not in the summary, which looked really cluttered. Happy to change it though if someone thinks it should be in both.

### DIFF
--- a/src/lucky/content_type_helpers.cr
+++ b/src/lucky/content_type_helpers.cr
@@ -1,19 +1,35 @@
+# These helpers check HTTP headers to determine "content type". Most check the
+# Content-Type header, check other information, such as the `X-Requested-With`
+# header.
 module Lucky::ContentTypeHelpers
   abstract def content_type : String
   abstract def headers : HTTP::Headers
 
+  # Check if the request is JSON
+  #
+  # This tests if the Content-Type header is `application/json`
   def json?
     content_type == "application/json"
   end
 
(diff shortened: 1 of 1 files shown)


src/lucky/content_type_helpers.cr
```diff
@@ -1,19 +1,34 @@
+# These helpers check HTTP headers to determine "content type". Most check the
+# Content-Type header, but some, such as `ajax?` check other fields
```
paulcsmith: maybe "check other information, such as the `X-Requested-With` header"?

src/lucky/content_type_helpers.cr
```diff
@@ -1,19 +1,34 @@
...
+# (`X-Requested-With`).
 module Lucky::ContentTypeHelpers
   abstract def content_type : String
   abstract def headers : HTTP::Headers
 
+  #
```
paulcsmith: I think a summary on these might look a bit better. Something like: "Check if the request is JSON/AJAZ/XML/etc."

edwardloveall:
Yup, those were good ideas. Thanks for the feedback!

PR #405, opened 2018-03-30
Add documentation for the expose macro

edwardloveall:
This finally clicked for me when I started thinking about expose being used in a base class that is inherited from. Even if that's not the final goal, it made me get the use case.

Related to, but not completing by any stretch of the imagination, #275.

### DIFF
--- a/src/lucky/exposable.cr
+++ b/src/lucky/exposable.cr
@@ -78,6 +78,33 @@ module Lucky::Exposeable
   #   end
   # end
   # ```
+  #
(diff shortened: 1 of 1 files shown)


paulcsmith:
You may be interested in other examples of `expose` here: https://github.com/paulcsmith/lucky-auth This is the sample app I'm doing for auth. I'll extract the code to a library and the Lucky CLI generators soon :D

edwardloveall:
Oh, yeah I like that private method thing. I'll add that.

Issue #408, opened 2018-04-04
Add `lucky watch --reload-browser`


mikeeus:
As discussed in [this pull request lucky-cli](https://github.com/luckyframework/lucky_cli/pull/175#issuecomment-378288754). The `--reload-browser` flag will add browser-sync to the `lucky watch` task.

PR #411, opened 2018-04-04
Split action generators into `action.api` and `action.browser`

### LINKED ISSUE #410: Seperate `browser` and `api` action generators
As discussed [in this luck_cli issue](https://github.com/luckyframework/lucky_cli/pull/175#issuecomment-378288754), there should be separate generators for browser and api actions.

mikeeus:
Closes #410.

With the new behavior you need to run `lucky gen.action.api` or `lucky gen.action.browser`.

### DIFF
--- a/spec/tasks/gen/action_spec.cr
+++ b/spec/tasks/gen/action_spec.cr
@@ -5,11 +5,8 @@ include CleanupHelper
 describe Gen::Action do
   it "generates a basic browser action" do
     with_cleanup do
-      io = IO::Memory.new
       valid_action_name = "Users::Index"
-      ARGV.push(valid_action_name)
-
-      Gen::Action::Browser.new.call(io)
+      io = generate valid_action_name, Gen::Action::Browser
 
       index_action = File.read("./src/actions/users/index.cr")
       index_action.should contain(valid_action_name)
@@ -22,11 +19,8 @@ describe Gen::Action do
 
   it "generates a basic api action" do
     with_cleanup do
-      io = IO::Memory.new
       valid_action_name = "Users::Index"
-      ARGV.push(valid_action_name)
-
-      Gen::Action::Api.new.call(io)
+      io = generate valid_action_name, Gen::Action::Api
 
       index_action = File.read("./src/actions/users/index.cr")
       index_action.should contain(valid_action_name)
@@ -39,11 +33,8 @@ describe Gen::Action do
 
   it "generates nested browser and api actions" do
     with_cleanup do
-      io = IO::Memory.new
       valid_nested_action_name = "Users::Announcements::Index"
-      ARGV.push(valid_nested_action_name)
-
-      Gen::Action::Browser.new.call(io)
+      io = generate valid_nested_action_name, Gen::Action::Browser
 
       index_action = File.read("src/actions/users/announcements/index.cr")
       index_action.should contain(valid_nested_action_name)
@@ -54,11 +45,8 @@ describe Gen::Action do
     end
 
     with_cleanup do
-      io = IO::Memory.new
       valid_nested_action_name = "Users::Announcements::Index"
-      ARGV.push(valid_nested_action_name)
-
-      Gen::Action::Api.new.call(io)
+      io = generate valid_nested_action_name, Gen::Action::Api
 
       index_action = File.read("src/actions/users/announcements/index.cr")
       index_action.should contain(valid_nested_action_name)
@@ -70,11 +58,8 @@ describe Gen::Action do
 
     it "snake cases filenames of a camel case action" do
       with_cleanup do
-        io = IO::Memory.new
         valid_camel_case_action_name = "Users::HostedEvents"
-        ARGV.push(valid_camel_case_action_name)
-
-        Gen::Action::Browser.new.call(io)
+        io = generate valid_camel_case_action_name, Gen::Action::Browser
 
         File.read("src/actions/users/hosted_events.cr")
             .should contain(valid_camel_case_action_name)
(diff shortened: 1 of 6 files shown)


tasks/gen/action_generator.cr
```diff
@@ -1,24 +1,26 @@
...
   end
 end
 
-class Gen::Action < LuckyCli::Task
+class Gen::ActionGenerator < LuckyCli::Task
   banner "Generate a new action"
 
-  def call(io : IO = STDOUT)
+  def call
+  end
```
paulcsmith: To get around this I believe you can make this class `abstract` That means that it can't be instantiated by itself, and that it doesn't need to implement the methods yet

`abstract class Gen::...`

spec/tasks/gen/action_spec.cr
```diff
@@ -3,31 +3,35 @@ require "../../spec_helper"
 include CleanupHelper
 
 describe Gen::Action do
-  it "generates a basic action" do
+  it "generates a basic browser action" do
```
paulcsmith: I think it would be helpful to also have a test for basic (non-nested) API actions as well

spec/tasks/gen/action_spec.cr
```diff
@@ -3,31 +3,35 @@ require "../../spec_helper"
...
 
-      File.read("./src/actions/users/index.cr")
-          .should contain(valid_action_name)
       io.to_s.should contain(valid_action_name)
       io.to_s.should contain("/src/actions/users")
     end
   end
 
-  it "generates a nested action" do
+  it "generates a nested api action" do
```
paulcsmith: I think it would be helpful to have a test for generating nested browser actions as well

spec/tasks/gen/action_spec.cr
```diff
@@ -3,31 +3,35 @@ require "../../spec_helper"
...
       io = IO::Memory.new
       valid_action_name = "Users::Index"
       ARGV.push(valid_action_name)
 
-      Gen::Action.new.call(io)
+      Gen::Action::Browser.new.call(io)
+
+      index_action = File.read("./src/actions/users/index.cr")
+      index_action.should contain(valid_action_name)
+      index_action.should contain("< BrowserAction")
```
paulcsmith: Since so much of these tests are duplicated, I think it would be helpful to extract a helper object for this. Here's what I'm thinking:

```crystal
# The `generate` method would generate the `io`, set the `ARGV` and call the generator and return the `io`
io = generate "Users::Index", with: Gen::Action::Browser 

# Custom expectation helper:
should_have_generated "#{valid_action_name} < BrowserAction", in: "./src/actions/users/index.cr"
```

paulcsmith:
Thanks so much for working on this!

tasks/gen/action_generator.cr
mikeeus: I think that because it inherits from `LuckyCli::Task` it gets instantiated automatically when `tasks.cr` requires the tasks. So when I make it an abstract class it spits this error out:

```
Error in line 1: while requiring "./spec/http_method_override_handler_spec.cr"

in spec/http_method_override_handler_spec.cr:1: while requiring "./spec_helper"

require "./spec_helper"
^

in spec/spec_helper.cr:3: while requiring "../tasks/**"

require "../tasks/**"
^

in tasks/gen/action_generator.cr:15: expanding macro

abstract class Gen::ActionGenerator < LuckyCli::Task
^

in macro 'inherited' /home/mikmak/crystal/lib/lucky/lib/lucky_cli/src/lucky_cli/task.cr:2, line 1:

>  1.     LuckyCli::Runner.tasks << self.new
   2. 
   3.     def name
   4.       "gen.action_generator"
   5.     end
   6.   

instantiating 'Gen::ActionGenerator:Class#new()'
in /usr/share/crystal/src/primitives.cr:36: can't instantiate abstract class Gen::ActionGenerator

class Reference
^
```

spec/tasks/gen/action_spec.cr
mikeeus: Sure, I'll duplicate the top test

tasks/gen/action_generator.cr
paulcsmith: Ah I see. That is something that should change in LuckyCLI.

This line:
https://github.com/luckyframework/lucky_cli/blob/master/src/lucky_cli/task.cr#L3

Should only be run if the type is *not* abstract, since abstract classes can't be instantiated. I'll add an issue.

I think for now to work around this you could instead make that class not inherit from `LuckyCLI::Task` and instead call it from within the generators like this: `Lucky::ActionGenerator.new(action_name : String, inherit_from : String).generate`

Does that make sense?

spec/tasks/gen/action_spec.cr
mikeeus: Yeah this is a good idea, it makes the tests clean and easy to read

mikeeus:
No problem, I'm done for tonight but I'll check again in the morning if I need to clean anything up. :)

tasks/gen/action_generator.cr
paulcsmith: @mikeeus I think this is the only part that should change. Other than that, this looks great! The reason it shouldn't inherit from LuckyCLI is that it will be listed as a task when running `lucky --help`, even though you can't run it. So extracting it to it's own class would fix that

mikeeus:
Yes not inheriting from `LuckyCli::Task` makes sense. I tried using an abstract class but it was giving me issues by not parsing `ARGV` to get the task name properly:
```
Index out of bounds
       /usr/share/crystal/src/indexable.cr:0:13 in 'first'
       tasks/gen/action/browser.cr:9:5 in 'action_name'
       tasks/gen/action/browser.cr:5:30 in 'call'
       /action_spec.cr:85:3 in 'generate'
```
So instead I opted to make `ActionGenerator` a module. What do you think?

---
Also I found a problem when running `lucky --help`. In issue #408 I called `parse_options` in `Sentry::ProcessRunner#initialize` by mistake:
https://github.com/luckyframework/lucky/blob/d5188f245be8ab29d8e6b03e87b65709a75314f9/tasks/watch.cr#L24-L26

When it should be called here in `Watch#call` instead since the `parse_option` method is declared there:
https://github.com/luckyframework/lucky/blob/d5188f245be8ab29d8e6b03e87b65709a75314f9/tasks/watch.cr#L141-L146

So I made that fix.

tasks/gen/action/browser.cr
```diff
@@ -0,0 +1,9 @@
+class Gen::Action::Browser < LuckyCli::Task
+  include Gen::ActionGenerator
```
paulcsmith: I like the module 👍

PR #412, opened 2018-04-04
Add --reload-browser flag to conditionally run browser-sync

### LINKED ISSUE #408: Add `lucky watch --reload-browser`

mikeeus:
Closes #408.

I used OptionParser to set a `@reload_browser` property and use that to conditionally start/reload browser-sync in the `create_app_processes` method.

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -16,7 +16,7 @@ module Sentry
     getter app_processes = [] of Process
     property successful_compilations
     property app_built
-    property reload_browser
+    property? reload_browser
 
     @app_built : Bool = false
     @successful_compilations : Int32 = 0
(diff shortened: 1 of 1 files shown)


tasks/watch.cr
```diff
@@ -38,10 +41,12 @@ module Sentry
...
-      if successful_compilations == 1
-        start_browsersync
-      else
-        reload_browsersync
+      if @reload_browser
+        if successful_compilations == 1
+          start_browsersync
+        else
+          reload_browsersync
+        end
```
paulcsmith: I think extracting a `reload_or_start_browser_sync` method would be nice so this is cleaned up a bit

tasks/watch.cr
```diff
@@ -15,11 +16,13 @@ module Sentry
     getter app_processes = [] of Process
     property successful_compilations
     property app_built
+    property reload_browser
```
paulcsmith: If you do `property?` you can then call `reload_browser?` which reads a bit better IMO

tasks/watch.cr
```diff
@@ -38,10 +41,12 @@ module Sentry
       end
 
       self.successful_compilations += 1
-      if successful_compilations == 1
-        start_browsersync
-      else
-        reload_browsersync
+      if @reload_browser
```
paulcsmith: Then you can use `if reload_browser?`

tasks/watch.cr
mikeeus: That's neat, it definitely reads better

PR #417, opened 2018-04-06
Routable docs

edwardloveall:
I [added](https://github.com/luckyframework/lucky/commit/e8340019fa907abdeef9cdac88f097e835e55853) `:nodoc:` to a few methods that I think aren't intended to be accessed, but let me know if that's not true.

One of the methods was `add_route`. While I don't _think_ we expect people to use `add_route`, it does contain used methods like `with`, `path`, and `url`. I tried adding documentation directly to those methods but they wouldn't show up since they are generated. Maybe it doesn't matter? Maybe there's another spot for them? Maybe there's a way to add documentation without a method to attach to?

Suggestions on that and my copy welcome 😄

### DIFF
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -1,3 +1,4 @@
+# Methods for routing HTTP requests and their parameters to actions.
 module Lucky::Routeable
   {% for http_method in [:get, :put, :post, :delete] %}
     # Define a route that responds to a {{ http_method.id.upcase }} request
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -1,5 +1,17 @@
 module Lucky::Routeable
   {% for http_method in [:get, :put, :post, :delete] %}
+    # Define a route that responds to a {{ http_method.id.upcase }} request
+    #
+    # Use these methods if you are not using a restful route or a custom path.
+    # For example:
+    # ```
+    # class Profile::ImageUpload
+    #   {{ http_method.id }} "/profile/image/:id" do
+    #     # action code here
+    #   end
+    # end
+    # ```
+    # will respond to an `HTTP {{ http_method.id.upcase }}` request.
     macro {{ http_method.id }}(path)
       add_route :{{ http_method.id }}, \{{ path }}, \{{ @type.name.id }}
 
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -39,6 +39,21 @@ module Lucky::Routeable
     end
   end
 
+  # Define a nested route that responds to the appropriate HTTP request
+  # automatically
+  #
+  # This works similarly to `action` but it will provide multiple parameters.
+  # For example:
+  # ```
+  # class Posts::Comments::Show
+  #   nested_action do
+  #     render_text "Post: #{post_id}, Comment: #{id}"
+  #   end
+  # end
+  # ```
+  #
+  # **Note:** The `singular` option will likely be removed soon. Try `get`,
+  # `post`, `put`, and `delete` with a custom path instead.
   macro nested_action(singular = false)
     infer_nested_route(singular: {{ singular }})
 
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -72,7 +72,17 @@ module Lucky::Routeable
 
   # Define a route that responds to the appropriate HTTP request automatically
   #
-  # A route needs a few pieces of information to be created:
+  # ```
+  # class Posts::Show
+  #   action do
+  #     render_text "Post: #{id}"
+  #   end
+  # end
+  # ```
+  #
+  # This action responds to the `/posts/:id` path.
+  #
+  # Each route needs a few pieces of information to be created:
   #
   # * The HTTP method, like `GET`, `POST`, `DELETE`, etc.
   # * The path, such as `/users/:id`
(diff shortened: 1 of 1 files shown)


src/lucky/routeable.cr
```diff
@@ -1,12 +1,25 @@
 module Lucky::Routeable
   {% for http_method in [:get, :put, :post, :delete] %}
+    # Define a route that responds to a {{ http_method.id.upcase }} request
+    #
+    # Use these methods if you are not using a restful route or a custom path.
```
paulcsmith: This makes it sound like you use this if you don't need a custom path. Maybe flip it? `Use these methods if you need a custom path or if you are not using a restful route`

src/lucky/routeable.cr
```diff
@@ -1,12 +1,25 @@
...
+    # For example:
+    # ```
```
paulcsmith: Newline above this I think

src/lucky/routeable.cr
```diff
@@ -1,12 +1,25 @@
 module Lucky::Routeable
   {% for http_method in [:get, :put, :post, :delete] %}
+    # Define a route that responds to a {{ http_method.id.upcase }} request
+    #
```
paulcsmith: I think linking to the guides would be handy as well, so they can get more info

src/lucky/routeable.cr
```diff
@@ -27,26 +40,71 @@ module Lucky::Routeable
     end
   end
 
+  # Define a nested route that responds to the appropriate HTTP request
+  # automatically
+  #
+  # This works similarly to `action` but it will provide multiple parameters.
+  # For example:
```
paulcsmith: Newline after this

src/lucky/routeable.cr
```diff
@@ -27,26 +40,71 @@ module Lucky::Routeable
...
+  # ```
+  # class Posts::Comments::Show
+  #   nested_action do
+  #     render_text "Post: #{post_id}, Comment: #{id}"
+  #   end
+  # end
+  # ```
+  #
+  # **Note:** The `singular` option will likely be removed soon. Try `get`,
+  # `post`, `put`, and `delete` with a custom path instead.
```
paulcsmith: Nice 👍

src/lucky/routeable.cr
```diff
@@ -27,26 +40,71 @@ module Lucky::Routeable
...
   end
 
+  # Define a nested route that responds to the appropriate HTTP request
+  # automatically
+  #
+  # This works similarly to `action` but it will provide multiple parameters.
+  # For example:
+  # ```
+  # class Posts::Comments::Show
+  #   nested_action do
```
paulcsmith: I think it would be good to also show what path this matches: `/posts/:post_id/comments/:id`

src/lucky/routeable.cr
```diff
@@ -27,26 +40,71 @@ module Lucky::Routeable
...
+  # class name to figure out the path, i.e. removing the `::` separators and
+  # adding underscores. The method is found via the last part of the class name:
+  #
+  # * `Index` -> `GET`
+  # * `Show` -> `GET`
+  # * `New` -> `GET`
+  # * `Create` -> `POST`
+  # * `Edit` -> `GET`
+  # * `Update` -> `PUT`
+  # * `Delete` -> `DELETE`
```
paulcsmith: It would be good to link here as well: https://luckyframework.org/guides/actions-and-routing/#automatically-generate-restful-routes

And/or use the list here: https://luckyframework.org/guides/actions-and-routing/#examples-of-automatically-generated-routes

src/lucky/routeable.cr
```diff
@@ -138,6 +197,30 @@ module Lucky::Routeable
     \{% end %}
   end
 
+  # Access a query parameter
+  #
+  # This will allow any action to accept and access query parameters. By
+  # default query parameters are ignored, but adding a `param` will make them
```
paulcsmith: I'm not sure I understand the "by default a query paramter will be ignored. It might be more clear to omit that phrase and instead say something like: `It will set up a method that can be used to access the parameter. If the type of the param does not allow `Nil`, then the query parameter will be required when generating paths and urls.`

src/lucky/routeable.cr
```diff
@@ -138,6 +197,30 @@ module Lucky::Routeable
...
+  #
+  # Visiting `/posts?page=10` would render the following:
+  #
+  # ```text
+  # Posts - Page 10
+  # ```
+  #
+  # Additionally, these parameters are typed. The path `/posts?page=ten` will
+  # raise a `Lucky::Exceptions::InvalidParam` error because `ten` is a string
+  # not an Int32.
```
paulcsmith: An example of generating a link would be good as well: `Posts::Index.with(2) # => "/posts?page=2"`

Maybe an example of a nilable para, as well?

paulcsmith:
Thanks for tackling this. This is gonna be really helpful

src/lucky/routeable.cr
edwardloveall: I think I was trying to say that if you don't use `param` any query parameter in the url is not in some params object and therefore inaccessible. The "setup a method" bit still feels a little but too much like describing the implementation not the effect to me, but I agree with your sentiment. I'll find another way.

src/lucky/routeable.cr
edwardloveall: Oh! I didn't know you could do that with `with`

src/lucky/routeable.cr
paulcsmith: Ah, I see. This isn't _quite_ right though. You can access params with `params.get(:page)`. It just isn't as safe, because it may or may not be there

edwardloveall: Ah gotcha. Ok, I'll incorporate that also.

src/lucky/routeable.cr
edwardloveall: Looking through the code, it looks like you can only use `with` to generate a query param if it's required. How would I generate a nillable param?

```crystal
query_params = {} of String => String
{% for param in PARAM_DECLARATIONS %}
  param_is_default_or_nil = {{ param.var }} == {{ param.value || nil }}
  unless param_is_default_or_nil
    query_params["{{ param.var }}"] = {{ param.var }}.to_s
  end
{% end %}
```

paulcsmith:
Thanks @edwardloveall  :D

PR #418, opened 2018-04-06
Add /docs to the gitignore instead of /doc

edwardloveall:
My guess is this was added in error or perhaps crystal changed the
generated directory at some point. Either way, /docs is the directory
that is generated, not /doc.


paulcsmith:
Whoops!

PR #433, opened 2018-04-25
Add documentation for the render macro

edwardloveall:
There are a couple methods (`render_html_page` and `perform_action`) that didn't seem to be useful to document, but lemme know if they should be used.

On the flipside, there are a few methods (`head`, `json`, `text`) that might be useful to document. But they are private and crystal does not render documentation for private methods.

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -1,4 +1,47 @@
 module Lucky::Renderable
+  # Create a page for an action to display
+  #
+  # `render` takes two arguments, a `page_class` and data to send to the page.
+  # The `page_class` is automatically extracted from the action. For example,
+  # the `Users::Index` action is converted into `Users::IndexPage`. You can
+  # also pass a page class directly if you need to. For example the
+  # `Users::Search` action can render the `Users::IndexPage`:
+  #
+  # ```crystal
+  # class Users::Search < BrowserAction
+  #   action do
+  #     # search for users
+  #
+  #     render Users::IndexPage, users: user_search_results
+  #   end
+  # end
+  # ```
+  #
+  # The second argument to `render` is used to pass data to the page. Each
+  # key/value pair must match up with the `needs` declarations for the page.
+  # For example, if we have a page like this:
+  #
+  # ```crystal
+  # class Users::IndexPage < MainLayout
+  #   needs users : UserQuery
+  #
+  #   def content
+  #     @users.each do |user|
+  #       # ...
+  #     end
+  #   end
+  # end
+  # ```
+  #
+  # Our action must use pass a `users` key to the `render` method like this:
+  #
+  # ```crystal
+  # class Users::Index < BrowserAction
+  #   action do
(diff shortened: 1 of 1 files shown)


src/lucky/renderable.cr
```diff
@@ -1,4 +1,47 @@
...
+  #
+  # Our action must use pass a `users` key to the `render` method like this:
+  #
+  # ```crystal
+  # class Users::Index < BrowserAction
+  #   action do
+  #     render users: UserQuery.new
+  #   end
+  # end
+  # ```
```
paulcsmith: Could you remove the docs where the page class is omitted? I'll be removing the page class inference part of this macro soon, and that will hopefully enable this to use a regular method instead of a macro.

src/lucky/renderable.cr
```diff
@@ -1,25 +1,9 @@
...
-  #     render Users::IndexPage, users: user_search_results
-  #   end
-  # end
-  # ```
-  #
-  # The second argument to `render` is used to pass data to the page. Each
-  # key/value pair must match up with the `needs` declarations for the page.
-  # For example, if we have a page like this:
+  # `render` is used to pass data to a page as a hash. Each key/value pair must
+  # match up with each `needs` declarations for that page. For example, if we
```
paulcsmith: I really like how you described the key/value matching up with needs declarations 👍

src/lucky/renderable.cr
```diff
@@ -1,25 +1,9 @@
...
-  #
-  #     render Users::IndexPage, users: user_search_results
-  #   end
-  # end
-  # ```
-  #
-  # The second argument to `render` is used to pass data to the page. Each
-  # key/value pair must match up with the `needs` declarations for the page.
-  # For example, if we have a page like this:
+  # `render` is used to pass data to a page as a hash. Each key/value pair must
```
paulcsmith: I think you can omit the "as a hash". It's not really a hash under the hood. It's a NamedTuple that is used to pass args to the initialize method. But that's probably not important. So just "`render` is used to pass data to a page" would probably suffice

src/lucky/renderable.cr
```diff
@@ -42,6 +26,23 @@ module Lucky::Renderable
   #   end
   # end
   # ```
+  #
+  # Note also that each peice of data is merged with any `expose` declarations:
```
paulcsmith: *piece

src/lucky/renderable.cr
```diff
@@ -42,6 +26,23 @@ module Lucky::Renderable
...
+  #
+  # ```crystal
+  # class Users::Index < BrowserAction
+  #   expose current_user
+  #
+  #   action do
+  #     # page recieves users AND current_user
+  #     render users: UserQuery.new
```
paulcsmith: I love this example 👍

edwardloveall:
Done! Thanks for the feedback.

paulcsmith:
Thanks!

PR #442, opened 2018-04-30
Handle 404s

russ:


src/lucky/server_settings.cr
```diff
@@ -4,7 +4,7 @@ module Lucky::ServerSettings
   extend self
 
   def host : String
-    settings["host"].as_s
+    ENV["HOST"]?.try(&.to_s) || settings["host"].as_s
```
paulcsmith: I'd have to try this out first to make it work as expected. Could you make this a separate PR with some details about what the use case is and how you're using it? I think I have a rough idea, but I always like to check before making changes to default config

src/lucky/errors.cr
```diff
@@ -7,8 +7,10 @@ module Lucky
 
   # Raised when a route could not be found
   class RouteNotFoundError < Error
-    def initialize(method : String, path : String)
-      super "Could not find route matching #{method} #{path}"
+    property context : HTTP::Server::Context
+
+    def initialize(@context : HTTP::Server::Context)
```
paulcsmith: I think this could also accept the params so it is easier to get at specific params when needed: `@params : Lucky::Params` would be the type signature

src/lucky/errors.cr
```diff
@@ -7,8 +7,10 @@ module Lucky
 
   # Raised when a route could not be found
   class RouteNotFoundError < Error
-    def initialize(method : String, path : String)
-      super "Could not find route matching #{method} #{path}"
+    property context : HTTP::Server::Context
```
paulcsmith: This can be a `getter` since it shouldn't be set outside this class. It also doesn't need a type since you defined it in the initialize. So in the end it could be: `getter context, params`

paulcsmith:
I made some changes in a separate PR based on your work. Thanks for starting this! https://github.com/luckyframework/lucky/pull/461

russ:
Awesome! Sorry about the delay. I'm working on some Crystal code on the side so I can get more comfortable with it. Best I do that before I start trying to deep dive into more projects.

Issue #444, opened 2018-05-01
Path without starting slash compiles fine, but yields 404 on visiting intended path


neurodynamic:
```
  get "sign_up" do
    text "Render something in Users::New"
  end
```

This code compiles fine, but creates an impossible `sign_up` route instead of a working `/sign_up` route. Easy mistake to make. Some sort of error on this kind of thing would definitely be helpful.

paulcsmith:
I totally agree! If you'd like to try to do this I can give you some pointers along the way, but if you don't have time I can tackle this in the next few days.

neurodynamic:
@paulcsmith I'd be happy to give it a shot. Do you think it'd be better to have the behavior be raising something like a `BadRoute` exception with an appropriate message, or to modify the route to automatically add the slash if it's missing (assuming there aren't circumstances where that would break something that otherwise worked?)?

paulcsmith:
I think it would be best to raise an error at compile time so that all the routes are consistent in the code. To do this you'll need to modify the `get/post/put/delete` macros:

* They are located here: https://github.com/luckyframework/lucky/blob/master/src/lucky/routeable.cr#L21
* At the very top of the method you can do something like this:

```crystal
{% unless path.starts_with?("/") %}
  {% raise "paths must start with '/'" %}
{% end %}
```

You can't really have an automated test for this (AFAIK) since it will cause a compiler error. So to test it you'll have to change one of the actions in spec/action_spec.cr to not start with a slash and see what the error looks like. I hope this makes sense. Ping me if you run into any trouble, and thanks for taking a shot at this!

PR #445, opened 2018-05-01
Adds "Paths must start with a slash" error to route methods

neurodynamic:
Resolves https://github.com/luckyframework/lucky/issues/444

This seems to work. I added a "For example" part to the message. Using string interpolation to put `path` in the error message kept putting escaped double-quotes `\"` into the resulting message, but for some reason concatenation with `+` did not, so I used that. I'm new to Crystal, so this may be some standard type of behavior I'm not aware of.

Thanks for your help!

### DIFF
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -20,7 +20,7 @@ module Lucky::Routeable
     # * [Routing](https://luckyframework.org/guides/actions-and-routing/#routing)
     macro {{ http_method.id }}(path)
       \{% unless path.starts_with?("/") %}
-        \{% path.raise "Paths must start with a slash. For example, '" + path + "' is not a valid path, but '/" + path + "' is." %}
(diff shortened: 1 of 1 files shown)


src/lucky/routeable.cr
```diff
@@ -19,6 +19,10 @@ module Lucky::Routeable
     # **See also** our guides for more information and examples:
     # * [Routing](https://luckyframework.org/guides/actions-and-routing/#routing)
     macro {{ http_method.id }}(path)
+      \{% unless path.starts_with?("/") %}
+        \{% path.raise "Paths must start with a slash. For example, '" + path + "' is not a valid path, but '/" + path + "' is." %}
```
paulcsmith: I think adding the example will definitely help. I think it could be shortened to this: `"Path must start with a slash. Example: \"/#{path}\""`. Also note the typo Paths -> Path at the beginning

paulcsmith:
Awesome. Thank you!

PR #446, opened 2018-05-01
Lucky::Params docs

edwardloveall:
These are pretty plain docs, but I didn't see much functionality or edge cases that needed to be explained. Please let me know if I missed anything 🙂 

Also, I made some methods private that didn't look like they were referenced anywhere else in the codebase.

### DIFF
--- a/src/lucky/params.cr
+++ b/src/lucky/params.cr
@@ -4,28 +4,84 @@ class Lucky::Params
   @request : HTTP::Request
   @route_params : Hash(String, String) = {} of String => String
 
+  # :nodoc:
   getter :request
+  # :nodoc:
   getter :route_params
 
+  # Create a new params object
+  #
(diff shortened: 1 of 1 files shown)


src/lucky/params.cr
```diff
@@ -4,28 +4,84 @@ class Lucky::Params
...
+  #
+  # ```crystal
+  # params.get?("page")    # 1 : (String | Nil)
+  # params.get?("missing") # nil : (String | Nil)
+  # ```
   def get?(key : String | Symbol) : String?
     route_params[key.to_s]? || body_param(key.to_s) || query_params[key.to_s]?
   end
 
+  # Retrive a file from the params hash, raise if key is absent
```
paulcsmith: Retrieve is misspelled in a couple spots

edwardloveall:
Hup, thank you. Fixed!

paulcsmith:
Thanks :D

PR #458, opened 2018-05-09
Make the 500 page return a 500 in production. fixes #457

### LINKED ISSUE #457: 500 error page returns 200 status in production
If your app throws a 500 error in production, the nice lucky page that is rendered returns a 200 status from the server. This causes issues with some load balancers that do certain things based on the status code returned. 

If you want to try this locally, you can do:
```
crystal build src/server.cr --release --no-debug -o app
LUCKY_ENV=production APP_DOMAIN=whatever.com SECRET_KEY_BASE=abc123 PORT=4000 ./app
```

Then just hit your `whatever.lvh.me:4000/your/error/page` in your browser and see the 200.

screen shot 2018-05-09 at 3 01 29 pm

jwoertink:
Ensure the 500 page rendered in production also returns a 500 status code. It already returns a 500 status in development with specs around it. I wasn't sure how to add a spec for this. I don't think it's a huge deal, but if you want a spec, I can try and figure that out.

### DIFF
--- a/spec/lucky/error_handling_spec.cr
+++ b/spec/lucky/error_handling_spec.cr
@@ -14,7 +14,7 @@ private class FakeErrorAction < Lucky::ErrorAction
   end
 
   def handle_error(error : Exception)
-    head status: 500
+    text "Oops"
   end
 end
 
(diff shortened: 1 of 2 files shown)


jwoertink:
Changing that line makes it fail with

```

paulcsmith:
That is ok since it is now returning text instead of an empty content type. So you can change the expectation to expect `"text/plain"` and that would be ok

PR #478, opened 2018-05-25
Add callback documentation (after/before)

edwardloveall:


src/lucky/action_callbacks.cr
```diff
@@ -24,14 +27,64 @@ module Lucky::ActionCallbacks
     \{% end %}
   end
 
+  # Run a method before an action starts
+  #
+  # Methods will run in the order that each `before` is defined. Also, each
```
edwardloveall: Is the order something you guarantee by the way? Right now that's just how the implementation works but I wasn't sure if it was intentional or a side-effect.

If it's intentional maybe we should have a test for it.

paulcsmith: That's a really good point. It is intentional, but it'd be good to have a test to ensure that it stays that way

src/lucky/action_callbacks.cr
```diff
@@ -24,14 +27,64 @@ module Lucky::ActionCallbacks
...
+  # method must return a `Lucky::Response` such as `redirect` or `render`, or
```
paulcsmith: Rather than having 2 `or`s I think it'd look cleaner to have `redirect`, `render`, or ...

src/lucky/action_callbacks.cr
```diff
@@ -24,14 +27,64 @@ module Lucky::ActionCallbacks
...
+  #   before check_if_signed_in
+  #   before confirm_destroy
+  #
+  #   delete "/:user_id" do
+  #     # destroy the user :(
+  #   end
+  #
+  #   def check_if_signed_in
+  #     if current_user.nil?
+  #       redirect to: SignInPage
```
paulcsmith: This needs an `else` -> `continue` branch

src/lucky/action_callbacks.cr
edwardloveall: I can only think of hacky ways to test like having callbacks append to the same file and then checking the line order. Got any better ideas? Can we mock things method calls?

paulcsmith: I explicitly checked order in LuckyRecord here: https://github.com/luckyframework/lucky_record/blob/master/spec/form_callbacks_spec.cr

An instance var with an array of strings that you append to in each callback. Then you can check that the instance var has the correct strings, e.g. `callbacks_that_run.should eq ["first_before", "second_before"]`

paulcsmith: That one also explicitly checks that it works for symbols or bare words

edwardloveall:
Updated. I'll see if I can figure out a way to test the order dependence

src/lucky/action_callbacks.cr
```diff
@@ -24,14 +27,64 @@ module Lucky::ActionCallbacks
     \{% end %}
   end
 
+  # Run a method before an action starts
+  #
+  # Methods will run in the order that each `before` is defined. Also, each
+  # method must call `redirect`, `render`, or `continue`:
```
paulcsmith: Actually just realized this may need to be clarified. Smething like: 

> the callback must return a `LuckyWeb::Response` using `render`, `redirect`, `json`, etc., or if you don't want to render anything `continue`

This is horribly worded, but the idea is that it says it must return a `LuckyWeb::Response`. That could be with render, redirect, json, head, xml or anything else like that. I don't want people to get confused and think they can only render or redirect. Does that make sense?

edwardloveall: Yeah, I gotcha. I'll futz with it

edwardloveall:
Alright, tested. Good call on the instance variable idea. That worked great!

PR #492, opened 2018-05-30
adding in nested params check for query params. Fixes #491

### LINKED ISSUE #491: Lucky Params doesn't handle nested query params
```crystal
request = build_request body: "", content_type: "" #helper taken from specs
request.query = "filter:name=taco&page=1&per=50"
params = Lucky::Params.new(request)
params.nested?("filter") #=> {}
```

This works fine for [json](https://github.com/luckyframework/lucky/blob/master/src/lucky/params.cr#L146), [form, or multipart](https://github.com/luckyframework/lucky/blob/master/src/lucky/params.cr#L158), but missing for the query.

jwoertink:
This is so we can check nested params in a query string.

### DIFF
--- a/spec/lucky/params_spec.cr
+++ b/spec/lucky/params_spec.cr
@@ -126,7 +126,6 @@ describe Lucky::Params do
     it "gets form encoded nested params" do
       request = build_request body: "user:name=paul&user:twitter_handle=@paulcsmith&something:else=1",
         content_type: "application/x-www-form-urlencoded"
-      request.query = "from=query"
 
       params = Lucky::Params.new(request)
 
@@ -154,6 +153,13 @@ describe Lucky::Params do
       params = Lucky::Params.new(request)
 
       params.nested(:user).should eq({"name" => "Paul", "age" => "28"})
+    end
+
+    it "gets query nested params" do
+      request = build_request body: "", content_type: ""
+      request.query = "filter:query=pizza&sort=desc"
+      params = Lucky::Params.new(request)
+      params.nested?("filter").should eq({"query" => "pizza"})
     end
 
     it "gets nested params after unescaping" do
--- a/spec/lucky/params_spec.cr
+++ b/spec/lucky/params_spec.cr
@@ -146,7 +146,7 @@ describe Lucky::Params do
       request = build_request body: {user: {name: "Bunyan", age: 102}}.to_json,
         content_type: "application/json"
       request.query = "user:active=true"
-      
+
       params = Lucky::Params.new(request)
 
       params.nested?(:user).should eq({"name" => "Bunyan", "age" => "102", "active" => "true"})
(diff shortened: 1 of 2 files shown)


src/lucky/params.cr
```diff
@@ -108,7 +108,7 @@ class Lucky::Params
     if json?
       nested_json_params(nested_key.to_s)
     else
-      nested_form_params(nested_key.to_s)
+      query_params.any? ? nested_query_params(nested_key.to_s) : nested_form_params(nested_key.to_s)
```
paulcsmith: I believe there is a bug here. With this logic if there are any query params it will check those and never check form params. So it would fail if you submitted a form that had a query param.

I think what the logic should be something like this:

* If there are form params and they have a matching nested key, -> use those
* If no forms params or there are form params, but not with a matching key -> check the query params for a nested key
* If no query params have the nested key -> return empty hash

Does that make sense?

spec/lucky/params_spec.cr
```diff
@@ -156,6 +155,13 @@ describe Lucky::Params do
       params.nested(:user).should eq({"name" => "Paul", "age" => "28"})
     end
 
+    it "gets query nested params" do
```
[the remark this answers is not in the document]
jwoertink: Yup. Want me to update [this](https://github.com/luckyframework/lucky/blob/master/spec/lucky/params_spec.cr#L136) and [this](https://github.com/luckyframework/lucky/blob/master/spec/lucky/params_spec.cr#L146) to match?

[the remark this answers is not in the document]
paulcsmith: That'd be great!

src/lucky/params.cr
jwoertink: Ah, yeah, you're right. I think I actually ran in to an issue with this because another spec was failing. I'll dig back in and switch up the logic to fit this and see how that goes.

jwoertink:
@paulcsmith I synced with master to make sure this branch is running on crystal 0.25.0, and then made that change. I know the code isn't the prettiest, but specs pass, and it all seems ok. I'd rather be correct than "clever" lol, but if you have a better way to write that, let me know. Thanks!

jwoertink:
I like that much better. Ok, so that update is made, and master is synced up.

jwoertink:
boom. Knocking out code.

spec/lucky/params_spec.cr
```diff
@@ -156,6 +165,20 @@ describe Lucky::Params do
       params.nested(:user).should eq({"name" => "Paul", "age" => "28"})
     end
 
+    it "gets nested query params" do
+      request = build_request body: "", content_type: ""
```
paulcsmith: Maybe do what you did with the JSON spec and add some form params here to make sure thy are merged?

jwoertink: hold on... laughing to hard at that gif... 😹

PR #496, opened 2018-06-07
Add Redirectable docs

edwardloveall:
To the user, this will likely feel like the same method, so that's the way I documented it. Instead of documenting each method, I just put some module docs up top to explain how all the different type
signatures work.

Let me know if you'd like me to document the individual methods more. I thought about marking them as `:nodoc:` but there seemed to be _some_ value in showing the type signatures.

### DIFF
--- a/src/lucky/redirectable.cr
+++ b/src/lucky/redirectable.cr
@@ -1,9 +1,10 @@
-# Redirect the browser
+# Redirect the request
 #
 # There are multiple ways to redirect inside of an action. The most common ways are to redirect to a `Lucky::Action` class, or a URL/path `String`. Both use the `redirect` method:
 #
 # ```crystal
 # redirect to: Users::Index
+# redirect to: Users::Show.with(user.id)
 # redirect to: "https://luckyframework.org/"
 # redirect to: "/users"
 # ```
@@ -11,26 +12,44 @@
 # By default, the method will set the status code to `302` A.K.A. "Found". If you want to customize the status code, you can pass it directly:
 #
 # ```crystal
-# # create a user
-# redirect to: Users::Index, status: 201
+# redirect to: Users::Index, status: 301
 #
 # # or use the built in enum value
-# redirect to: Users::Index, status: Status::Created
+# redirect to: Users::Index, status: Status::MovedPermanently
 # ```
 #
-# Internally, all the different methods in this module eventually use the method that takes a `String`. However, it's recommended you pass a `Lucky::Action` class if possible because it guarentees runtime safety.
+# You can find a list of all of the possible statuses [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/action.cr).
+#
+# Internally, all the different methods in this module eventually use the
+# method that takes a `String`. However, it's recommended you pass a
+# `Lucky::Action` class if possible because it guarantees runtime safety.
 module Lucky::Redirectable
   # Redirect using a `Lucky::RouteHelper`
+  #
+  # ```crystal
+  # route = Lucky::RouteHelper.new(method: :get, path: "/users/1")
+  # redirect to: route, status: 301
+  # ```
   def redirect(to route : Lucky::RouteHelper, status = 302)
     redirect to: route.path, status: status
   end
 
-  # Redirect using a `Lucky::Action`
+  # Redirect to a `Lucky::Action`
+  #
+  # ```crystal
+  # redirect to: Users::Index
+  # redirect to: Users::Show.with(user.id), status: 301
+  # ```
   def redirect(to action : Lucky::Action.class, status = 302)
     redirect to: action.route, status: status
   end
 
   # Redirect using a `String`
+  #
+  # ```crystal
+  # redirect to: "/users"
+  # redirect to: "/users/1", status: 301
+  # ```
   def redirect(to path : String, status = 302)
     context.response.headers.add "Location", path
     context.response.headers.add "Turbolinks-Location", path
@@ -39,6 +58,10 @@ module Lucky::Redirectable
   end
 
   # Redirect using a `String` and a `Status` value
+  #
+  # ```crystal
+  # redirect to: "/users", status: Status::MovedPermanently
+  # ```
   def redirect(to path : String, status : Lucky::Action::Status = Lucky::Action::Status::Found)
     redirect(path, status.value)
   end
--- a/src/lucky/redirectable.cr
+++ b/src/lucky/redirectable.cr
@@ -27,8 +27,7 @@ module Lucky::Redirectable
   # Redirect using a `Lucky::RouteHelper`
   #
   # ```crystal
-  # route = Lucky::RouteHelper.new(method: :get, path: "/users/1")
-  # redirect to: route, status: 301
+  # redirect to: Users::Show.with(user.id), status: 301
   # ```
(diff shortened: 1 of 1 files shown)


src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
...
+# ```
+#
+# By default, the method will set the status code to `302` A.K.A. "Found". If you want to customize the status code, you can pass it directly:
+#
+# ```crystal
+# # create a user
+# redirect to: Users::Index, status: 201
+#
+# # or use the built in enum value
+# redirect to: Users::Index, status: Status::Created
```
edwardloveall: I would put all the `Status` enum values here, but I think linking to them once they are documented would be better.

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
+# Redirect the browser
+#
+# There are multiple ways to redirect inside of an action. The most common ways are to redirect to a `Lucky::Action` class, or a URL/path `String`. Both use the `redirect` method:
+#
+# ```crystal
+# redirect to: Users::Index
+# redirect to: "https://luckyframework.org/"
+# redirect to: "/users"
```
paulcsmith: I'm thinking of removing the `to` and recommending people pass just the string or action: `redirect Users::Index` thoughts?

src/lucky/redirectable.cr
paulcsmith: I think it would help to use 3xx codes here so it is more realistic http://www.restapitutorial.com/httpstatuscodes.html 

Some example enums would be these: https://github.com/luckyframework/lucky/blob/3d15a40d47b8aa16b806c92c3b698ab54af642a0/src/lucky/action.cr#L35-L40

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
...
+#
+# ```crystal
+# # create a user
+# redirect to: Users::Index, status: 201
+#
+# # or use the built in enum value
+# redirect to: Users::Index, status: Status::Created
+# ```
+#
+# Internally, all the different methods in this module eventually use the method that takes a `String`. However, it's recommended you pass a `Lucky::Action` class if possible because it guarentees runtime safety.
```
paulcsmith: guarentees -> guarantees

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
+# Redirect the browser
+#
+# There are multiple ways to redirect inside of an action. The most common ways are to redirect to a `Lucky::Action` class, or a URL/path `String`. Both use the `redirect` method:
+#
+# ```crystal
+# redirect to: Users::Index
```
paulcsmith: Almost might be worth showing an example that takes params:

```
redirect Users::Show.with(user.id)
```

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
...
+# ```
+#
+# Internally, all the different methods in this module eventually use the method that takes a `String`. However, it's recommended you pass a `Lucky::Action` class if possible because it guarentees runtime safety.
 module Lucky::Redirectable
+  # Redirect using a `Lucky::RouteHelper`
   def redirect(to route : Lucky::RouteHelper, status = 302)
     redirect to: route.path, status: status
   end
 
+  # Redirect using a `Lucky::Action`
```
paulcsmith: Might be helpful to give a small example because people may not know the difference between a RouteHelper and an Action:

```
redirect Users::Index
```

And in the other one: `redirect Users::Show.with(user.id)`

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,44 @@
...
 
+  # Redirect using a `String`
   def redirect(to path : String, status = 302)
     context.response.headers.add "Location", path
     context.response.headers.add "Turbolinks-Location", path
     context.response.status_code = status
     Lucky::Response.new(context, "", "")
   end
 
+  # Redirect using a `String` and a `Status` value
```
paulcsmith: Maybe a link to the file with a list of statuses? Not ideal, but is probably helpful until docs are published

src/lucky/redirectable.cr
edwardloveall: I kind of like the `to` to be honest, but I can't say I have a strong argument _for_ it. I've never removed explicitness and been sad later. It's also pretty short, and I like when methods can read more like english.

Also, you don't _need_ to use it because of how crystal handles arguments:

```cr
require "./src/lucky"
require "./spec/spec_helper"
include ContextHelper

class Users::Index < Lucky::Action
  get "/users" do
    text "you won't see me"
  end
end

index = Users::Index.new(build_context, params)
route = Lucky::RouteHelper.new(method: :get, path: "/users")

index.redirect(route).context.response.headers
index.redirect(to: route).context.response.headers
```

edwardloveall:
Thanks, great feedback. Ready for more.

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,67 @@
...
+# By default, the method will set the status code to `302` A.K.A. "Found". If you want to customize the status code, you can pass it directly:
+#
+# ```crystal
+# redirect to: Users::Index, status: 301
+#
+# # or use the built in enum value
+# redirect to: Users::Index, status: Status::MovedPermanently
+# ```
+#
+# You can find a list of all of the possible statuses [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/action.cr).
```
edwardloveall: Not the best link, but it seemed better than linking to a soon-to-be out of date commit or a line number that might change.

paulcsmith: 👍 This should work

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,68 @@
...
+#
+# Internally, all the different methods in this module eventually use the
+# method that takes a `String`. However, it's recommended you pass a
+# `Lucky::Action` class if possible because it guarantees runtime safety.
 module Lucky::Redirectable
+  # Redirect using a `Lucky::RouteHelper`
+  #
+  # ```crystal
+  # route = Lucky::RouteHelper.new(method: :get, path: "/users/1")
```
paulcsmith: I think it's pretty rare to instantiate a RouteHelper manually. A RouteHelper is used when you use `MyAction.with(..)`

So this example would be something like `redirect to: Users::Show.with(id)`

src/lucky/redirectable.cr
```diff
@@ -1,19 +1,68 @@
...
+  # ```
   def redirect(to route : Lucky::RouteHelper, status = 302)
     redirect to: route.path, status: status
   end
 
+  # Redirect to a `Lucky::Action`
+  #
+  # ```crystal
+  # redirect to: Users::Index
+  # redirect to: Users::Show.with(user.id), status: 301
```
paulcsmith: This line wouldn't work for this because it would generate a RouteHelper

edwardloveall: Oh I see, yeah. I guess it's not possible to pass parameters with an action class. It would need a third argument.

src/lucky/redirectable.cr
edwardloveall: This is why I'm doing documentation 😄

paulcsmith:
Looks great!!

PR #515, opened 2018-06-18
Docs for hash extensions

edwardloveall:
I'm starting to go through and document things that haven't been touched in a long time according to git. It seems like they would be the most stable and therefore safest to document.

### DIFF
--- a/src/charms/hash_extensions.cr
+++ b/src/charms/hash_extensions.cr
@@ -1,8 +1,30 @@
 class Hash
+  # Return the **nillable** value of a hash key
+  #
+  # This returns the value stored in the hash. Useful for getting values out of
+  # the parameter hash. Returns `nil` if the value doesn't exist:
+  #
+  # ```crystal
+  # action do
+  #   params.get(:name) # => "Karin" : (String | Nil)
+  #   params.get(:asdf) # => nil : (String | Nil)
+  # end
+  # ```
   def get(key : String | Symbol)
     self[key.to_s]?
   end
 
+  # Return the value of a hash key
+  #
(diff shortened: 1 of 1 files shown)


src/charms/hash_extensions.cr
```diff
@@ -1,8 +1,30 @@
 class Hash
+  # Return the **nillable** value of a hash key
+  #
+  # This returns the value stored in the hash. Useful for getting values out of
+  # the parameter hash. Returns `nil` if the value doesn't exist:
+  #
+  # ```crystal
+  # action do
+  #   params.get(:name) # => "Karin" : (String | Nil)
+  #   params.get(:asdf) # => nil : (String | Nil)
```
paulcsmith: Though this example does work, the `params` method returns a `Lucky::Params` object (which also has a `get` and `get?`) method

It may be better to instantiate a new hash and use `get` on it:

```crystal
my_hash = {"name" => "Jane")
my_hash.get("name")
my_hash.get(:name)
```

It should also be worth mentioning that it converts symbols to strings. That's one of the main reasons this was added. It's like HashWithIndifferentAccess in Rails. Makes it easier to access hashes with either strings or symbols

edwardloveall: Ah yup, I assume they were the same. I'll put the type conversion note in there as well. Thanks!

edwardloveall:
Alright, tell me what ya think of that. I referred to the code example because I think it's more clear than writing it out in prose, but tell me if you disagree 😄 

Also, when you merge can you squash these down or let me do it? The intermediary commits don't need to be there 🙂

PR #524, opened 2018-06-24
Allow setting nil as the default value, closes #448

### LINKED ISSUE #448: When setting up nilable query params, allow setting `nil` as default
Right now it blows up. 

Another alternative is to not allow setting nil and let the developer know they can remove the default

hanneskaeufler:
I think this should be okay.
Previously, this would result in a compile time error.


paulcsmith:
And congrats on the win on Saturday :)

PR #541, opened 2018-07-14
Add custom config to log handler output. fixes #480

### LINKED ISSUE #480: Allow customizing the log message
https://github.com/luckyframework/lucky/issues/440#issuecomment-392229762

Something like this:

```
Lucky::LogHandler.configure do
  settings.formatter = ->(context: HTTP::Context) do
    io << context.method # etc.
  end
end
```

Here is where logs are currently printed: https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/src/lucky/log_handler.cr#L30

Here's where a configuration option can be added to override the way logs are printed: https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/src/lucky/log_handler.cr#L7

jwoertink:
This first commit actually tanks crystal:

```
[15:04PM] lucky (features/custom_log_message)$ crystal spec/lucky/log_handler_spec.cr
Module validation failed: !dbg attachment points at wrong subprogram for function
!12 = distinct !DISubprogram(name: "request_formatter?", linkageName: "request_formatter?", scope: !5, file: !5, line: 7, type: !6, isLocal: true, isDefinition: true, scopeLine: 7, isOptimized: false, unit: !0, variables: !2)
%"->" ()* @"*Lucky::LogHandler::Settings::request_formatter?:(Proc(Lucky::LogHandler, HTTP::Server::Context, Time, Time::Span, String) | Nil)"
  %0 = call %"->"* @"~Lucky::LogHandler::Settings::request_formatter:read"(), !dbg !13
!13 = !DILocation(line: 7, column: 3, scope: !14)
!14 = distinct !DISubprogram(name: "show_timestamps", linkageName: "show_timestamps", scope: !5, file: !5, line: 7, type: !6, isLocal: true, isDefinition: true, scopeLine: 7, isOptimized: false, unit: !0, variables: !2)
!14 = distinct !DISubprogram(name: "show_timestamps", linkageName: "show_timestamps", scope: !5, file: !5, line: 7, type: !6, isLocal: true, isDefinition: true, scopeLine: 7, isOptimized: false, unit: !0, variables: !2)
 (Exception)
  from Crystal::CodeGenVisitor#finish:Nil
  from Crystal::Compiler#codegen<Crystal::Program, Crystal::ASTNode+, Array(Crystal::Compiler::Source), String>:(Tuple(Array(Crystal::Compiler::CompilationUnit), Array(String)) | Nil)
  from Crystal::Compiler#compile<Array(Crystal::Compiler::Source), String>:Crystal::Compiler::Result
  from Crystal::Command#run_command<Bool>:Nil
  from Crystal::Command#run:(Bool | Crystal::Compiler::Result | Nil)
  from main
Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
```

Until that is solved, let me know your thoughts on implementation. To keep the current format the same, we would need to pass the log handler instance in, and make those helper methods public. Maybe those methods need to be moved out to a module? 

EDIT: running `crystal spec/ --no-debug` all specs pass.

### DIFF
--- a/spec/lucky/log_handler_spec.cr
+++ b/spec/lucky/log_handler_spec.cr
@@ -54,11 +54,7 @@ describe Lucky::LogHandler do
 
   context "when configured to log timestamps" do
     it "logs timestamp" do
-      begin
-        Lucky::LogHandler.configure do
-          settings.show_timestamps = true
-        end
-
+      Lucky::LogHandler.temp_config(show_timestamps: true) do
         io = IO::Memory.new
         called = false
         log_io = IO::Memory.new
@@ -70,21 +66,13 @@ describe Lucky::LogHandler do
         log_output.should contain("GET")
         log_output.should match(/#{Time.now.to_s("%Y-%m-%d")}/)
         called.should be_true
-      ensure
-        Lucky::LogHandler.configure do
-          settings.show_timestamps = false
-        end
       end
     end
   end
 
   context "when configured to be disabled" do
     it "logs timestamp" do
-      begin
-        Lucky::LogHandler.configure do
-          settings.enabled = false
-        end
-
+      Lucky::LogHandler.temp_config(enabled: false) do
         io = IO::Memory.new
         called = false
         log_io = IO::Memory.new
@@ -95,10 +83,6 @@ describe Lucky::LogHandler do
         log_output = log_io.to_s
         log_output.should eq ""
         called.should be_true
-      ensure
-        Lucky::LogHandler.configure do
-          settings.enabled = true
-        end
       end
     end
   end
@@ -121,11 +105,7 @@ describe Lucky::LogHandler do
 
   context "when configured with custom log formatter" do
     it "logs emoji" do
-      begin
-        Lucky::LogHandler.configure do
-          settings.log_formatter = EmojiLogFormatter.new
-        end
-
+      Lucky::LogHandler.temp_config(log_formatter: EmojiLogFormatter.new) do
         called = false
         log_io = IO::Memory.new
         context = build_context("PATCH")
@@ -135,10 +115,6 @@ describe Lucky::LogHandler do
         log_output = log_io.to_s.chomp
         log_output.should eq "🐵 PATCH - BOOM"
         called.should be_true
-      ensure
-        Lucky::LogHandler.configure do
-          settings.log_formatter = DefaultLogFormatter.new
-        end
       end
     end
   end
--- a/spec/lucky/log_handler_spec.cr
+++ b/spec/lucky/log_handler_spec.cr
@@ -3,6 +3,12 @@ require "http/server"
 
 include ContextHelper
 
+class EmojiLogFormatter < Lucky::LogFormatters::Base
+  def format(context, time, elapsed) : String
+    "🐵 #{context.request.method} - BOOM"
+  end 
+end
+
 describe Lucky::LogHandler do
   it "logs" do
     io = IO::Memory.new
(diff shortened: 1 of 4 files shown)


src/lucky/log_handler.cr
```diff
@@ -2,9 +2,13 @@ require "colorize"
 
 class Lucky::LogHandler
   include HTTP::Handler
+  alias LogProc = ::Proc(Lucky::LogHandler, HTTP::Server::Context, Time, Time::Span, String)
 
   Habitat.create do
     setting show_timestamps : Bool
+    setting request_formatter : LogProc = ->(handler : Lucky::LogHandler, context : HTTP::Server::Context, time : Time, elapsed : Time::Span) {
+      "#{context.request.method} #{handler.colored_status_code(context.response.status_code)} #{context.request.resource}#{handler.timestamp(time)} (#{handler.elapsed_text(elapsed)})"
+    }
```
paulcsmith: How about instead of using a Proc, there would be a class in here called `LogFormatter`

Something like this:

```
abstract class LogFormatter
  abstract def format(context, time, elapsed) : String

  # Include elapsed_text, colored_status_code, etc.
end
```

Then the default would be:

```
class DefaultLogFormatter < LogFormatter
  def log_request(context, time, elapsed)
    "#{context.request.method} #{handler.colored_status_code(context.response.status_code)} #{context.request.resource}#{handler.timestamp(time)} (#{handler.elapsed_text(elapsed)})"
  end
end

setting log_formatter : LogFormatter = DefaultLogFormatter.new
```

Does that make sense?

jwoertink: ah. nice. That might be nicer. I'll try that out.

jwoertink:
Ok, this is updated now with the new structure. I like it better. There is one method still sitting as a small duplication which is the `timestamp` method in the handler, and the formatter. This is because the log_exception still uses it. 

One option could be to have a second `format` method in the log formatter that could take the exception as an argument then update the log based on that. Another option could be a separate config? (but that would start making things really confusing). If you have some ideas, or if you're cool with how it is, let me know.

spec/lucky/log_handler_spec.cr
```diff
@@ -112,6 +118,30 @@ describe Lucky::LogHandler do
...
+        end
+
+        called = false
+        log_io = IO::Memory.new
+        context = build_context("PATCH")
+
+        call_log_handler_with(log_io, context) { called = true }
+
+        log_output = log_io.to_s.chomp
+        log_output.should eq "🐵 PATCH - BOOM"
```
paulcsmith: I think EmojiFormatter should be the default... :P

src/lucky/log_formatters/default_log_formatter.cr
```diff
@@ -0,0 +1,6 @@
+class DefaultLogFormatter < Lucky::LogFormatters::Base
+
```
paulcsmith: Could you run `crystal tool format src spec` to remove this and make sure the other stuff is formatted ok?

src/lucky/log_formatters/base.cr
```diff
@@ -0,0 +1,37 @@
...
+    when 400..499
+      "#{status_code.colorize(:yellow)}"
+    when 500..599
+      "#{status_code.colorize(:red)}"
+    else
+      "#{status_code}"
+    end
+  end
+  
+  private def timestamp(time)
```
paulcsmith: Maybe this could be a public class method so it can be used by the handler as well?

spec/lucky/log_handler_spec.cr
jwoertink: lol

PR #543, opened 2018-07-19
Add enabled setting to LogHandler

mikeeus:
For https://github.com/luckyframework/lucky_cli/issues/230

Adds `enabled` setting to `LogHandler` and returns early before logging if disabled.


mikeeus:
@paulcsmith Sure thing, I just updated it

PR #546, opened 2018-07-20
adding in helper to catch plain text content types

jwoertink:
For being able to do 

```crystal
if plain?
  render_some_text
else
  render(SomeHTMLPage)
end
```

EDIT: per [this](https://www.w3.org/Protocols/rfc1341/7_1_Text.html) adding in the charset as an optional parameter, I updated to allow `text/plain; charset=WHATEVER`

### DIFF
--- a/spec/lucky/content_type_helper_spec.cr
+++ b/spec/lucky/content_type_helper_spec.cr
@@ -64,10 +64,6 @@ describe Lucky::ContentTypeHelpers do
     set_content_type "text/plain; charset=UTF8" do
       FakeAction.new.plain?.should be_true
     end
-
-    set_content_type "plain/text" do
-      FakeAction.new.plain?.should be_false
-    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


spec/lucky/content_type_helper_spec.cr
```diff
@@ -55,6 +55,20 @@ describe Lucky::ContentTypeHelpers do
...
+  it "works for plain text" do
+    set_content_type "text/plain" do
+      FakeAction.new.plain?.should be_true
+    end
+
+    set_content_type "text/plain; charset=UTF8" do
+      FakeAction.new.plain?.should be_true
+    end
+
+    set_content_type "plain/text" do
```
paulcsmith: Is this in the spec or used in the wild? I haven’t seen it before but if it’s a common thing then I think it makes sense to add. Just want to make sure

jwoertink: No, I often catch myself typing that wrong, so I figured I'd use it. I guess it's sort of irrelevant because it could really be replaced with anything. 🤷‍♂️

paulcsmith: In that case, I think removing it would be best since it isn't technically correct

paulcsmith:
Just one comment. Looks good otherwise!

jwoertink:
Thanks! That's done.

PR #549, opened 2018-07-23
Ability to pass in "valueless" attributes. Fixes #518

### LINKED ISSUE #518: Allow passing valueless attributes to HTML methods
```crystal
video attrs: [:allow_fullscreen], class: "my-video"
text_input f.name, attrs: [:required, :autofocus], class: "input"
```

jwoertink:
Initial work on adding in [boolean attributes](http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes) (aka valueless attributes).

This initial push has a failing spec, and while I'm working on that, I just commented out everything else in that file. This will eventually work like: 
```crystal
text_input f.name, attrs: [:required, :autofocus], class: "input"
```
and then output

```html
<input type="text" name="whatever" class="input" required autofocus />
```

Resolves #518


src/lucky/tags/base_tags.cr
```diff
@@ -75,14 +75,22 @@ module Lucky::BaseTags
...
+
   private def merge_options(html_options, tag_attrs)
     options = {} of String => String
     tag_attrs.each do |key, value|
       options[key.to_s] = value
     end
 
     html_options.each do |key, value|
-      options[key.to_s] = value
+      options[key.to_s] = value.as(String)
```
jwoertink: I had to put this here (for now at least) because I'm getting 
> no overload matches 'Hash(String, String)#[]=' with types String, (Array(Symbol) | String)

At some point, the compiler thinks `value` could have `Array(Symbol)`. This is from passing the `attrs: [:required]` in to the `html_options`. Though, that array never actually gets in here

src/lucky/tags/base_tags.cr
```diff
@@ -75,14 +75,23 @@ module Lucky::BaseTags
...
+
   private def merge_options(html_options, tag_attrs)
     options = {} of String => String
     tag_attrs.each do |key, value|
       options[key.to_s] = value
     end
 
     html_options.each do |key, value|
-      options[key.to_s] = value
+      next if key == :boolean_attrs
```
jwoertink: I'd rather this key not even make it in to here, but then we would need to constantly convert NamedTuple in to hashes to reject that key while still passing the boolean_attrs around.

paulcsmith:
@jwoertink I'd create another branch and link to the code so that they have all the info they need. I'm not sure what the issue is so might be better to hold off on this, but that is up to you!

jwoertink:
@paulcsmith Ok, so I gave this another shot in a different way. Spec passes, and crystal doesn't crash. However, I feel like there's some extra duplication going on here. Maybe that's ok? Let me know your thoughts on this, and I'll start adding this in to all the other tags as well.

src/lucky/tags/base_tags.cr
```diff
@@ -30,7 +30,15 @@ module Lucky::BaseTags
       end
     end
 
-    def {{method_name.id}}(boolean_attrs : String = "", options = EMPTY_HTML_ATTRS, **other_options, &block)
+    def {{method_name.id}}(options = EMPTY_HTML_ATTRS, **other_options, &block)
```
jwoertink: How do I just make this one call the next override method? 
```crystal
{{method_name.id}}("", options = EMPTY_HTML_ATTRS, **other_options, &block) 
#=> src/lucky/tags/form_helpers.cr:11: can't use `yield` inside a proc literal or captured block
```
Adding a default arg to the second one causes any tag that uses a NamedTuple literal to fail like 
```
div({class: "test"}) #=> no overload matches 'TestRender#header' with type NamedTuple(class: String)

**Comment by paulcsmith** [comment]
Just a heads up that I saw this, but need to spend some time poking around with it. The tag/input code is a bit tricky so I just want to double check before merging in

**Comment by jwoertink** [comment]
@paulcsmith I merged in the latest master updates since there were some conflicts. I resolved those, and made sure specs pass. The CI error is related to the latest crystal stuff it looks like.

**Comment by jwoertink** [comment]
Sweet! Checks passing now. Merged latest master and checked locally on 0.26.1

File: src/lucky/tags/input_helpers.cr [hunk 4]
```diff
@@ -59,56 +59,26 @@ module Lucky::InputHelpers
...
-  end
-
-  generate_helpful_error_for color_input
-
-  def color_input(field : LuckyRecord::FillableField, **html_options)
-    generate_input(field, "color", html_options)
-  end
-
-  generate_helpful_error_for hidden_input
+  {% for f in ["text", "email", "file", "color", "hidden", "number", "url", "search", "range"] %}
```
paulcsmith: Very nice :) What do you think about changing `f` to `input_type`?

paulcsmith:
And actually I'll merge it and I can make that change since it is so small!

PR #553, opened 2018-07-25
Add match macro for routes with custom HTTP methods

### LINKED ISSUE #551: Add 'match' macro to actions
> Would it make sense to include a match method that takes any HTTP verb as a symbol? I’m likely to want to match PROPFIND, PROPPATCH, REPORT, etc. too.
> https://gitter.im/luckyframework/Lobby?at=5b575246f477e4664abc5274

It would look like this:

```crystal
class MySpecialAction < Api/BrowserAction
  match :report, "my-path" do
  end
end
```

Ideally, it should raise at compile-time if given an uppercase symbol. This will help maintain consistency.

The code should go in the `Routeable` module: https://github.com/luckyframework/lucky/blob/master/src/lucky/routeable.cr. This bit of code will be helpful: https://github.com/luckyframework/lucky/blob/master/src/lucky/routeable.cr#L21-L29

And here are the specs: https://github.com/luckyframework/lucky/blob/master/spec/lucky/action_spec.cr#L134-L175

joeldrapper:
Use this if you need to match a route with a custom HTTP method (verb). For example:

```crystal
class Profile::Show
  match :options, "/profile" do
    # action code here
  end
end
```

Will respond to an `HTTP OPTIONS` request.

Closes #551

### DIFF
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -19,15 +19,38 @@ module Lucky::Routeable
     # **See also** our guides for more information and examples:
     # * [Routing](https://luckyframework.org/guides/actions-and-routing/#routing)
     macro {{ http_method.id }}(path)
-      \{% unless path.starts_with?("/") %}
-        \{% path.raise "Path must start with a slash. Example: '/" + path + "'" %}
-      \{% end %}
-
-      add_route :{{ http_method.id }}, \{{ path }}, \{{ @type.name.id }}
-
-      setup_call_method(\{{ yield }})
+      match(:{{ http_method.id }}, \{{ path }}) do
+        \{{ yield }}
+      end
     end
   {% end %}
+
+  # Define a route with a custom HTTP method.
+  #
+  # Use this method if you need to match a route with a custom HTTP method (verb).
+  # For example:
+  #
+  # ```
+  # class Profile::Show
+  #   match :options, "/profile" do
+  #     # action code here
+  #   end
+  # end
+  #
+  # Will respond to an `HTTP OPTIONS` request.
+  macro match(method, path)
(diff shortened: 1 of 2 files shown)


src/lucky/routeable.cr
```diff
@@ -19,16 +19,39 @@ module Lucky::Routeable
...
     macro {{ http_method.id }}(path)
-      \{% unless path.starts_with?("/") %}
-        \{% path.raise "Path must start with a slash. Example: '/" + path + "'" %}
-      \{% end %}
-
-      add_route :{{ http_method.id }}, \{{ path }}, \{{ @type.name.id }}
-
-      setup_call_method(\{{ yield }})
+      match(:{{ http_method.id }}, \{{ path }}) do
+        \{{ yield }}
```
joeldrapper: This is a bit strange but I can’t seem to figure out how to forward blocks from macros. I’m quite new to Crystal.

src/lucky/routeable.cr
```diff
@@ -19,16 +19,39 @@ module Lucky::Routeable
...
+  #
+  # ```
+  # class Profile::Show
+  #   match :options, "/profile" do
+  #     # action code here
+  #   end
+  # end
+  #
+  # Will respond to an `HTTP OPTIONS` request.
+  macro match(method, path)
```
joeldrapper: 🤔 I wonder if this should actually take a splat of methods here so it can match on more than one.

src/lucky/routeable.cr
paulcsmith: Macros are tricky. This is how I've done it in the past and I'm not aware of something better. So I think this is good!

PR #558, opened 2018-07-27
check to see if the browsersync port is currently in use

### LINKED ISSUE #261: If port 3001 is taken, raise a helpful error
Browsersync will just find the next port if 3001 is taken. This breaks reloading though so instead we should check the port and if it is taken, raise an error so the user knows about it. There should also be a way to specific a different Browsersync port. This can probably be done by adding a setting using Habitat

jwoertink:
Closes #261 

This checks to see if port 3001 is currently in use before trying to boot browsersync. If it is, then we raise an exception and show a pretty message to the user. 

It may not be needed on reload, so I could probably just drop it down to only checking around start. Also, I wasn't sure how to actually test this. Let me know if you see anything.

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -10,6 +10,8 @@ module Sentry
   FILE_TIMESTAMPS  = {} of String => String # {file => timestamp}
   BROWSERSYNC_PORT = 3001
 
+  class BrowserSyncPortInUseError < Exception; end
+
   class ProcessRunner
     include LuckyCli::TextHelpers
 
@@ -41,15 +43,28 @@ module Sentry
 
       self.successful_compilations += 1
       if reload_browser?
-        reload_or_start_browser_sync
+        begin
+          reload_or_start_browser_sync
+        rescue Sentry::BrowserSyncPortInUseError
+          puts "Another process is currently using the BrowserSync port #{BROWSERSYNC_PORT}".colorize(:red)
+        end
       end
     end
 
     private def reload_or_start_browser_sync
       if successful_compilations == 1
+        check_browsersync_port_availability
         start_browsersync
       else
         reload_browsersync
+      end
+    end
+
+    private def check_browsersync_port_availability
+      if File.executable?(`which lsof`.chomp)
+        io = IO::Memory.new
+        Process.run("lsof -i :#{BROWSERSYNC_PORT}", output: io, error: STDERR, shell: true)
+        raise Sentry::BrowserSyncPortInUseError unless io.to_s.empty?
       end
     end
 


tasks/watch.cr
```diff
@@ -41,18 +43,31 @@ module Sentry
...
       if successful_compilations == 1
+        check_browsersync_port_availability
         start_browsersync
       else
         reload_browsersync
       end
     end
 
+    private def check_browsersync_port_availability
+      if File.executable?(`which lsof`.chomp)
```
paulcsmith: Nice 👍

tasks/watch.cr
```diff
@@ -41,18 +43,31 @@ module Sentry
...
+        io = IO::Memory.new
+        Process.run("lsof -i :#{BROWSERSYNC_PORT}", output: io, error: STDERR, shell: true)
```
paulcsmith: Can this run without `shell` or is it required?

tasks/watch.cr
```diff
@@ -41,18 +43,31 @@ module Sentry
...
+        raise Sentry::BrowserSyncPortInUseError.new unless io.to_s.empty?
```
paulcsmith: Rather than raising an error and rescuing it, what do you think about renaming this method to `browsersync_port_is_available?` and return a Bool. Then create the processes or print the error. I think that's faster and a bit more straightforward than raising and rescuing an error

tasks/watch.cr
```diff
@@ -41,18 +43,31 @@ module Sentry
 
       self.successful_compilations += 1
       if reload_browser?
-        reload_or_start_browser_sync
+        begin
+          reload_or_start_browser_sync
+        rescue Sentry::BrowserSyncPortInUseError
+          puts "Another process is currently using the BrowserSync port #{BROWSERSYNC_PORT}".colorize(:red)
```
paulcsmith: Any ideas on how we could make it easier for someone to figure out what process is using it or what to do? 

I'm also ok with this since it's better than not letting the user now, it'd just be super cool if there was a simple command they could run to find/stop processes running on a certain port. But I'm not aware of such a command

tasks/watch.cr
jwoertink: it's required as far as I can tell. I think I would need to know the location of the exec without the shell. Does this add a performance hit?

tasks/watch.cr
jwoertink: @paulcsmith What are your thoughts on this output for the pretty error message? 
screen shot 2018-08-01 at 2 47 57 pm

tasks/watch.cr
jwoertink: Updated.

PR #560, opened 2018-08-07
Add gen.resource.browser

### LINKED ISSUE #539: Allow setting editor for `lucky exec` using a yml file

paulcsmith:
Closes #539

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ install:
   - shards install
 script:
   - crystal spec
-  # Make sure everything works even without LuckyMigrator
-  - crystal spec -D without-migrator
   - crystal tool format spec src --check
   - shards build
   - bin/ameba
--- a/spec/tasks/gen/resource_spec.cr
+++ b/spec/tasks/gen/resource_spec.cr
@@ -1,0 +1,78 @@
+require "../../spec_helper"
+
+include CleanupHelper
+include GeneratorHelper
+
+describe Gen::Action do
+  it "generates actions, model, form and query" do
+    with_cleanup do
+      io = generate Gen::Resource::Browser, "User", "name:String"
+
+      should_create_files_with_contents io,
+        "./src/actions/users/index.cr": "class Users::Index < BrowserAction",
+        "./src/actions/users/show.cr": "class Users::Show < BrowserAction",
+        "./src/actions/users/new.cr": "class Users::New < BrowserAction",
+        "./src/actions/users/create.cr": "class Users::Create < BrowserAction",
+        "./src/actions/users/edit.cr": "class Users::Edit < BrowserAction",
+        "./src/actions/users/update.cr": "class Users::Update < BrowserAction",
+        "./src/actions/users/delete.cr": "class Users::Delete < BrowserAction"
+      should_create_files_with_contents io,
+        "./src/pages/users/index_page.cr": "class Users::IndexPage < MainLayout",
+        "./src/pages/users/show_page.cr": "class Users::ShowPage < MainLayout",
+        "./src/pages/users/new_page.cr": "class Users::NewPage < MainLayout",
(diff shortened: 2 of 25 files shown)


tasks/gen/resource/browser.cr
```diff
@@ -0,0 +1,168 @@
...
+  private def columns : Array(Lucky::GeneratedColumn)
+    column_definitions.not_nil!.map do |column_definition|
+      column_name, column_type = column_definition.split(":")
+      Lucky::GeneratedColumn.new(name: column_name, type: column_type)
+    end
+  end
+
+  private def generate_resource
+    Lucky::ResourceTemplate.new(resource_name, columns).render("./src/")
+    LuckyMigrator::MigrationGenerator.new("Create" + pluralized_resource).generate
```
paulcsmith: Need to make the migration generator also accept a list of columns

spec/tasks/gen/resource_spec.cr
```diff
@@ -0,0 +1,68 @@
...
+  it "displays an error if given a pluralized resource" do
+    io = generate Gen::Resource::Browser, "Users"
+
+    io.to_s.should contain("Resource must be singular")
+  end
+
+  it "raises if column options are missing" do
+    io = generate Gen::Resource::Browser, "User"
+
+    io.to_s.should contain("mMust provide valid columns for the resource")
```
hanneskaeufler: Is the `mMust` intentional?

spec/tasks/gen/resource_spec.cr
```diff
@@ -0,0 +1,68 @@
...
+
+    io.to_s.should contain("Namespaced resources are not supported")
+  end
+end
+
+private def generate(generator : Class, *options)
+  options.each { |option| ARGV.push(option) }
+  io = IO::Memory.new
+  generator.new.call(io)
+  io
```
hanneskaeufler: What do you think about `IO::Memory.new.tap { |io| generator.new.call(io) }` as a oneliner? *Desperately trying to use `tap` somewhere*

paulcsmith: I like it! 👍

spec/tasks/gen/resource_spec.cr
paulcsmith: Oops :)

.travis.yml
```diff
@@ -3,8 +3,6 @@ install:
   - shards install
 script:
   - crystal spec
-  # Make sure everything works even without LuckyMigrator
-  - crystal spec -D without-migrator
```
paulcsmith: No point in having LuckyRecord and not LuckyMigrator. So I'm adding LuckyMigrator so specs and code are a bit simpler. At some later point Migrator and Record will be merged since they are also so intertwined. It's a pain to open separate PRs to make updates

spec/tasks/gen/resource_spec.cr
```diff
@@ -0,0 +1,78 @@
...
+  it "raises if column options are not formatted right" do
+    with_cleanup do
+      io = generate Gen::Resource::Browser, "User", "name", "String"
+
+      io.to_s.should contain("Must provide valid columns for the resource")
+    end
+  end
+
+  it "does not accept namespaced resources yet" do
+    with_cleanup do
```
paulcsmith: Cleanup, just in case an error is not triggered and it creates some files accidentally

tasks/gen/resource/browser.cr
```diff
@@ -0,0 +1,194 @@
...
+    io.puts "Generated #{class_name.colorize.green} in #{filename.colorize.green}"
+  end
+
+  private def resource_name
+    resource_name?.not_nil!
+  end
+
+  private def resource_name?
+    ARGV.first?
+  end
```
paulcsmith: I'm not loving how these generators work. In the future I want to extract something to consolidate a lot of the common stuff and make it simpler (things like log output, validations, etc)

PR #575, opened 2018-08-19
Fix Missing hash key error when accessing nested_key on json params

Xosmond:
Hello just encountered a bug when getting a hash value by key, when accessing the hash from the parsed json:

`Missing hash key`

### DIFF
--- a/spec/lucky/params_spec.cr
+++ b/spec/lucky/params_spec.cr
@@ -142,6 +142,16 @@ describe Lucky::Params do
       params.nested?(:user).should eq({"name" => "Paul", "age" => "28"})
     end
 
+    it "gets empty JSON params when nested key is missing" do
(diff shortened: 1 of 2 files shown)


Xosmond:
Added Paul :)

spec/lucky/params_spec.cr
```diff
@@ -142,6 +142,16 @@ describe Lucky::Params do
...
 
+    it "gets empty JSON params when nested key is missing" do
+      request = build_request body: "{}",
+        content_type: "application/json"
+      request.query = "from=query"
+
+      params = Lucky::Params.new(request)
+
+      params.nested?(:user).should eq({} of String => JSON::Any)
+    end
```
paulcsmith: Perfect! Thank you

Issue #576, opened 2018-08-21
Implement "send_file"-like from within an action


jayroh:
Rails has `send_file`: https://api.rubyonrails.org/v5.1/classes/ActionController/DataStreaming.html#method-i-send_file

Kemal has `send_file`: https://github.com/kemalcr/kemal/blob/a5870e7d24e5ec75c956bcf3e4423f55a2c4ff78/src/kemal/helpers/helpers.cr#L112

Would be great to see something like that supported in Lucky's base action class(es).  Something like 

```crystal
route do
  file "path/to/file.jpg"

  # or, to allow the browser to render in-line:
  file "path/to/file.jpg", inline: true
end
```

jayroh:
Closed with PR #589

PR #589, opened 2018-08-27
Add support for HTTP responses with files

jayroh:
Addressing issue #576.

Similar to Rails' ActionController [send_file] method, this commit adds
support for the renderable class to not only render text, but to also
respond with the contents of a file.

This also allows for designating whether a file should be downloaded
upon hitting the action, or rendered inline (for images, or maybe a text
file).

[send_file]: https://edgeapi.rubyonrails.org/classes/ActionController/DataStreaming.html#method-i-send_file

Example usage:

```crystal
class Avatars::Show < BrowserAction
  get "/avatars/:id" do
    file "public/assets/images/users/#{id}.jpg",
      disposition: "inline",        # default: "attachment"
      filename: "avatar-#{id}.jpg", # setting filename useful when using "attachment" disposition
      content_type: "image/jpeg"    # useful for when generating image on the fly, or when file
                                    # does not have an extension
  end
end
```


spec/lucky/file_response_spec.cr
```diff
@@ -0,0 +1,101 @@
...
+include ContextHelper
+
+describe Lucky::FileResponse do
+  describe "#print" do
+    describe "file is missing" do
+      it "raises an exception" do
+        context = build_context
+
+        expect_raises Lucky::Exceptions::MissingFile, /^Cannot read file/ do
+          print_file_response(context, "nope")
```
paulcsmith: One cool feature in Crystal is that you can use keyword args for any argument (you don't need to specify them in the method signature as a keyword arg). So you could do `print_file_response(context, file: "nope")` so it is clearer what the argument is for. This could be done in the other ones as well

spec/lucky/file_response_spec.cr
```diff
@@ -0,0 +1,101 @@
...
+        end
+      end
+    end
+
+    describe "status_code" do
+      it "uses the default status if none is set" do
+        context = build_context
+        print_file_response(context, "lucky_logo.png")
+        context.response.status_code.should eq Lucky::TextResponse::DEFAULT_STATUS
```
paulcsmith: Could you add newlines between setup/exercise/verify? I think this is plenty readable as-is, but adding the newlines would fit in with what is done in most of the other specs

src/lucky/file_response.cr
```diff
@@ -0,0 +1,81 @@
...
+                 @path : String,
+                 @content_type : String? = nil,
+                 @disposition : String = "attachment",
+                 @filename : String? = nil,
+                 @status : Int32? = nil,
+                 @debug_message : String? = nil)
+  end
+
+  def print
+    raise Lucky::Exceptions::MissingFile.new("Cannot read file #{path}") unless file_exists?
```
paulcsmith: Instead of manually passing in the message, what do you think about just passing the path, and setting the full message in the error class similar to the `MissingParam` exceptions?

src/lucky/file_response.cr
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def set_response_headers
+    context.response.content_length = File.size(full_path)
+    context.response.content_type = content_type
+    context.response.headers["Accept-Ranges"] = "bytes"
+    context.response.headers["X-Content-Type-Options"] = "nosniff"
+    context.response.headers["Content-Transfer-Encoding"] = "binary"
+
+    if filename
```
paulcsmith: Just to make this extra clear, what do you think about something like `if filename_overridden?` or `if custom_filename?`?

src/lucky/file_response.cr
```diff
@@ -0,0 +1,81 @@
...
+      ".svg"   => "image/svg+xml",
+      ".txt"   => "text/plain",
+      ".webm"  => "video/webm",
+      ".woff"  => "application/font-woff",
+      ".woff2" => "font/woff2",
+      ".xml"   => "application/xml",
+      ""       => "application/octet-stream",
+    }[extension]
+
+    return inferred_content_type
```
paulcsmith: Instead of the `return` you could remove the local variable and return and it should work the same

src/lucky/renderable.cr
```diff
@@ -96,6 +96,22 @@ module Lucky::Renderable
...
+                   status : Int32? = nil)
+    Lucky::FileResponse.new(context, path, content_type, disposition, filename, status)
+  end
+
+  private def file(path,
+                   content_type : String? = nil,
+                   disposition : String = "attachment",
+                   filename : String? = nil,
+                   status : Lucky::Action::Status = Lucky::Action::Status::OK)
+    Lucky::FileResponse.new(context, path, content_type, disposition, filename, status.value)
```
paulcsmith: Instead of instantiating the `FileResponse` here, what do you think about calling the other `file` method. That way there is just one method that is the source of all truth. Right now it probably doesn't buy us much to do this, but in the future it might be helpful.

spec/lucky/file_response_spec.cr
```diff
@@ -0,0 +1,101 @@
+require "../spec_helper"
+
+include ContextHelper
+
```
paulcsmith: I think it might be nice to have one integration type spec. You can see an example here: https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/spec/lucky/action_rendering_spec.cr. I'm imaging it would have a describe for "sending files"

src/lucky/file_response.cr
jayroh: I like that, yeah. I think "custom_filename?" is a good one

src/lucky/file_response.cr
jayroh: duh, of course. Remnants of some previous hacking around with this.

spec/lucky/file_response_spec.cr
jayroh: Agreed. Workin on it!

jayroh:
@paulcsmith feedback was greatly appreciated! Kept me honest 😄. I think I addressed everything you noted by now and I feel good about it!

One question for you: if there's any good spot to do some inline documentation here, where would it be? Any auto-generated docs from the source should definitely have a note about this in some shape.

Furthermore - after shipping this I'll pull down the website and do a little something in the guides to make sure this is surfaced.

jayroh:
Also: this will need a rebase and squash pre-merge :)

src/lucky/renderable.cr
```diff
@@ -96,6 +96,22 @@ module Lucky::Renderable
     end
   end
 
+  private def file(path,
```
paulcsmith: If you could add some inline documentation here that would be awesome. You can add `# :ditto:` to the method below to it is copied for that file too.

jayroh: So, I added some docs in there but the method in `Renderable` that delegates to the `FileResponse` class is private, so I guess it doesn't get generated to `./docs`. I put it, instead, inside the `FileResponse` class.  I think this is better actually as it gets closer to the implementation than a method that pretty much only delegates to it.

What do you think, @paulcsmith?

src/lucky/file_response.cr
```diff
@@ -0,0 +1,125 @@
+class Lucky::FileResponse < Lucky::Response
+  # Return a file's contents for the request.
```
paulcsmith: I think this comment should go above the class declaration since it doesn't correspond to any method in particular

src/lucky/renderable.cr
paulcsmith: Great idea! ❤️

src/lucky/file_response.cr
```diff
@@ -0,0 +1,125 @@
...
+  # ```crystal
+  # class Rendering::File < Lucky::Action
+  #   get "/file" do
+  #     file "spec/fixtures/lucky_logo.png"
+  #   end
+  # end
+  # ```
+  #
+  # For a plain text file with no extension, have it downloaded with the file
+  # named "custom.html"and the content_type "text/html"
```
paulcsmith: Space after `html"`

src/lucky/file_response.cr
jayroh: Ooooooo

jayroh:
Excellent! I think it's ready to go!

paulcsmith:
🎉  :D

PR #600, opened 2018-09-09
Add Lucky::Request class from lucky_cli's AppRequest

### LINKED ISSUE #557: Add Lucky::Request for testing api applications
Related to https://github.com/luckyframework/lucky_cli/pull/223

Essentially move the `AppRequest` module to `Lucky` as `Lucky::Request`

mikeeus:
Closes #557 

This PR moves the AppRequest class over to Lucky as Lucky::Request and adds specs. A couple notable differences:

- Added url, headers and query_params getters to keep track of request state for specs.
```crystal
  @url : String? = nil
  @headers : HTTP::Headers? = nil
  @query_params : String? = nil
```
- Removed user type on `as user` param. (used to be `Authentic::PasswordAuthenticatable` but it is still expected that the user has a `generate_token` method.

### DIFF
--- a/src/lucky/request.cr
+++ b/src/lucky/request.cr
@@ -1,0 +1,79 @@
+require "http/client"
+
+class Lucky::Request
+  getter! response
+  @response : HTTP::Client::Response? = nil
+
+  def get(path : String, as user : Authentic::PasswordAuthenticatable? = nil, headers : HTTP::Headers? = nil)
+    if user
+      headers ||= HTTP::Headers.new
+      headers.add "Authorization", user.generate_token
+    end
+
+    request("GET", path, nil, headers)
+  end
+
+  def put(path : String, body : Hash(String, String), as user : Authentic::PasswordAuthenticatable? = nil, headers : HTTP::Headers? = nil)
+    if user
+      headers ||= HTTP::Headers.new
+      headers.add "Authorization", user.generate_token
+    end
+
+    request("PUT", path, body, headers)
+  end
+
+  def post(path : String, body : Hash(String, String), as user : Authentic::PasswordAuthenticatable? = nil, headers : HTTP::Headers? = nil)
+    if user
+      headers ||= HTTP::Headers.new
+      headers.add "Authorization", user.generate_token
+    end
+
+    request("POST", path, body, headers)
+  end
+
+  def delete(path : String, as user : Authentic::PasswordAuthenticatable? = nil, headers : HTTP::Headers? = nil)
+    if user
+      headers ||= HTTP::Headers.new
+      headers.add "Authorization", user.generate_token
+    end
+
+    request("DELETE", path, nil, headers)
+  end
+
+  def response
+    @response.not_nil!
+  end
+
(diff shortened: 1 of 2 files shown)


mikeeus:
@paulcsmith I think it's relatively simple, can you review this when you have time?

src/lucky/request.cr
```diff
@@ -0,0 +1,87 @@
...
+  @url : String? = nil
+  @headers : HTTP::Headers? = nil
+  @query_params : String? = nil
+  @response : HTTP::Client::Response? = nil
+
+  def get(path : String, as user = nil, headers : HTTP::Headers? = nil)
+    if user
+      headers ||= HTTP::Headers.new
+      headers.add "Authorization", user.generate_token
+    end
```
robacarp: is it worth making these 4 lines into a helper function?

mikeeus: Thanks for pointing that out!

src/lucky/request.cr
```diff
@@ -0,0 +1,72 @@
+require "http/client"
+
+class Lucky::Request
+  getter! url
+  getter! headers
+  getter! query_params
+  getter! response
```
paulcsmith: I’ve been thinking about these getters and I think we can do something so they can be removed. The advantage is that we don’t need the ! Which means we have a bit better type safety. 

What I think could be done is to set up a TestServer like we did in LuckyFlow: 
https://github.com/luckyframework/lucky_flow/blob/3878da1bead09df01f65dc55f755c09b28ba7598/spec/spec_helper.cr#L10

https://github.com/luckyframework/lucky_flow/blob/master/spec/support/test_server.cr

It would probably work a bit differently in that the TestServer would record the HTTP method, headers and path so it can be inspected. 

Then in the specs you can make a real request to the test server instead of testing that the instance variables are set in Lucky:mRequest. In fact, I think you could remove the instance variables since they don’t need to be accessed outside the class anymore. 

Lmk if this makes sense. It’s a bit tricky to implement and I may nit have explained it well. I’m happy to go more in depth if that would help.

paulcsmith: Oh and here’s an example of using the test server in LuckyFlow to stub requests: 
https://github.com/luckyframework/lucky_flow/blob/master/spec/lucky_flow_spec.cr

mikeeus: From what I understand, we're wrapping the server creation in a new class, then capturing the requests we make to it so we can inspect them later.

So the class would look something like this:

```crystal
class TestServer
  delegate listen, close, to: @server

  def initialize(port : Int32)
    @server = App.new
  end

  def self.request
    # return last request
  end

  def self.response
    # return last response
  end
end
```
And we would use it like `TestServer.response.headers["Authorization"].should contain "token"`. 

Is this what you had in mind?

paulcsmith: Yeah I think that should work!

paulcsmith:
Thanks so much for starting on this @mikeeus. We went with #689 for now.

I hope things are going well for you!

PR #605, opened 2018-09-14
Log action callbacks

paulcsmith:
Closes: #604 

screen shot 2018-09-14 at 2 12 16 pm

### DIFF
--- a/src/lucky/action_callbacks.cr
+++ b/src/lucky/action_callbacks.cr
@@ -100,7 +100,10 @@ module Lucky::ActionCallbacks
       #   end
 
       if callback_result.is_a?(Lucky::Response)
+        Lucky::ActionCallbacks.log_stopped_callback(context,"{{ callback_method.id }}")
         return callback_result
+      else
+        Lucky::ActionCallbacks.log_continued_callback(context,"{{ callback_method.id }}")
       end
     {% end %}
   end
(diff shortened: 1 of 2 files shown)


src/lucky/action_callbacks.cr
```diff
@@ -120,11 +123,32 @@ module Lucky::ActionCallbacks
...
+    context.add_debug_message("Stopped at #{callback_method_with_color}")
+  end
+
+  # :nodoc:
+  def self.log_continued_callback(
+    context : HTTP::Server::Context,
+    callback_method_name : String
+  ) : Void
+    callback_method_with_color = callback_method_name.colorize(HTTP::Server::Context::DEBUG_COLOR)
+    context.add_debug_message("Ran #{callback_method_with_color}")
```
edwardloveall: What do you think about `Callback: method_name` here instead? Ran is accurate but maybe not as instructive?

paulcsmith: I had that at first, but it made the log *super* noisy, especially when there were 5 or 6 callbacks. I think it is more clear, but once you know what it means I think the shorter version is nicer.

Can you think of a way to make it clear *and* short?

edwardloveall: I think `Ran` is fine then.

paulcsmith: Ok 🏃‍♂️

paulcsmith:
Thanks for the review!

PR #609, opened 2018-09-14
Refactor the cookie system

edwardloveall:
The tl;dr of this is:

* Cookie sub-system has been refactored to be more safe and debuggable
* Now you use `cookies.set(:key, "value")`, `cookies.get(:key)`, and `cookies.get?(:key)`
* Old `[]=`, `[]`, and `[]?` syntax doesn't work
* Flash is has `success`, `info`, and `failure` keys now (instead of `warning`)
* You'll want to make sure you have a session key configured. probably best to call is `_name_of_app_session`:

```
Lucky::SessionCookie.configure do |settings|
  settings.key = "_app_session"
end
```

See also:
* https://github.com/luckyframework/lucky_cli/pull/256
* https://github.com/luckyframework/authentic/pull/16


src/lucky/cookies/cookie_jar.cr
```diff
@@ -0,0 +1,56 @@
+class Lucky::CookieJar
+  alias Key = String | Symbol
+  private property store : HTTP::Cookies
+  @changed = false
+  EXPIRATION      = 1.year.from_now
```
edwardloveall: @Sija made a good point that this should be a method (or value) so if a crystal app is running for a long time, the value will be updated.

If you compile this gist you can see that the constant stays, well, constant while the method updates: https://gist.github.com/edwardloveall/da12432b41704187c1a175dceffe3e7a

src/lucky/cookies/flash_store.cr
```diff
@@ -0,0 +1,53 @@
...
+
+  @now = {} of String => String
+  @next = {} of String => String
+
+  def self.from_session(session : Lucky::SessionCookie) : Lucky::FlashStore
+    new.from_session(session)
+  end
+
+  def from_session(session : Lucky::SessionCookie) : Lucky::FlashStore
+    json = session.get?(SESSION_KEY) || "{}"
```
Sija:
```cr
session.get?(SESSION_KEY).try do |json|
  JSON.parse(json).as_h.each do |key, value|
    @now[key.to_s] = value.to_s
  end
end
```

This will avoid parsing an empty object (`"{}"`).

src/lucky/cookies/cookie_jar.cr
Sija: This should be `1.year` (`Time::Span`).
Also possibly a `class_property` to make it configurable if needed.

src/lucky/cookies/session_cookie.cr
```diff
@@ -0,0 +1,45 @@
...
+  Habitat.create do
+    setting key : String
+  end
+
+  alias Key = String | Symbol
+  private property store = {} of String => String
+  @changed = false
+
+  def initialize(cookie : Lucky::MaybeCookie = Lucky::NullCookie.new)
+    contents = cookie.value || "{}"
```
Sija: ditto

src/lucky/cookies/flash_store.cr
```diff
@@ -0,0 +1,53 @@
...
+    self.class.new
+  end
+
+  private def all
+    @now.merge(@next)
+  end
+
+  {% for shortcut in [:failure, :info, :success] %}
+    def {{ shortcut.id }} : String
+      get?(:{{ shortcut.id }}) || ""
```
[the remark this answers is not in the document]
edwardloveall: Great idea! Added in be0ebcd

spec/spec_helper.cr
```diff
@@ -11,6 +11,10 @@ Lucky::SessionCookie.configure do |settings|
   settings.key = "_app_session"
 end
 
+Lucky::CookieJar.configure do |settings|
+  settings.expiration = 1.year
```
Sija: `expiration` -> `default_expiration`?

src/lucky/cookies/flash_store.cr
```diff
@@ -31,6 +31,10 @@ class Lucky::FlashStore
       get?(:{{ shortcut.id }}) || ""
     end
 
+    def {{ shortcut.id }}? : String?
+      get?(:{{ shortcut.id }}) || nil
```
Sija: ` || nil` part is redundant, IIRC `#get?` already returns nilable value.

src/lucky/context_extensions.cr
```diff
@@ -1,24 +1,27 @@
...
+  setter cookies : Lucky::CookieJar?
+  setter session : Lucky::SessionCookie?
+  setter flash : Lucky::FlashStore?
 
   getter debug_messages : Array(String) = [] of String
   property? hide_from_logs : Bool = false
 
   def cookies
-    @cookies ||= Lucky::Cookies::Store.build(request, Lucky::Server.settings.secret_key_base)
+    @cookies ||= Lucky::Cookies::Processors::Encryptor.read(
```
Sija: I'd suggest using `abstract class Lucky::Cookies::Processor` along with composition pattern, which will decouple processor from the `CookieJar` class itself.

src/lucky/context_extensions.cr
```diff
@@ -1,24 +1,27 @@
-require "./session/**"
-
 class HTTP::Server::Context
   DEBUG_COLOR = :green
-  setter session : Lucky::Session::AbstractStore?
```
Sija: Having abstract store would allow using another backends, like redis for instance.

src/lucky/cookies/flash_store.cr
edwardloveall: Hup, right you are. Thanks for all the careful looking at this PR!

src/lucky/context_extensions.cr
edwardloveall: I think we'll want that eventually, but for now we're only backing the session with cookies. When that time comes, you suggestion is excellent! 👍

spec/lucky/cookies/processors/encryptor_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+
+include ContextHelper
+
+describe Lucky::Cookies::Processors::Encryptor do
+  describe ".read" do
+    it "returns a decrypted CookieJar" do
+      a_value = encryptor.encrypt("some cookie value")
+      a_value = Base64.strict_encode(a_value)
+      b_value = encryptor.encrypt("another cookie value")
+      b_value = Base64.strict_encode(b_value)
```
paulcsmith: I was a bit tripped out about the double assignment. What do you think about putting it on one line? Maybe with shorter text so it fits?

spec/lucky/flash_handler_spec.cr
```diff
@@ -2,22 +2,30 @@ require "../../spec_helper"
...
+    context1.flash.success.should eq("Yay!")
+
+    Lucky::FlashHandler.new.call(context1)
+
+    next_json = context1.session.get(Lucky::FlashStore::SESSION_KEY)
+    context2 = build_context
+    context2.session.set(Lucky::FlashStore::SESSION_KEY, next_json)
+
+    context2.flash.success.should eq("")
+  end
```
paulcsmith: The previous spec also checks that flash messages set for the next request show up in the next request. I may be misreading this but I don’t think this new spec covers that

spec/lucky/session_handler_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+  it "persists the cookies across multiple requests" do
+    context_1 = build_context
+    context_1.cookies.set(:email, "test@example.com")
+    Lucky::SessionHandler.new.call(context_1)
+
+    request = build_request
+    cookie_header = context_1.response.cookies.map do |cookie|
+      cookie.to_cookie_header
+    end.join(", ")
+    request.headers.add("Cookie", cookie_header)
```
paulcsmith: Could you explain why cookies.set and setting the header is required? I’m not sure I understand

src/charms/cookie.cr
```diff
@@ -0,0 +1,36 @@
...
+  def secure(value : Bool)
+    self.secure = value
+    self
+  end
+
+  def http_only(value : Bool)
+    self.http_only = value
+    self
+  end
+end
```
paulcsmith: I’m loving these shortcuts!

src/lucky/context_extensions.cr
```diff
@@ -1,24 +1,27 @@
...
+      from: request
+    )
   end
 
   def session
-    @session ||= Lucky::Session::Store.new(cookies).build
+    @session ||= begin
+      cookie = cookies.get?(Lucky::SessionCookie.settings.key)
+      Lucky::SessionCookie.new(cookie)
+    end
```
paulcsmith: Nice! I like this approach

src/lucky/context_extensions.cr
```diff
@@ -1,24 +1,27 @@
...
-    @cookies ||= Lucky::Cookies::Store.build(request, Lucky::Server.settings.secret_key_base)
+    @cookies ||= Lucky::Cookies::Processors::Encryptor.read(
+      from: request
+    )
   end
 
   def session
-    @session ||= Lucky::Session::Store.new(cookies).build
+    @session ||= begin
+      cookie = cookies.get?(Lucky::SessionCookie.settings.key)
```
paulcsmith: I don’t think you need to set this local var

src/lucky/cookies/cookie_jar.cr
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def clear
+    @store = HTTP::Cookies.new
+  end
+
+  def who_took_the_cookies_from_the_cookie_jar?
+    raise "Edward Loveall"
+  end
```
paulcsmith: Thief!

src/lucky/cookies/flash_store.cr
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def from_session(session : Lucky::SessionCookie) : Lucky::FlashStore
+    session.get?(SESSION_KEY).try do |json|
+      JSON.parse(json).as_h.each do |key, value|
+        @now[key.to_s] = value.to_s
+      end
+    end
+    self
+  rescue e : JSON::ParseException
```
paulcsmith: I don’t think we should rescue this like it was before. I feel like if there is an error parsing that should be raised. What do you think?

src/lucky/cookies/session_cookie.cr
```diff
@@ -0,0 +1,51 @@
+class Lucky::SessionCookie
+  Habitat.create do
+    setting key : String
+  end
+
+  alias Key = String | Symbol
+  private property store = {} of String => String
+  @changed = false
+
+  def initialize(cookie : Lucky::MaybeCookie = Lucky::NullCookie.new)
```
paulcsmith: I’m a bit nervous about initializing this with a NullCookie. It seems like that might be set on accident. What do you think about requiring setting the cookie and having another class method that sets null cookie:

def self.null
  new.tap do |session|
    session.store = NullCookie.new
  end
end

I’m on my iPad so sorry about any typos:P

src/lucky/session_handler.cr
```diff
@@ -4,9 +4,31 @@ class Lucky::SessionHandler
   def call(context : HTTP::Server::Context)
     call_next(context)
 
-    if context.session.changed?
-      context.session.set_session
-      context.cookies.write(context.response.headers)
+    write_session(context)
+    write_cookies(context)
+    check_cookie_size(context)
```
paulcsmith: I LOVE how this reads!

src/lucky/session_handler.cr
```diff
@@ -4,9 +4,31 @@ class Lucky::SessionHandler
...
+  end
+
+  private def write_session(context : HTTP::Server::Context)
+    return if !context.session.changed?
```
paulcsmith: It’s not documented but so far most of the codebase avoids early returns. In a case like this I think: if context.session.changed? Should work well

src/lucky/cookies/flash_store.cr
edwardloveall: Yeah I think you're right. Raising is better than silent failing. Maybe I could present the contents of the flash in the error?

paulcsmith: Good idea! I think that would be helpful for debugging

src/lucky/context_extensions.cr
paulcsmith: Never mind! I reviewed this on my iPad and for whatever reason it looked like 18 and 19 were the same line. This looks good as-is 👍

spec/lucky/flash_handler_spec.cr
edwardloveall: Right you are. Added in https://github.com/luckyframework/lucky/pull/609/commits/e9463f09d8c40d246911ef64c5a5c8f2e8726540

src/lucky/cookies/flash_store.cr
edwardloveall: Added in https://github.com/luckyframework/lucky/pull/609/commits/ef17cb359d8cc74c48fb026705d4189faaad6131

src/lucky/cookies/session_cookie.cr
edwardloveall: I added this https://github.com/luckyframework/lucky/pull/609/commits/159e34b7fe70392449727e4f5b62d19ebf243cdb

As the commit message calls out, it required that I created a new way for the session to bootstrap itself. Lemme know what you think of what I did.

I have a `Lucky::CooikeJar.session_cookie` method which is what you use to initialize a `Lucky::SessionCookie` which seems... wrong? Bad? What do you think? Should I rename the method or class?

src/lucky/session_handler.cr
edwardloveall: Nailed it: 8cca80f3

edwardloveall:
Note about the failing test: I'm not sure why it's failing. My guess is that crystal changed the formatter slightly and so checking against 0.26.1 changes nothing but master fails. I haven't gotten down dirty with building crystal on my own yet to figure it out.

Possible place it changed: https://github.com/crystal-lang/crystal/commit/ff5f54144db618f86a5cab307bf9aacce1080bea

spec/lucky/session_handler_spec.cr
paulcsmith: I think this makes sense, but it isn't super clear from the code. Maybe adding something to the description like `persists the cookies across multiple requests when set with header or in Lucky`. That sounds horrible, but maybe that triggers some ideas?

src/lucky/cookies/cookie_jar.cr
```diff
@@ -0,0 +1,72 @@
+class Lucky::CookieJar
+  alias Key = String | Symbol
+  private property store : HTTP::Cookies
+  @changed = false
+  MAX_COOKIE_SIZE = 4096
+
+  Habitat.create do
+    setting default_expiration : Time::Span | Time::MonthSpan
```
paulcsmith: I didn't know there was a `MonthSpan`! Cool

spec/lucky/session_handler_spec.cr
edwardloveall: Good distinction of "proper code but description could be clearer" 👍 Updated the description and some variables in a61bc1b.

edwardloveall:
I swear I did `crystal tool format` already, but not it worked ¯\\\_(ツ)_/¯

src/lucky/page_helpers/time_helpers.cr
```diff
@@ -35,7 +35,7 @@ module Lucky::TimeHelpers
 
   private def distance_in_minutes(distance : Int32) : String
     case distance
-    when      1 then "a minute"
+    when 1      then "a minute"
     when 2...45 then "#{distance} minutes"
     else
       "about an hour"
```
Sija: Is this correct? Could `distance` be more than 60 minutes?

paulcsmith: Good question! It’s inky when less than 60: 
https://github.com/luckyframework/lucky/blob/0e6ee45d851ccc872ae06882a11ee354d0af37e2/src/lucky/page_helpers/time_helpers.cr#L10

Issue #614, opened 2018-09-25
css_link method outputs wrong markup for link tag


jwoertink:
When you use `css_link` it will output a `<link></link>` tag which is invalid markup.

PR #615, opened 2018-09-25
fixing css_link method. Fixes: #614

jwoertink:
The current implementation causes the `link` tag to output a closing `</link>` tag. I've added a new tag helper method called `inline_tag` which doesn't need to take a block or call yield. Just pass in options like you normally would.

### DIFF
--- a/src/lucky/tags/custom_tags.cr
+++ b/src/lucky/tags/custom_tags.cr
@@ -32,4 +32,10 @@ module Lucky::CustomTags
     yield
     @view << "</#{name}>"
   end
+
+  def inline_tag(name : String, options = EMPTY_HTML_ATTRS, **other_options)
+    merged_options = merge_options(other_options, options)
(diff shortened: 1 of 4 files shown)


src/lucky/tags/custom_tags.cr
```diff
@@ -32,4 +32,10 @@ module Lucky::CustomTags
     yield
     @view << "</#{name}>"
   end
+
+  def inline_tag(name : String, options = EMPTY_HTML_ATTRS, **other_options)
```
paulcsmith: What do you think of calling this empty_tag? I think they are usually referred to as empty/void tags https://www.456bereastreet.com/archive/201005/void_empty_elements_and_self-closing_start_tags_in_html/

paulcsmith: Oh and could you add some brief docs for this new method noting that it will not be generated with an end tag?

paulcsmith:
Thanks!

PR #616, opened 2018-10-01
added documentation for the from_headers method

ajwann:
I tried my best to follow the guide on writing proper docs. Please let me know if I've done it correctly and I'll put up a couple more PR's.


edwardloveall:
Oof this is bad luck. We're right in the middle of a large refactor of the cookie system (#609) so this will soon be out of date. 🙁 

Care to pick a different class to document? I don't think anyone's claimed anything in `/src/charms`?

PR #621, opened 2018-10-07
added script to run tests and related commands

ajwann:
If everything looks good with this test script, I can add it to the other lucky libraries.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,8 +1,8 @@
 #! /bin/bash
 
-printf "\nRunning 'crystal spec'\n\n"
+printf "\nrunning specs with 'crystal spec'\n\n"
 crystal spec
-printf "\nRunning 'shards build'\n\n"
+printf "\nbuilding shards with 'shards build'\n\n"
 shards build
-printf "\nRunning 'crystal tool format src spec'\n\n"
+printf "\nformatting code with 'crystal tool format src spec'\n\n"
 crystal tool format src spec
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+printf "\nRunning 'crystal spec'\n\n"
+crystal spec
+printf "\nRunning 'shards build'\n\n"
+shards build
+printf "\nRunning 'crystal tool format src'\n\n"
+crystal tool format src
```
paulcsmith: Could you add “spec” after  “src” so it formats both directories?

bin/test
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+printf "\nRunning 'crystal spec'\n\n"
+crystal spec
+printf "\nRunning 'shards build'\n\n"
+shards build
+printf "\nRunning 'crystal tool format src'\n\n"
```
ajwann: I'm not sure if we want to run ```crystal tool format``` on more than the source directory. I did see some errors when I ran it on the project root.

paulcsmith: Maybe “formatting code with crystal tool format src spec”

bin/test
ajwann: Sure, no problem.

ajwann:
Actually let me add one more step to the contributing section, it looks a bit odd the way I left it.

PR #626, opened 2018-10-11
Adds some extra specs. fixes #601

### LINKED ISSUE #601: Properly support HEAD requests
https://ochronus.com/http-head-request-good-uses/

GET requests should respond to HEAD, but should only return headers and not the response body

jwoertink:
Adds specs to confirm that using a HEAD request on a GET route works with latest LuckyRouter update. 

These specs will fail until [LuckyRouter](https://github.com/luckyframework/lucky_router/commit/9a8b31872a44429a921280da91acf2604a33dc3a) is updated to a new version, and this repo pulls in that new update.

### DIFF
--- a/spec/lucky/text_response_spec.cr
+++ b/spec/lucky/text_response_spec.cr
@@ -22,9 +22,22 @@ describe Lucky::TextResponse do
       print_response(context, status: nil)
       context.response.status_code.should eq 300
     end
+
+    it "prints no body with a head call" do
+      context = build_context("HEAD")
+      print_response_with_body(context, "Body", "text/plain", nil)
+      context.request.method.should eq "HEAD"
+      context.request.body.to_s.should eq ""
+      context.response.status_code.should eq 200
+      context.response.headers["Content-Type"].should eq "text/plain"
+    end
   end
 end
 
 private def print_response(context : HTTP::Server::Context, status : Int32?)
-  Lucky::TextResponse.new(context, "", "", status: status).print
+  print_response_with_body(context, "", "", status)
+end
+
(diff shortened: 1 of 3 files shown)


paulcsmith:
To get this passing could you change the version to `branch: master`? And could you make an issue to remember to roll it back?

spec/lucky/text_response_spec.cr
```diff
@@ -22,9 +22,22 @@ describe Lucky::TextResponse do
...
+    end
   end
 end
 
 private def print_response(context : HTTP::Server::Context, status : Int32?)
-  Lucky::TextResponse.new(context, "", "", status: status).print
+  print_response_with_body(context, "", "", status)
+end
+
+private def print_response_with_body(context : HTTP::Server::Context, body : String = "", content_type : String = "", status : Int32? = nil)
```
paulcsmith: Since this is only used by the one spec, I think you can make `body` required (no default) and remove the content_type and status

spec/lucky/text_response_spec.cr
```diff
@@ -22,9 +22,22 @@ describe Lucky::TextResponse do
...
       context.response.status_code.should eq 300
     end
+
+    it "prints no body with a head call" do
+      context = build_context("HEAD")
+      print_response_with_body(context, "Body", "text/plain", nil)
+      context.request.method.should eq "HEAD"
+      context.request.body.to_s.should eq ""
+      context.response.status_code.should eq 200
+      context.response.headers["Content-Type"].should eq "text/plain"
```
paulcsmith: It is probably ok to skip this expectation

Issue #599, opened 2018-09-05
Add link back to resource from ShowPage when generating gen.resource.browser


paulcsmith:

paulcsmith:
@citizen428 I added a description here as well :)

Issue #598, opened 2018-09-04
Add "New" button to gen.resource.browser


paulcsmith:

citizen428:
If you add a slightly more verbose description for people who are out of the loop I can make this (and similar/related things like #599) part of my Hacktoberfest, it's been too long since I contributed here 😃

paulcsmith:
That would be wonderful! Description added :)

citizen428:
Great, thanks @paulcsmith! I probably won't get around to it until the weekend, but we have a long one coming up, so hopefully I'll manage to find time for all of them :-)

PR #629, opened 2018-10-13
Resource browser

### LINKED ISSUE #596: Remind user to view migration and run it after gen.resource.browser
Something like "Run generated migrations with #{"lucky db.migrate".colorize.green}"

Before: 
https://github.com/luckyframework/lucky/blob/6e79ccb93d4c3d451181700d3af665a102144796/tasks/gen/resource/browser.cr#L19

citizen428:
Closes #596, #598, #599 

@paulcsmith Figured a single PR is easier than 3 separate ones considering we are talking about 3 line of changes here. Also: I made "New" a link, not a button, since it seemed more in line with actions on `ShowPage`.

### DIFF
--- a/tasks/gen/templates/resource/pages/{{folder_name}}/index_page.cr.ecr
+++ b/tasks/gen/templates/resource/pages/{{folder_name}}/index_page.cr.ecr
@@ -3,7 +3,8 @@ class <%= pluralized_resource %>::IndexPage < MainLayout
   quick_def page_title, "All"
 
   def content
-    h1 "All"
+    h1 "All <%= pluralized_resource %>"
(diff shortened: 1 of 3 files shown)


tasks/gen/templates/resource/pages/{{folder_name}}/index_page.cr.ecr
```diff
@@ -4,6 +4,7 @@ class <%= pluralized_resource %>::IndexPage < MainLayout
 
   def content
     h1 "All"
```
paulcsmith: While you’re in here could you add the resource name to this?

citizen428:
@paulcsmith Done 😃

PR #640, opened 2018-10-22
Have the error handler return a 422 on InvalidParam instead of 500.

### LINKED ISSUE #628: Discussion: Handling Lucky::Exceptions::InvalidParam
When you have an action that requires a param like `page : Int32`, lucky will throw the `Lucky::Exceptions::InvalidParam` error and return a 500. In cases of people trying to "hack" you, this can cause issues in environments like Elasticbeanstalk where too many 500 errors will cause your instance to be auto-scaled down.

```
Required param "page" with value "32B" couldn't be parsed to a "Int32"
```

After some talks about this, it may be better to have the server return a [422](https://httpstatuses.com/422). This is the only example I've come across, but there may be other examples. I think this exception should still be raised so that way it's still reported to error reporting, but have the server return a status not 500 range.

jwoertink:
fixes #628 

We still raise the exception to allow people to log it in to their status reporting tools, but when it's caught by the ErrorHandler, the server would now return a 422 instead of a 500 error.


src/lucky/error_handler.cr
```diff
@@ -34,8 +34,18 @@ class Lucky::ErrorHandler
...
     context
   end
+
+  private def status_code_by_error(error : Exception)
+    case error
+    when Lucky::Exceptions::InvalidParam
+      Lucky::Action::Status::UnprocessableEntity.value
+    else
+      Lucky::Action::Status::InternalServerError.value
+    end
```
paulcsmith: What if instead of using conditionals we add a module that can be mixed in to errors. Then any error can implement it and set a status code. Here's roughly what I mean:

```crystal
module Lucky::DefaultErrorResponse
  abstract def default_lucky_status_code : Int32
end
```

Then an error can implement it:

```crystal
class Lucky::Exceptions::InvalidParam
  include Lucky::DefaultErrorResponse
  
  def default_lucky_status_code
    422
  end
end
```

In the handler you could do something like this:

```
private def status_code_by_error(error: Exception)
  Lucky::Action::Status::InternalServerError.value
end

private def status_code_by_error(error: Lucky::DefaultErrorResponse)
  error.default_lucky_status_code
end
```

That way we can easily add status code for other errors, or shards can add this too w/o us needing to change Lucky. What do you think? Does that make sense?

Also I don't love the names I suggested :P If you have a better idea definitely feel free to change my suggestion!

jwoertink: Interesting! I like that better. Let me take a look.

paulcsmith: Cool! Another idea for the module might be `Lucky::HasDefaultErrorResponse`

jwoertink:
Ok, I made that change. I like how it reads a lot better, but naming is hard. 

Here's a few names I thought up. Let me know if you like any of these better, or think of an even better one
```crystal
Lucky::ErrorResponseCode 
Lucky::ErrorResponseHelper 
Lucky::ErrorRespondable 
Lucky::DefaultErrorResponse
```

src/lucky/exceptions.cr
```diff
@@ -22,6 +23,10 @@ module Lucky
       def message : String?
         "Required param \"#{param_name}\" with value \"#{param_value}\" couldn't be parsed to a \"#{param_type}\""
       end
+
+      def default_response_status_code
```
paulcsmith: Actually maybe “default” isn’t necessary. And it might be good to be clear it is an HTTP status code. So what do you think about calling it http_error_code?

src/lucky/exceptions.cr
```diff
@@ -22,6 +23,10 @@ module Lucky
...
+        422
```
paulcsmith: Maybe this could use one of the Staus enums so future readers don’t have to look it up?

paulcsmith:
I really like the Respondale name. Maybe HttpRespondable? Since we already know it’s an error and this way it’s clear it is for HTTP responses. What do you think?

spec/lucky/error_handling_spec.cr
```diff
@@ -46,6 +49,21 @@ describe Lucky::ErrorHandler do
     context.response.status_code.should eq(500)
   end
 
+  it "returns a 422 status for InvalidParam exceptions" do
```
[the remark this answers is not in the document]
jwoertink: I'd actually like to keep this one because we specially saying InvalidParam throws a 422, but I like the idea of adding a new test for that case. I've added a second case that tests for a custom response

jwoertink:
Ok those are updated. I definitely like `HttpRespondable` better. Right now it's used for an error code, but it's generic enough that if we needed another type of code, that would be a good spot.

spec/lucky/error_handling_spec.cr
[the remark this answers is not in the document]
paulcsmith: I didn't explain what I meant! I think we should still test InvalidParam, but we can unit test it in a `invalid_param_spec` that checks that the method returns a `422`.

Does that make sense? I don't think we need to go through the error handler to test every error that implements the module. We can test it in a generic fashion and then unit test errors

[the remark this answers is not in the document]
paulcsmith: For example:

```crystal
describe InvalidParamError do
it "responds with a custom HTTP code" do
  error = InvalidParamError.new()
  

  error.should be_a(Lucky::HTTPRespondable)
  error.http_code.should eq(422)
end
end
```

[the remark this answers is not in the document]
jwoertink: Oh, I see what you're saying. Ok. I'll move that one then.

spec/lucky/error_handling_spec.cr
```diff
@@ -46,6 +57,18 @@ describe Lucky::ErrorHandler do
     context.response.status_code.should eq(500)
   end
 
+  it "returns the defined status code if the raised exception defines one" do
+    error_handler = Lucky::ErrorHandler.new(action: FakeErrorAction)
+    error_handler.next = ->(_ctx : HTTP::Server::Context) {
+      raise ShortAndStoutError.new("I'm a little teapot")
```
paulcsmith: I had this song stuck in my head for the last hour now 😂

jwoertink: LOL! Nice

PR #662, opened 2018-12-31
Add `Lucky::Routable.url_without_query_params`

### LINKED ISSUE #407: Add Action.url_without_query_params
This is helpful when you want to use query params in an action, but need to generate a URL without query params. I think this is only useful when generating a URL for a webhook. For example, when you are setting up Slack auth: 

```crystal
class SlackSignIns::Create < BrowserAction
  params code : String

  get "/auth/slack" do
  end
end

# And in a view
link "Sign in with Slack", "slackapi.com/auth?redirect_uri=#{SlackSignIns::Create.url_without_query_params}"
```

Also, have a nice error message so if you try to instantiate a URL that is missing query params, ask user to provider the query params, OR use 'url_without_query_params'. Don't do this for path helpers though since I don't think it makes sense to have a path without query params. I'm open to changing that if someone has a valid use case though

HarrisonB:
Closes #407

I broke up `Lucky::Routable.route` because it seemed to be getting a bit long.

If anyone has any tips on DRYing up
```crystal
{% for param in path_params %}       
  {{ param.gsub(/:/, "").id }}
{% end %}
```
for defining arguments, that would be much appreciated!

### DIFF
--- a/src/lucky/routeable.cr
+++ b/src/lucky/routeable.cr
@@ -175,6 +175,19 @@ module Lucky::Routeable
       route(*args, **named_args).url
     end
 
+    def self.url_without_query_params(
+    {% for param in path_params %}
+      {{ param.gsub(/:/, "").id }},
+    {% end %}
+    )
+      path = path_from_parts(
+        {% for param in path_params %}
+          {{ param.gsub(/:/, "").id }},
+        {% end %}
+      )
+      Lucky::RouteHelper.new({{ method }}, path).url
+    end
+
     def self.route(
     {% for param in path_params %}
       {{ param.gsub(/:/, "").id }},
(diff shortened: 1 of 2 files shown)


src/lucky/routeable.cr
```diff
@@ -175,6 +175,19 @@ module Lucky::Routeable
...
+    def self.url_without_query_params(
+    {% for param in path_params %}
+      {{ param.gsub(/:/, "").id }},
+    {% end %}
+    )
+      path = path_from_parts(
+        {% for param in path_params %}
+          {{ param.gsub(/:/, "").id }},
+        {% end %}
+      )
```
paulcsmith: I wish I knew of a good way to DRY up macros like this, but I don't know of a way :(

PR #679, opened 2019-01-30
raise CookieOverflow when any cookie size exceeds the limit

skojin:
Now it takes all cookies, concat them, then checks size
this PR checks size of each cookie individually


skojin:
Now this path check size of all cookies, I just checked how rails 4.2 works, it check size only of signed cookies. And allow to set cookies with size > 4kb.

skojin:
@paulcsmith please review this path, this I set cookies individually, and it will help me in fix for https://github.com/luckyframework/lucky/issues/745

PR #713, opened 2019-03-02
Add with_defaults

paulcsmith:
This will allow you to use a block to set some default values
and then override or modify them. This will allow some super cool stuff
with Class components and inputs. Stuff like this:

```crystal
mount Shared::Field.new(f.email), 
  &.email_input(placeholder: "Email here...", append_class: "compact-input")
```

Where `Shared::Field` sets up default classes, field, placeholder, and
whatever else you want.

```crystal
class Shared::Field(T) < BaseComponent
  needs field : Avram::FillableField(T)

  def render
    label_for @field
    with_defaults field: @field, class: "form-input" do |html|
      yield html
    end
  end
end
```

### DIFF
--- a/src/lucky/tags/with_defaults.cr
+++ b/src/lucky/tags/with_defaults.cr
@@ -1,0 +1,88 @@
+# Set up defaults arguments for HTML tags.
+#
+# This is automatically included in Pages and Cells.
+module Lucky::WithDefaults
+  # This is typically used in Cells and helper methods to set up defaults for
+  # reusable components.
+  #
+  # Example in a page or cell:
+  #
+  #    with_defaults field: form.email, class: "input" do |html|
+  #      html.email_input placeholder: "Email"
+  #    end
+  #
+  # Is the same as:
+  #
+  #     email_input field: form.email, class: "input", placeholder: "Email"
+  def with_defaults(**named_args)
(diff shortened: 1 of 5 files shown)


src/lucky/tags/with_defaults.cr
```diff
@@ -0,0 +1,73 @@
...
+        {% if appended_class_arg %}
+          overridden_html_class = "#{@named_args[:class]?} #{{{ appended_class_arg.value }}}"
+        {% end %}
+        {% named_args = named_args.reject { |arg| arg.name == :append_class } %}
+
+        {% replace_class_arg = call.named_args.find { |arg| arg.name == :replace_class } %}
+        {% if replace_class_arg %}
+          overridden_html_class = "#{{{ replace_class_arg.value }}}"
+        {% end %}
+        {% named_args = named_args.reject { |arg| arg.name == :replace_class } %}
```
paulcsmith: This is really unpleasant. I'll try to find a way to clean this up. I also need to see how this affects compile time performance

paulcsmith: Performance can probably be improved but I think it is fine. 5000 calls with `with_defaults` and it added 1-2 seconds. This will probably be used sparingly (mostly for form inputs) so I think this is pretty reasonable

Issue #719, opened 2019-03-07
Get rid of `show_debug_output` and make error handling more intuitive


HarrisonB:
Perhaps I don't understand this setting, but just because you want to see the debug info doesn't mean you don't want error handlers to behave as they normally do (giving correct status codes, response bodies, etc.)

paulcsmith:
I think you understand it but it’s just confusing 😂 

Thank you for opening this. I want to rework this so that errors you handle are always displayed and you only get the stacktrace for unhandled ones. I’m open to ideas here but I think what I’d like to do is rename Errors::Show to UnhandledErrors::Show and then have a module called RescueErrors that is included in the base actions that lets you rescue errors and display them

Still need to figure out the internals but that’s a high level view of how I think it would work. Thoughts?

HarrisonB:
I think I understand a bit, but I have some questions:

* Would it send the stacktrace to STDERR? Or to your logger?
* Which errors would `RescueErrors` catch? Does it have priority over `UnhandledErrors::Show`?

paulcsmith:
Good question. I'm not sure though. I think for the stack trace it'd print to the logger. 

Your second question makes me think there should not be an `UnhandledErrors::Show`...because the division is unclear. I might need to rethink that a bit so there is *just* `RescueErrors`. Basically I want one place where you define error handling, but there also needs to be a way for unhandled exceptions to bubble up when in dev. I'm not 100% sure of the best way to do this yet

HarrisonB:
Why not just put that behavior in ResuceErrors or some base it inherits from but output dependent on env?

paulcsmith:
I think that might work well! Still trying to figure out how that would work with errors outside of actions though. For example, what if there is an error outside of an action, like in a middleware. If the rescued errors are in the base action classes they won't be caught. So maybe we do need two separate ones. Or maybe instead of a mixing it is a class that can be used by both actions *and* an ErrorHandler

```crystal
class BrowserAction
  rescue_using RescueErrors
end

# and in the ErrorHandler middleware
Lucky::ErrorHandler.new(rescue_using: RescueErrors)

class RescueErrors < Lucky::ErrorAction

    def handle_error(error : Exception)
    if Lucky.settings.show_debug_output?
      render_exception_page
     else
      Lucky.logger.error(error.inspect_with_backtrace)

      message = "An unexpected error occurred"

      json Errors::ShowSerializer.new(message), status: 500
  end
end
```

Very rough example but it might work

[...]

PR #721, opened 2019-03-08
Call error action regardless of `show_debug_output`

HarrisonB:
Before, the response code and body could vary with `show_debug_output`.

Also make `render_exception_page` class-level and public so it can be
used by error actions.

Works toward resolving #719.

### DIFF
--- a/spec/lucky/error_handling_spec.cr
+++ b/spec/lucky/error_handling_spec.cr
@@ -68,21 +68,4 @@ describe Lucky::ErrorHandler do
     context.response.headers["Content-Type"].should eq("text/plain")
     context.response.status_code.should eq(418)
   end
-
-  context "when configured to show debug output" do
-    it "prints debug output instead of calling the error action" do
-      Lucky::ErrorHandler.temp_config(show_debug_output: true) do
-        fake_io = IO::Memory.new
-        error_handler = Lucky::ErrorHandler.new(action: FakeErrorAction, error_io: fake_io)
-        error_handler.next = ->(_ctx : HTTP::Server::Context) { raise UnhandledError.new }
-
-        context = error_handler.call(build_context).as(HTTP::Server::Context)
-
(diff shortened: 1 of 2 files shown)


src/lucky/error_handler.cr
```diff
@@ -13,20 +13,10 @@ class Lucky::ErrorHandler
...
-    error.inspect_with_backtrace(error_io)
-    render_exception_page(context, error)
-    context
-  end
-
-  private def render_exception_page(context, error)
+  def self.render_exception_page(context, error)
     context.response.reset
     context.response.status_code = 500
     context.response.content_type = "text/html"
```
paulcsmith: On the line right below this, instead of printing right away, can you change this to return a Lucky::TextResponse? That way it can be rendered in a Lucky::Action

https://github.com/luckyframework/lucky/blob/1a16bebd932195118d48369aa2a22bb185f36dac/src/lucky/text_response.cr

```crystal
Lucky::TextResponse.new(
  context: context,
  content_type: "text/html",
  status_code: 500,
  body: Lucky::ExceptionPage.for_runtime_exception(context, error)
)
```

spec/lucky/error_handling_spec.cr
```diff
@@ -69,20 +69,13 @@ describe Lucky::ErrorHandler do
...
-        fake_io.to_s.should contain("UnhandledError")
-        fake_io.to_s.should contain("from spec/lucky/error_handling_spec.cr")
-      end
+  describe ".render_exception_page" do
+    it "returns a exception page as a response with a 500 status" do
+      context = build_context
+      error = Exception.new
+      response = Lucky::ErrorHandler.render_exception_page(context, error)
+      response.should be_a(Lucky::Response)
+      response.status.should eq(500)
```
paulcsmith: Love it!

paulcsmith:
Thanks @HarrisonB. This is gonna be a really nice change when it is released

PR #730, opened 2019-03-16
Fixes allowing a symbol for status in a redirect.

### LINKED ISSUE #729: Unable to use symbols for status
When you set a status like in a redirect, you can currently use a Int32, or Action::Status enum, but you can't use a symbol.

```

    redirect to: "https://google.com", status: :moved_permanently
    ^~~~~~~~

in lib/lucky/src/lucky/redirectable.cr:54: no overload matches 'HTTP::Server::Response#status_code=' with type Symbol
Overloads are:
 - HTTP::Server::Response#status_code=(status_code : Int32)

                     ^~~~~~~~~~~
```

Crystal should handle this, so we are probably missing a conversion somewhere.

jwoertink:
Fixes #729 

The only relevant change here is adding a type to status. That forces the compiler to use the enum one. I moved that method override up thinking it was an order thing at first. I left it since it made sense. It calls the method below it.

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -43,6 +43,12 @@ class Rendering::JSON::WithTypedStatus < Lucky::Action
   end
 end
 
+class Rendering::JSON::WithSymbolStatus < Lucky::Action
+  get "/foo" do
+    json({name: "Paul"}, status: :created)
+  end
+end
+
 class Rendering::HeadOnly < Lucky::Action
   get "/foo" do
     head status: 204
@@ -52,6 +58,12 @@ end
 class Rendering::HeadOnly::WithTypedStatus < Lucky::Action
   get "/foo" do
     head status: Status::NoContent
+  end
+end
+
+class Rendering::HeadOnly::WithSymbolStatus < Lucky::Action
+  get "/foo" do
+    head status: :no_content
   end
 end
 
@@ -70,6 +82,12 @@ end
 class Rendering::Text::WithTypedStatus < Lucky::Action
   get "/foo" do
     text "Anything", status: Status::Created
+  end
+end
+
+class Rendering::Text::WithSymbolStatus < Lucky::Action
+  get "/foo" do
+    text "Anything", status: :created
   end
 end
 
@@ -134,6 +152,9 @@ describe Lucky::Action do
 
     status = Rendering::JSON::WithTypedStatus.new(build_context, params).call.status
     status.should eq 201
+
+    status = Rendering::JSON::WithSymbolStatus.new(build_context, params).call.status
+    status.should eq 201
   end
 
   it "renders head response with no body" do
@@ -142,6 +163,9 @@ describe Lucky::Action do
     response.status.should eq 204
 
     response = Rendering::HeadOnly::WithTypedStatus.new(build_context, params).call
+    response.status.should eq 204
+
+    response = Rendering::HeadOnly::WithSymbolStatus.new(build_context, params).call
     response.status.should eq 204
   end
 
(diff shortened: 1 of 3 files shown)


src/lucky/redirectable.cr
```diff
@@ -42,29 +42,29 @@ module Lucky::Redirectable
...
+  end
+
   # Redirect using a `String`
   #
   # ```crystal
   # redirect to: "/users"
   # redirect to: "/users/1", status: 301
   # ```
-  def redirect(to path : String, status = 302) : Lucky::TextResponse
+  def redirect(to path : String, status : Int32 = 302) : Lucky::TextResponse
```
paulcsmith: Oh I see! Nice catch. Thanks for adding the extra specs too

src/lucky/redirectable.cr
```diff
@@ -42,29 +42,29 @@ module Lucky::Redirectable
...
+  #
+  # ```crystal
+  # redirect to: "/users", status: Status::MovedPermanently
+  # ```
+  # You can find a list of all of the possible statuses [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/action.cr).
+  def redirect(to path : String, status : Lucky::Action::Status = Lucky::Action::Status::Found) : Lucky::TextResponse
+    redirect(path, status.value)
+  end
+
   # Redirect using a `String`
```
paulcsmith: While you're in here could you change this to `Redirect to the given path, with an optional `Int32` status`

src/lucky/redirectable.cr
```diff
@@ -42,29 +42,29 @@ module Lucky::Redirectable
...
   #
   # ```crystal
   # redirect to: "/users"
   # redirect to: "/users/1", status: 301
   # ```
```
paulcsmith: And maybe add a note after the example that recommends using the version with a status for readability. Something like `Note: we recommend using the method above that accepts a human friendly version of the status`

src/lucky/redirectable.cr
```diff
@@ -42,29 +42,29 @@ module Lucky::Redirectable
     redirect to: action.route, status: status
   end
 
+  # Redirect using a `String` and a `Status` value
```
paulcsmith: Maybe `Redirect to the given path, with a human friendly status`

src/lucky/redirectable.cr
```diff
@@ -42,29 +42,29 @@ module Lucky::Redirectable
...
+  #
+  # ```crystal
+  # redirect to: "/users", status: Status::MovedPermanently
+  # ```
+  # You can find a list of all of the possible statuses [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/action.cr).
+  def redirect(to path : String, status : Lucky::Action::Status = Lucky::Action::Status::Found) : Lucky::TextResponse
```
paulcsmith: I think we can get rid of the default here since the other one already has a default. What do you think? Does that still work?

spec/lucky/action_rendering_spec.cr
```diff
@@ -43,6 +43,12 @@ class Rendering::JSON::WithTypedStatus < Lucky::Action
   end
 end
 
+class Rendering::JSON::WithSymbolStatus < Lucky::Action
+  get "/foo" do
+    json({name: "Paul"}, status: :created)
```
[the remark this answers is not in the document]
jwoertink: Yeah, that works. Do you want this spec to look like that instead? Or add another one?

src/lucky/redirectable.cr
jwoertink: Yup. It still works.

src/lucky/redirectable.cr
```diff
@@ -42,22 +42,23 @@ module Lucky::Redirectable
     redirect to: action.route, status: status
   end
 
-  # Redirect using a `String` and a `Status` value
+  # Redirect to the given path, with a human friendly status
   #
   # ```crystal
   # redirect to: "/users", status: Status::MovedPermanently
```
paulcsmith: Oh and can you add an example using a symbol with a comment saying Crystal will convert it to the matching enum at compile time?

spec/lucky/action_rendering_spec.cr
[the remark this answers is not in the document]
paulcsmith: Nah it’s ok as is. Thanks for checking :)

PR #731, opened 2019-03-18
WIP: added in a new fallback route feature.

### LINKED ISSUE #728: [Feature] Adding a fallback route
This has been brought up a few times, but I'd like to actually propose something. Related to https://github.com/luckyframework/lucky_router/issues/12 I propose a `fallback` method. 

```crystal
class Fallback::Index < BrowserAction
  fallback :get do
    render(SomePage::Index)
  end
end
```

My thought is there's just another [container](https://github.com/luckyframework/lucky_router/blob/master/src/lucky_router/matcher.cr#L5) that stores the fallbacks. The fallbacks would be scoped to each HTTP method so that way you can fallback on GET and not worry about POST, etc..

a fallback is just when the router can't find anything that matches, then it looks for one of these to be defined. You would only be allowed 1 fallback per HTTP Method. This is also not to be confused with [route globbing](https://github.com/luckyframework/lucky_router/issues/3). With route globbing you may want to match something like `/posts/*dates` where dates could be `2019` ,`2019/04`, `2019/04/01`, etc... With the fallback it's just "I didn't find anything that matched, so just run this code instead".

This could happen [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/route_handler.cr#L11) which would be updated to something like:

```crystal
    if handler
      Lucky.logger.debug("Handled by #{handler.payload.to_s.colorize.bold}")
      handler.payload.new(context, handler.params).perform_action
    elsif Lucky::Router.has_fallback?(context.request)
      Lucky::Router.fallback_handler(context, handler.params)
    else
      call_next(context)
    end
```

The main usecase @russ and I have for this is working with SPA. When you want to do deep linking in to your app, Lucky throws a route not found exception. There's a few current "hacks" for this:

1. piggyback off `handle_error(error : Lucky::RouteNotFoundError)` in `Errors::Show` which would allow you to render a specific page. The downside is this puts some view rendering logic in an errors action where your app now is based around throwing an error. Also, handling when it really should be a 404 becomes tricky.

2. Define all the same routes you have in your SPA router in Lucky as well. This means duplicating routes in 2 different languages. Very prone to missing something.

jwoertink:
Fixes #728 

I labeled this as WIP since I'm sure there will be a few comments on this. Plus I'm going to implement this in to an app to feel it out a bit and see what might be missing.

This adds in the feature of a fallback route using

```crystal
  fallback do
    text "whatever here"
  end
end
```

This also removes the `Lucky::Router::INSTANCE` constant to a new private class property due to an issue with the specs where each spec would carry over the instance variables. Doing it this way allows us to reset that internally. It should be noted that this `reset!` is not intended to be a public API, and anyone else reading the code that uses it should expect nothing less than an explosion 😆

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -126,6 +126,12 @@ class Rendering::File::Missing < Lucky::Action
   end
 end
 
+class Rendering::FallbackRoute < Lucky::Action
+  fallback do
+    text "Hey, you found me!"
+  end
+end
+
 describe Lucky::Action do
   describe "rendering HTML pages" do
     it "render assigns" do
@@ -205,5 +211,11 @@ describe Lucky::Action do
     response = Rendering::File::CustomContentType.new(build_context, params).call
     response.status.should eq 200
     response.content_type.should eq "text/html"
+  end
+
+  it "renders from a fallback" do
+    response = Rendering::FallbackRoute.new(build_context, params).call
+    response.body.should eq "Hey, you found me!"
+    response.status.should eq 200
   end
 end
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -126,12 +126,6 @@ class Rendering::File::Missing < Lucky::Action
   end
 end
 
-class Rendering::FallbackRoute < Lucky::Action
-  fallback do
-    text "Hey, you found me!"
-  end
-end
-
 describe Lucky::Action do
   describe "rendering HTML pages" do
     it "render assigns" do
@@ -211,11 +205,5 @@ describe Lucky::Action do
     response = Rendering::File::CustomContentType.new(build_context, params).call
     response.status.should eq 200
     response.content_type.should eq "text/html"
-  end
-
-  it "renders from a fallback" do
-    response = Rendering::FallbackRoute.new(build_context, params).call
-    response.body.should eq "Hey, you found me!"
-    response.status.should eq 200
   end
 end
--- a/spec/lucky/route_not_found_handler_spec.cr
+++ b/spec/lucky/route_not_found_handler_spec.cr
@@ -20,12 +20,7 @@ describe Lucky::RouteNotFoundHandler do
     end
   end
 
-  it "takes an optional fallback action" do
-    context = build_context(path: "/non-existent")
-    context.request.method = "GET"
-
-    handler = Lucky::RouteNotFoundHandler.new(fallback: SampleFallbackAction::Index)
-    handler.next = ->(_ctx : HTTP::Server::Context) {}
-    handler.call(context)
+  it "has the fallback_action set from a fallback route" do
+    Lucky::RouteNotFoundHandler.fallback_action.should eq SampleFallbackAction::Index
   end
 end
--- a/spec/lucky/route_not_found_handler_spec.cr
+++ b/spec/lucky/route_not_found_handler_spec.cr
@@ -23,4 +23,29 @@ describe Lucky::RouteNotFoundHandler do
   it "has the fallback_action set from a fallback route" do
     Lucky::RouteNotFoundHandler.fallback_action.should eq SampleFallbackAction::Index
   end
+
+  it "responds with a fallback action" do
+    output = IO::Memory.new
+    context = build_context_with_io(output, path: "/non-existent")
+    context.request.method = "GET"
+    handler = Lucky::RouteNotFoundHandler.new
+    handler.next = ->(_ctx : HTTP::Server::Context) {}
+
+    handler.call(context)
+
+    context.response.close
+    output.to_s.should contain "Last chance"
+  end
+
+  it "still raises a Lucky::RouteNotFoundError for non GET requests" do
+    output = IO::Memory.new
+    context = build_context_with_io(output, path: "/non-existent")
+    context.request.method = "POST"
+    handler = Lucky::RouteNotFoundHandler.new
+    handler.next = ->(_ctx : HTTP::Server::Context) {}
+
+    expect_raises(Lucky::RouteNotFoundError) do
+      handler.call(context)
+    end
+  end
 end
--- a/spec/lucky/route_not_found_handler_spec.cr
+++ b/spec/lucky/route_not_found_handler_spec.cr
@@ -2,6 +2,12 @@ require "../spec_helper"
 
 include ContextHelper
 
+class SampleFallbackAction::Index < Lucky::Action
+  fallback do
+    text "Last chance"
+  end
+end
+
 describe Lucky::RouteNotFoundHandler do
   it "raises a Lucky::RouteNotFoundError" do
     context = build_context(path: "/foo/bar")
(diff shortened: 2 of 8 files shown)


spec/lucky/action_rendering_spec.cr
```diff
@@ -126,6 +126,12 @@ class Rendering::File::Missing < Lucky::Action
   end
 end
 
+class Rendering::FallbackRoute < Lucky::Action
+  fallback do
+    text "Hey, you found me!"
```
paulcsmith: Love it! haha

spec/lucky/router_spec.cr
```diff
@@ -1,7 +1,13 @@
 require "../../spec_helper"
 
 describe Lucky::Router do
+  # TODO: use this once there's a way to scope to just this block
+  # Spec.before_each do
+  #  Lucky::Router.reset!
+  # end
```
paulcsmith: I'm not sure how to do this, but maybe asking in the forum on Gitter would help. Would definitely be helpful to be able to use this just in a single describe block

spec/lucky/router_spec.cr
```diff
@@ -10,9 +16,18 @@ describe Lucky::Router do
...
     Lucky::Router.find_action(:head, "/test").should_not be_nil
     Lucky::Router.find_action("head", "/test").should_not be_nil
   end
+
+  it "finds a fallback route" do
+    Lucky::Router.reset!
+    Lucky::Router.add_fallback(Lucky::Action)
+
+    Lucky::Router.find_action(:get, "/test").should_not be_nil
+    Lucky::Router.find_action(:get, "/test").should be_a Lucky::FallbackRoute
```
paulcsmith: Maybe `/non-existent-route` just to make it extra clear?

spec/lucky/action_rendering_spec.cr
```diff
@@ -206,4 +212,10 @@ describe Lucky::Action do
     response.status.should eq 200
     response.content_type.should eq "text/html"
   end
+
+  it "renders from a fallback" do
+    response = Rendering::FallbackRoute.new(build_context, params).call
+    response.body.should eq "Hey, you found me!"
+    response.status.should eq 200
+  end
```
paulcsmith: What do you think about moving to a `fallback_action_spec.cr`? And then adding another spec for `it "does not generate route helpers"` and it checks that the fallback action does not `respond_to?` `route`, `with`, `url`, `path` or `url_without_query_params`

src/lucky/router.cr
```diff
@@ -1,19 +1,32 @@
 class Lucky::Router
-  INSTANCE = new
+  alias RouteMatch = LuckyRouter::Match(Lucky::Action.class) | Lucky::FallbackRoute | Nil
+  private class_property instance = new
 
   getter :routes
+  getter fallback : Lucky::FallbackRoute?
```
paulcsmith: Also I think we should pass in the Fallback **class** if possible to match what we do with regular routes

src/lucky/router.cr
```diff
@@ -1,19 +1,32 @@
 class Lucky::Router
-  INSTANCE = new
+  alias RouteMatch = LuckyRouter::Match(Lucky::Action.class) | Lucky::FallbackRoute | Nil
```
paulcsmith: Instead of having multiple types here, what if we remove `Lucky::FallbackRoute` and instead instantiate a `LuckyRouter::Match`. Something like this:

```crystal
  def find_action(method, path) : RouteMatch
     @matcher.match(method.to_s.downcase, path) || LuckyRouter::Match.new(fallback, {} of String => String)
end
```

And that or is pretty long so might be worth extracting a private def `fallback_match`

paulcsmith: I do like this alias though and think it'd be cool to keep it!

spec/lucky/router_spec.cr
jwoertink: I asked, you can't. It doesn't matter where the `before_each` is at, it'll apply to ALL `it` blocks. Adding it in causes some other specs to fail, so I'm not sold on it being in there.

src/lucky/router.cr
jwoertink: Wouldn't that require updating that shard separately?

paulcsmith: You can instantiate a `LuckyRouter::Match` by yourself since it is a public class. As long as the first method match the signature (`Lucky::Action.class` in this case) then it will return the same object. I believe what I wrote should work, but it requires the fallback to be a class not an instance (so it matches the type signature)

spec/lucky/router_spec.cr
paulcsmith: Hmm. I think that should be something added to Crystal, but yeah I think we can remove the comments and add it where it is necessary like you did in this spec 👍

spec/lucky/route_not_found_handler_spec.cr
```diff
@@ -13,4 +19,13 @@ describe Lucky::RouteNotFoundHandler do
...
     end
   end
+
+  it "takes an optional fallback action" do
+    context = build_context(path: "/non-existent")
+    context.request.method = "GET"
+
+    handler = Lucky::RouteNotFoundHandler.new(fallback: SampleFallbackAction::Index)
+    handler.next = ->(_ctx : HTTP::Server::Context) {}
+    handler.call(context)
```
jwoertink: this spec is useless at the moment... Should it even do anything?

paulcsmith: This took awhile to figure out but it needs a `context.server.response.close` to close the io. Here's some code:

```crystal
  it "takes an optional fallback action" do
    output = IO::Memory.new
    context = build_context_with_io(output, path: "/non-existent")
    context.request.method = "GET"
    handler = Lucky::RouteNotFoundHandler.new
    handler.next = ->(_ctx : HTTP::Server::Context) {}

    handler.call(context)

    context.response.close
    output.to_s.should contain "Last chance"
  end
```

I also think another spec where the method is "POST" and raises instead

jwoertink: lol. I actually had that same code, but I missed that `context.response.close` and kept getting an error of unexpected http end. Good call. I'll add those two.

jwoertink:
wtf? apparently I don't know how to git this morning. 😕

jwoertink:
@paulcsmith ok those router changes are removed from this PR now. Do you want me to send another PR to re-implement those with the reset? Or just open an issue to handle later?

paulcsmith:
Oh and did you try this out on your app and can confirm it works?

jwoertink:
I like leaving it as a constant; however, I think maybe the specs need to be a bit more robust. The fact that each spec instance carries over information from a previous spec isn't good. But I think that can be a low priority situation since it doesn't really affect an app (that I can think of)

paulcsmith:
https://github.com/luckyframework/website/issues/224

paulcsmith:
@jwoertink Yeah that makes sense. I think an issue to handle it later would be good. Making them more robust would be great.

Also what do you think about catch-all v. fallback? Example `catch_all do/end` and `CatchAllAction`? I don't have a strong opinion, but I figured I'd ask in case one sounds better than the other

jwoertink:
It does work on my app locally. This app isn't in production just yet, but it's working as intended. There's a weird issue now with the fact that lucky no longer throws a 404 on GET routes when the fallback is added. Though, I spoke with a few different react devs and they say that's sort of common. I guess with rails it's pretty common to use a glob route, and have the SPA handle the routing...

And on that note, the SPA will handle non-existent routes (provided you handle that yourself), it's just lucky returns a 200 as opposed to 404. I think that's ok though, just maybe needs to be documented?

jwoertink:
I do like `catch_all`, though I wonder if that would imply for ALL HTTP methods? as opposed to just GET? We can always try one and then rename later if people think it's confusing?

paulcsmith:
Yeah let’s just stick with fallback for now :)

And that is one of the bummers about the SPA life. Though I’d suggest something like this in the FallbackAction most of the time:

```crystal
fallback do
  if json?
    json Errors::ShowSerializer.new("Not found: #{request.path}"), status: :not_found
  else
    # do your thing
  end
end
```

That way JSON responses still get a 404. That can be documented in the guides

PR #732, opened 2019-03-19
Adding a new exception to be raised when not using a tag within a tag.

### LINKED ISSUE #86: If a string is returned inside a tag block, raise helpful error
```crystal
div do
  "my string"
end
```

The inner string will not be output because you need to use `text "my string"`

So it should raise an error like

```
Return "my_string", but was expecting a tag to be used.

If you want to display the text, do this:

    text "my_string"
```

jwoertink:
Fixes #86 

In the current version of lucky, if you do this:

```crystal
div do
  "my string"
end
```
It will render `<div></div>` which may confuse some people first starting out. This adds an exception that will give a friendly error letting you know that didn't work, and how to fix it.

### DIFF
--- a/src/lucky/tags/_tag_content_checkable.cr
+++ b/src/lucky/tags/_tag_content_checkable.cr
@@ -1,0 +1,30 @@
+module Lucky::TagContentCheckable
+  # If a tag has a nested tag (`IO`) or nil
+  # then return that content.
+  #   ```
+  #     div { }
+  #     div { div { } }
+  #     div { text "" }
+  #   ```
+  private def tag_content_check!(content : IO?)
+    content
+  end
+
+  # A tag can only have another tag or nil as a
+  # nested value. Anything else should raise a compile-time error
+  #   ```
+  #     div { "this will fail" }
+  #   ```
+  private def tag_content_check!(content)
+    {%
+      raise <<-MESSAGE
+
+      A tag in #{@type} has a nested String, but it must return a tag or `text`.
+
+      If you want to display text, try this:
+
+        text "my string"
+      MESSAGE
+    %}
+  end
+end
--- a/src/lucky/tags/_tag_content_checkable.cr
+++ b/src/lucky/tags/_tag_content_checkable.cr
@@ -1,20 +1,20 @@
 module Lucky::TagContentCheckable
   # If a tag has a nested tag (`IO`) or nil
   # then return that content.
-  #   ```
-  #     div { }
-  #     div { div { } }
-  #     div { text "" }
-  #   ```
+  # ```
+  # div { }
+  # div { div { } }
+  # div { text "" }
+  # ```
   private def tag_content_check!(content : IO?)
     content
   end
 
   # A tag can only have another tag or nil as a
   # nested value. Anything else should raise a compile-time error
-  #   ```
-  #     div { "this will fail" }
-  #   ```
+  # ```
+  # div { "this will fail" }
+  # ```
   private def tag_content_check!(content)
     {%
       raise <<-MESSAGE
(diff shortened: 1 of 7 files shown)


src/lucky/tags/custom_tags.cr
```diff
@@ -43,4 +43,10 @@ module Lucky::CustomTags
     tag_attrs = build_tag_attrs(merged_options)
     view << "<#{name}" << tag_attrs << ">"
   end
+
+  private def tag_content_check(content)
+    return if content.nil?
+    return content if typeof(content) == IO::Memory
+    raise Lucky::Exceptions::InvalidTagError.new(content.to_s)
+  end
```
paulcsmith: What if we do this at compile time instead!?

```
tag_content_check!(content : String)
  {% content.raise “the error message” %}
end

tag_content_check!(content)
  content
end
```

paulcsmith: Was on mobile. These should be methods :P

jwoertink: I like the idea of compile time. I don't think there's a way to really test it in crystal just yet. The only thing though using the method overload way is maybe catching for more than just string? like `div { 1 }` or `div { :taco }`. Does that matter? If not, then I'll start with string. I guess we can always add to it later too.

paulcsmith: I think you'd need to have some code in spec/support that has an error and then you'd have to shell out to Crystal, run it, and check that it errors and writes to STDERR as expected...but I dunno :P Haven't tried it.

You've got a point about catching stuff though. Maybe we flip it around so `content : IO?` comes first and returns the content, and `content` comes after and raises. Could you try that out on of your apps and see if it still compiles ok? If so that might be perfect because it will stop just about anything. Even stuff like this `span { user }` where someone accidentally passes the object instead of an attribute of the object.

If that doesn't work though then I think catching `String | Int32` is fine for the first pass 👍

paulcsmith: However, we should explicitly allow `:taco`. That should print a taco emoji :P

jwoertink:
Ok, this works now. Here's a screen shot:
Screen Shot 2019-03-20 at 11 21 17 AM

One caveat with this is that since it's a method being called, there's no way to get the line number at compile-time. We'd have to move it to a macro which makes things very difficult and would require some refactoring. I think this is a good start as the current implementation doesn't throw any errors, and just leaves you confused.

src/lucky/tags/_tag_content_checkable.cr
```diff
@@ -0,0 +1,32 @@
...
+
+      If you want to display text, try this:
+
+        div do
+          text "my string"
+        end
+      MESSAGE
+    %}
+  end
+end
```
paulcsmith: Love the module and the docs 👍

src/lucky/tags/base_tags.cr
```diff
@@ -42,7 +43,7 @@ module Lucky::BaseTags
       merged_options = merge_options(other_options, options)
       tag_attrs = build_tag_attrs(merged_options)
       view << "<{{tag.id}}" << tag_attrs << boolean_attrs << ">"
-      yield
+      tag_content_check!(yield)
```
paulcsmith: How about `check_tag_content!`?

src/lucky/tags/base_tags.cr
```diff
@@ -1,4 +1,5 @@
 module Lucky::BaseTags
+  include Lucky::TagContentCheckable
```
paulcsmith: Thoughts on `Lucky::CheckTagContent`?

PR #734, opened 2019-03-20
add HSTS header to ForceSSLHandler.

### LINKED ISSUE #523: Add HSTS options to ForceSSLHandler
* https://blog.bigbinary.com/2016/08/24/rails-5-adds-more-control-to-fine-tuning-ssl-usage.html
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

jwoertink:
Fixes #523 

This allows you to configure setting [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#Preloading_Strict_Transport_Security) when you have SSL enabled. By default this is on as well as including the subdomains option and a default age of 365 days. If the option is set to nil, then it's just disabled and no extra header is set.

### DIFF
--- a/spec/lucky/force_ssl_handler_spec.cr
+++ b/spec/lucky/force_ssl_handler_spec.cr
@@ -53,6 +53,24 @@ describe Lucky::ForceSSLHandler do
       context.response.status_code.should eq 200
       context.response.headers["Location"]?.should be_nil
     end
+
+    it "enables HSTS with 1 year max age and subdomains by default" do
+      context = build_context(path: "/path")
+
+      run_force_ssl_handler context
+
+      context.response.headers["Strict-Transport-Security"].should eq "max-age=31536000; includeSubDomains"
+    end
+
+    it "allows setting hsts max-age to 180 days with no subdomains" do
+      context = build_context(path: "/path")
+
+      Lucky::ForceSSLHandler.temp_config(hsts: {max_age: 180.days, include_subdomains: false}) do
+        run_force_ssl_handler context
+      end
+
+      context.response.headers["Strict-Transport-Security"].should eq "max-age=15552000"
+    end
   end
 end
 
--- a/src/lucky/force_ssl_handler.cr
+++ b/src/lucky/force_ssl_handler.cr
@@ -26,8 +26,8 @@ class Lucky::ForceSSLHandler
 
   Habitat.create do
     setting redirect_status : Int32 = Lucky::Action::Status::PermanentRedirect.value
-    setting enabled : Bool = true
-    setting hsts : Hsts? = {max_age: 365.days, include_subdomains: true}
+    setting enabled : Bool
+    setting hsts : Hsts
   end
 
   def call(context)
@@ -55,12 +55,12 @@ class Lucky::ForceSSLHandler
 
   # Read more about [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)
   private def add_hsts_if_enabled(context : HTTP::Server::Context)
-    sts_value = settings.hsts.try do |hsts|
-      String.build do |s|
+    settings.hsts.try do |hsts|
+      sts_value = String.build do |s|
         s << "max-age=#{hsts[:max_age].total_seconds.to_i}"
         s << "; includeSubDomains" if hsts[:include_subdomains]
       end
+      context.response.headers["Strict-Transport-Security"] = sts_value
     end
-    context.response.headers["Strict-Transport-Security"] = sts_value if sts_value
   end
 end
--- a/src/lucky/force_ssl_handler.cr
+++ b/src/lucky/force_ssl_handler.cr
@@ -22,11 +22,12 @@
 # ```
 class Lucky::ForceSSLHandler
   include HTTP::Handler
+  alias Hsts = NamedTuple(max_age: Time::Span, include_subdomains: Bool)
 
   Habitat.create do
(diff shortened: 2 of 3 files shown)


src/lucky/force_ssl_handler.cr
```diff
@@ -26,6 +26,7 @@ class Lucky::ForceSSLHandler
   Habitat.create do
     setting redirect_status : Int32 = Lucky::Action::Status::PermanentRedirect.value
     setting enabled : Bool = true
+    setting hsts : NamedTuple(max_age: Time::Span, include_subdomains: Bool)? = {max_age: 365.days, include_subdomains: true}
```
paulcsmith: What do you think about making a type alias for this called `Hsts`?

src/lucky/force_ssl_handler.cr
```diff
@@ -46,7 +47,18 @@ class Lucky::ForceSSLHandler
 
   private def redirect_to_secure_version(context : HTTP::Server::Context)
     context.response.status_code = settings.redirect_status
+    add_hsts_if_enabled(context)
```
paulcsmith: Love the name. Super clear and concise

src/lucky/force_ssl_handler.cr
```diff
@@ -46,7 +47,18 @@ class Lucky::ForceSSLHandler
...
     context.response.headers["Location"] =
       "https://#{context.request.host}#{context.request.resource}"
   end
+
+  private def add_hsts_if_enabled(context : HTTP::Server::Context)
+    if !!settings.hsts
```
paulcsmith: Instead of using an if and `not_nil!` what do you think about using `try`?

```crystal
settings.hsts.try do |hsts|
   # Then you can use it without `not_nil!`
end
```

src/lucky/force_ssl_handler.cr
paulcsmith: Also I love the `Time::Span` instead of an `Int` 👍

src/lucky/force_ssl_handler.cr
```diff
@@ -46,7 +47,18 @@ class Lucky::ForceSSLHandler
 
   private def redirect_to_secure_version(context : HTTP::Server::Context)
     context.response.status_code = settings.redirect_status
+    add_hsts_if_enabled(context)
     context.response.headers["Location"] =
       "https://#{context.request.host}#{context.request.resource}"
   end
+
+  private def add_hsts_if_enabled(context : HTTP::Server::Context)
```
paulcsmith: Could you link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security in a comment?

spec/lucky/force_ssl_handler_spec.cr
```diff
@@ -53,6 +53,24 @@ describe Lucky::ForceSSLHandler do
...
       context.response.headers["Location"]?.should be_nil
     end
+
+    it "enables HSTS with 1 year max age and subdomains by default" do
+      context = build_context(path: "/path")
+
+      run_force_ssl_handler context
+
+      context.response.headers["Strict-Transport-Security"].should eq "max-age=31536000; includeSubDomains"
+    end
```
paulcsmith: I think that HSTS should be enabled if ForceSSL is on, but I think it should be visible to people so it is easy to discover and configure. 

So instead of setting the default here, what if we set it to `nil` but add comments here: https://github.com/luckyframework/lucky_cli/blob/6956ffe067cf0e4033767253754139360d2c6f9e/src/web_app_skeleton/config/server.cr.ecr#L26

This would be awesome too because we could include a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security and explain a bit about it

Once this is merged I can go ahead and rework those comments to mention this

jwoertink: Since forcessl is enabled by default, the hsts should be as well. If you set it to `nil`, then it will be disabled. We can set the default config in the lucky_cli to nil and forcessl to `Env.production?` Then just adding the comments on how to configure plus updating the website should be good.

paulcsmith: ForceSSL isn't enabled by default. It is disabled and has comments on how to enable it. I think this is reasonable since a lot of hobby type sites don't have SSL. I think most frameworks take this approach (making it optional).

With that said, I think yore's right that if force ssl is enabled, it should also have the header. I do think we should document in the class documentation at the top of ForceSslHandler so people know what the defaults are when looking at the API docs (which still need to be finished :P)

src/lucky/force_ssl_handler.cr
```diff
@@ -46,7 +48,19 @@ class Lucky::ForceSSLHandler
...
+
+  # Read more about [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)
+  private def add_hsts_if_enabled(context : HTTP::Server::Context)
+    sts_value = settings.hsts.try do |hsts|
+      String.build do |s|
+        s << "max-age=#{hsts[:max_age].total_seconds.to_i}"
+        s << "; includeSubDomains" if hsts[:include_subdomains]
+      end
+    end
+    context.response.headers["Strict-Transport-Security"] = sts_value if sts_value
```
paulcsmith: I'm pretty sure this can be done inside the try block without asking to a temp variable. Might be a little less code/conditionals that way

PR #735, opened 2019-03-21
Adding in secure header modules.

### LINKED ISSUE #565: Add SecureBrowserHeadersHandler
Sets the headers listed in these issues:

* https://github.com/luckyframework/authentic/issues/11
* https://github.com/luckyframework/authentic/issues/12
* https://github.com/luckyframework/authentic/issues/13

May need customization. If so, can be done either in the initializer, or if it might be environment specific, it can use Habitat like in the [`ForceSslHandler`](https://github.com/luckyframework/lucky/blob/046b54b6147a748429107ed5dca085cc647f3c9e/src/lucky/force_ssl_handler.cr#L26-L29)

jwoertink:
Fixes #565

This adds in the concept of SecureHeaders modules. I'm starting with the FrameGuard, but if this all looks good, I'll add in a couple more to this PR. 

For these you will include the modules you want to use, and they will each require a specific method to be defined which returns their config options.

```crystal
  include Lucky::SecureHeaders::SetFrameGuard
  get "/some-action" do
    #...
  end

  def frame_guard_options
    {allow_from: "http://lucky.lvh.me:5000"}
  end
end
```

- [x] X-Frame-Options header
- [x] XSS Protection
- [x] Anti MIME sniffing

### DIFF
--- a/spec/lucky/secure_headers_spec.cr
+++ b/spec/lucky/secure_headers_spec.cr
@@ -2,27 +2,27 @@ require "../../spec_helper"
 
 include ContextHelper
 
-class FrameGuardRoutes::WithSame < Lucky::Action
+class FrameGuardRoutes::WithSameorigin < Lucky::Action
   include Lucky::SecureHeaders::SetFrameGuard
 
   get "/so_custom" do
     text "test"
   end
 
-  def frame_guard_options
-    {allow_from: "same"}
+  def frame_guard_value
+    "sameorigin"
   end
 end
 
-class FrameGuardRoutes::WithNowhere < Lucky::Action
+class FrameGuardRoutes::WithDeny < Lucky::Action
   include Lucky::SecureHeaders::SetFrameGuard
 
   get "/so_custom" do
     text "test"
   end
 
-  def frame_guard_options
-    {allow_from: "nowhere"}
+  def frame_guard_value
+    "deny"
   end
 end
 
@@ -33,26 +33,44 @@ class FrameGuardRoutes::WithURL < Lucky::Action
     text "test"
   end
 
-  def frame_guard_options
-    {allow_from: "https://tacotrucks.food"}
+  def frame_guard_value
+    "https://tacotrucks.food"
+  end
+end
+
+class FrameGuardRoutes::WithBadValue < Lucky::Action
+  include Lucky::SecureHeaders::SetFrameGuard
+
+  get "/so_custom" do
+    text "test"
+  end
+
+  def frame_guard_value
+    "hax0rz"
   end
 end
 
 describe Lucky::SecureHeaders do
   describe "SetFrameGuard" do
     it "sets the X-Frame-Options header with sameorigin" do
-      route = FrameGuardRoutes::WithSame.new(build_context, params).call
+      route = FrameGuardRoutes::WithSameorigin.new(build_context, params).call
       route.context.response.headers["X-Frame-Options"].should eq "sameorigin"
     end
 
     it "sets the X-Frame-Options header with deny" do
-      route = FrameGuardRoutes::WithNowhere.new(build_context, params).call
+      route = FrameGuardRoutes::WithDeny.new(build_context, params).call
       route.context.response.headers["X-Frame-Options"].should eq "deny"
     end
 
     it "sets the X-Frame-Options header to allow from tacotrucks" do
       route = FrameGuardRoutes::WithURL.new(build_context, params).call
       route.context.response.headers["X-Frame-Options"].should eq "allow-from https://tacotrucks.food"
+    end
+
+    it "throws an error when given a bad value" do
+      expect_raises(Exception) do
+        route = FrameGuardRoutes::WithBadValue.new(build_context, params).call
+      end
     end
   end
 end
--- a/spec/lucky/secure_headers_spec.cr
+++ b/spec/lucky/secure_headers_spec.cr
@@ -50,6 +50,14 @@ class FrameGuardRoutes::WithBadValue < Lucky::Action
   end
 end
 
+class XSSGuardRoutes::Index < Lucky::Action
+  include Lucky::SecureHeaders::SetXSSGuard
+
+  get "/so_custom" do
+    text "test"
+  end
+end
+
 describe Lucky::SecureHeaders do
   describe "SetFrameGuard" do
     it "sets the X-Frame-Options header with sameorigin" do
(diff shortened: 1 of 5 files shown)


src/lucky/secure_headers/set_frame_guard.cr
```diff
@@ -0,0 +1,49 @@
...
+      abstract def frame_guard_options : NamedTuple(allow_from: String)
+
+      private def set_frame_guard_header
+        context.response.headers["X-Frame-Options"] = frame_guard_value(frame_guard_options)
+        continue
+      end
+
+      private def frame_guard_value(options)
+        case options[:allow_from].downcase
+        when "same", "sameorigin"
```
edwardloveall: How do you feel about removing this case statement and just passing the value through? Is there a danger to setting `same` or `nowhere`? Or maybe there are conventions I'm missing. I could also see going the other way of throwing an error if the values are anything but `sameorigin`, `deny`, or a valid URL. What do you think?

jwoertink: Well, there's only 3 possible options as far as browsers go, so it has to be `sameorigin`, `deny`, or a valid URL. I liked the "nowhere" that helmet is using because it reads nice `allow_from: "nowhere"`; however, that's not necessary, and does add a little extra cruft. I'm down for just passing in the value and raising an exception when it's not one of those 3. If you're cool with that, I'll make that change

paulcsmith: I love the idea of raising if it doesn't match those 3!

jwoertink: Ok, made that change. If y'all are cool with that, then I'll implement the other two 😄

spec/lucky/secure_headers_spec.cr
```diff
@@ -0,0 +1,76 @@
...
+      route = FrameGuardRoutes::WithSameorigin.new(build_context, params).call
+      route.context.response.headers["X-Frame-Options"].should eq "sameorigin"
+    end
+
+    it "sets the X-Frame-Options header with deny" do
+      route = FrameGuardRoutes::WithDeny.new(build_context, params).call
+      route.context.response.headers["X-Frame-Options"].should eq "deny"
+    end
+
+    it "sets the X-Frame-Options header to allow from tacotrucks" do
```
paulcsmith: Taco trucks are always allowed. Good call

spec/lucky/secure_headers_spec.cr
```diff
@@ -0,0 +1,76 @@
...
+      route = FrameGuardRoutes::WithURL.new(build_context, params).call
+      route.context.response.headers["X-Frame-Options"].should eq "allow-from https://tacotrucks.food"
+    end
+
+    it "throws an error when given a bad value" do
+      expect_raises(Exception) do
```
paulcsmith: It might be good to check the message text as well. I believe that is the second argument. You can check just a bit of the text, no need to do the whole message. Maybe `"You set frame_guard_value to hax0rz"`

src/lucky/secure_headers/set_frame_guard.cr
```diff
@@ -0,0 +1,56 @@
...
+    #     {allow_from: "deny"}
+    #   end
+    # end
+    # ```
+    #
+    # ### Options
+    # The `frame_guard_options` method must be defined and return `NamedTuple(allow_from: String)`
+    # The `allow_from` key can have one of 3 String values:
+    # - `"same"` or `"sameorigin"`
+    # - `"nowhere" or `"deny"`
```
paulcsmith: I believe these lines need to be updated

src/lucky/secure_headers/set_frame_guard.cr
```diff
@@ -0,0 +1,56 @@
...
+          v
+        when /^https?:\/\/[0-9a-zA-Z_-]+\.?[0-9a-zA-Z_-].+$/
+          "allow-from #{v}"
+        else
+          raise <<-MESSAGE
+
+          You set frame_guard_value to #{value}, but it must be one of these options:
+            - "sameorigin"
+            - "deny"
+            - A valid URL
```
paulcsmith: Love it 👍 Thoughts on a newline before the start of the list though?

src/lucky/secure_headers/set_frame_guard.cr
```diff
@@ -0,0 +1,56 @@
...
+        context.response.headers["X-Frame-Options"] = check_frame_guard_value!(frame_guard_value)
+        continue
+      end
+
+      private def check_frame_guard_value!(value : String)
+        v = value.downcase
+        case v
+        when "sameorigin", "deny"
+          v
+        when /^https?:\/\/[0-9a-zA-Z_-]+\.?[0-9a-zA-Z_-].+$/
```
paulcsmith: I am nervous that this Regex might be too strict. I don't really know, but it could bite us in the future by disallowing a valid URL. What do you think about instead checking if the first characters start with `http` and leave whether it is valid up. to the user. This will still catch the most common errors (misspelled sameorigin, or some total incorrect value), but still let people use whatever URL they want. 

Thoughts?

spec/lucky/secure_headers_spec.cr
jwoertink: oh, I didn't know you can pass a partial message.

src/lucky/secure_headers/set_frame_guard.cr
jwoertink: yup. Good call.

src/lucky/secure_headers/set_frame_guard.cr
jwoertink: Makes sense. If it's close enough it should be fine.

src/lucky/secure_headers/set_xss_guard.cr
```diff
@@ -0,0 +1,33 @@
+module Lucky
+  module SecureHeaders
+    # This module sets the HTTP header [X-XSS-Protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection).
+    # It's job is responsible for telling the browser to not render a page if it detects cross-site scripting.
+    # For Internet Explorer browsers less than 9, it's recommended to disabled it. Read more on [Microsoft](https://blogs.msdn.microsoft.com/ieinternals/2011/01/31/controlling-the-xss-filter/).
```
edwardloveall: I think it might be good to tell people that lucky implements this recommendation so they don't think they have to write code to do so. What do you think?

jwoertink:
Isn’t that implied though? Like, the reason I use lucky in general is so I don’t have to write the code 🤔 lol

src/lucky/secure_headers/set_xss_guard.cr
paulcsmith: I think what Edward meant was that this comment makes it sound like you have to disable. It seems like Lucky isn't doing it for you. 

So maybe something like:

`Lucky disables this header for Internet Explorer 9 because it is recommended to do so. Read more on...`

edwardloveall: Ah yes, sorry about that. Poor communication on my part, thanks @paulcsmith. It does seem like the framework is instructing developers to opt out of this behavior in IE9 and below, when in fact the framework is doing that work for the developer.

jwoertink: ah! Gotcha. Ok, that makes sense. My bad 😆

PR #738, opened 2019-03-25
Adds a new special space tag to generate &nbsp; entity.

### LINKED ISSUE #669: html whitespace control
Whitespace in html is a subtle beast. Related to #668, there have been a few times when I want to add a space between tags, but because lucky html is method driven there is no easy way to do this. In regular html, a space, newline, or any other whitespace character would work. 

This snippet is an example:

```crystal
para do
  text <<-TEXT
  ICMP Echo is a low level network protocol which checks to see a host is responsive
  and measures the time of that response. It is equivalent to the 
  TEXT
  a "ping", href: "https://en.wikipedia.org/wiki/Ping_%28networking_utility%29"
  text " utility."
end
```

The rendering I'm after should look something like this markdown-formatted paragraph:

> ICMP Echo is a low level network protocol which checks to see a host is responsive and measures the time of that response. It is equivalent to the [ping](https://en.wikipedia.org/wiki/Ping_%28networking_utility%29) utility.

Specifically the embedded link in the paragraph, where I want this spacing: `[...] equivalent to the <a ...>ping</a> utility.`

In order to accomplish that, I've had to add a dangling space at the end of the heredoc as well as leading space on the second `text` method (it's there, but you can't see it rendered in this comment). The dangling space is hard to see, even with `:set list`, and a lot of people have their editors configured to trim them automatically...but I can't add a comment to prevent the trim, because it's a heredoc :complainy-complainerson-mc-complainerface:

I've run into similar issues with [other whitespace-trimming syntax dsls](https://github.com/jeromegn/slang/issues/29), and [I know that](http://haml.info/docs/yardoc/file.REFERENCE.html#tilde) many [other rendering engines](https://api.rubyonrails.org/classes/ActionView/Base.html) have dealt with this problem 🍌 

It feels brittle to add spaces like that -- Is there currently some way around this that doesn't make things so brittle? 🌮

jwoertink:
Fixes #669 

This adds a new method `space` that generates a `&nbsp;` HTML entity character. This method takes an optional argument for the number of spaces you'd like to render. 

It can be used like:

```crystal
link "Home", to: ""
span do
  space
  text "|"
  space
end
link "About", to: ""
# <a href="">Home</a><span>&nbsp;|&nbsp;</span><a href="">About</a>
```

```crystal
text "HALP!"
space(4)
text "I'm out of tacos"
# HALP!&nbsp;&nbsp;&nbsp;&nbsp;I'm out of tacos"
```

### DIFF
--- a/src/lucky/tags/specialty_tags.cr
+++ b/src/lucky/tags/specialty_tags.cr
@@ -58,19 +58,19 @@ module Lucky::SpecialtyTags
   end
 
   # Generates an escaped HTML `&nbsp;` entity for the number of times specified
-  # by `how_many`. By default it generates 1 space character.
+  # by `how_many`. By default it generates 1 non-breaking space character.
   #
   # ```
   # link "Home", to: Home::Index
   # span do
-  #   space
+  #   nbsp
   #   text "|"
-  #   space
+  #   nbsp
   # end
   # link "About", to: About::Index
   # ```
   # Would generate `<a href="/">Home</a><span>&nbsp;|&nbsp;</span><a href="/about">About</a>`
-  def space(how_many : Int32 = 1)
+  def nbsp(how_many : Int32 = 1)
     how_many.times { raw("&nbsp;") }
     view
   end
--- a/src/lucky/tags/specialty_tags.cr
+++ b/src/lucky/tags/specialty_tags.cr
@@ -56,4 +56,22 @@ module Lucky::SpecialtyTags
   def raw(string : String)
     view << string
   end
+
+  # Generates an escaped HTML `&nbsp;` entity for the number of times specified
+  # by `how_many`. By default it generates 1 space character.
+  #
+  # ```
+  # link "Home", to: Home::Index
+  # span do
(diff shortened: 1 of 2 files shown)


src/lucky/tags/specialty_tags.cr
```diff
@@ -56,4 +56,22 @@ module Lucky::SpecialtyTags
...
+  # link "Home", to: Home::Index
+  # span do
+  #   space
+  #   text "|"
+  #   space
+  # end
+  # link "About", to: About::Index
+  # ```
+  # Would generate `<a href="/">Home</a><span>&nbsp;|&nbsp;</span><a href="/about">About</a>`
+  def space(how_many : Int32 = 1)
```
paulcsmith: Nice! I think the only change I think would be helpful is calling this method `nbsp`. That way it is clear it is a non breaking space and not just a space string or something

jwoertink: but then my branch name won't make sense 😝

PR #741, opened 2019-03-26
added in config for asset building retry time. Defaults to 5seconds.

### LINKED ISSUE #717: Allow customizing manifest retry timeout
https://gitter.im/luckyframework/Lobby?at=5c7d7df053efa91203b20816

jwoertink:
Fixes #717 

This allows you to customize the length of time the asset building will wait which by default is 5 seconds. In the case of asset building really large projects, it may take up to 20 seconds (like in my case) to build out the assets. When this happens, there's a race condition that causes a missing asset error. To currently circumvent this, I have to run `yarn dev` and `crystal src/server.cr` in 2 separate terminal tabs instead of just running `lucky dev`. 

I tried this out locally, and it *seems* to work... Though, I'm not sure how to write a spec to really test it. Also, since it's ran through a `run` macro call, doing a `puts` style debugging doesn't seem to work. 🤷‍♂️

### DIFF
--- a/src/run_macros/generate_asset_helpers.cr
+++ b/src/run_macros/generate_asset_helpers.cr
@@ -1,13 +1,17 @@
 require "json"
+require "yaml"
 require "colorize"
 
 class AssetManifestBuilder
-  MANIFEST_PATH = File.expand_path("./public/mix-manifest.json")
-  MAX_RETRIES   =   20
-  RETRY_AFTER   = 0.25
+  MANIFEST_PATH       = File.expand_path("./public/mix-manifest.json")
+  WATCH_PATH          = File.expand_path("./config/watch.yml")
+  MAX_RETRIES         =   20
+  DEFAULT_RETRY_AFTER = 0.25
 
   property retries
   @retries : Int32 = 0
+
+  @retry_interval : Float64? = nil
 
   def self.build_with_retry
     new.build_with_retry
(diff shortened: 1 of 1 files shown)


jwoertink:
Ok, this isn't done. I just tested it by setting it to 90 seconds which is way more than enough and it failed. I think I see why, but it's going to require a bit more of a refactor on how this whole things works. 

How it works now is [this line](https://github.com/luckyframework/lucky/blob/master/src/run_macros/generate_asset_helpers.cr#L17) returns true after the js and images are done because they finish generating by the time it gets there. The CSS (in my case) hasn't finished, so it just goes on.

What I'm thinking here is something more like:

```crystal
private def check_wait_time
  if retries < MAX_RETRIES
    self.retries += 1
    sleep(retry_interval)
    check_wait_time
  end
  build_or_raise
end

def build_or_raise
  if manifest_exists?
      build
    else
      raise_error
  end
end
```

The plus side is being able to control the waiting time. The downside is a few things...
1. You may forget where that's configured
2. If build time increases, you could run in to the same issue again (though fixable now)
3. If build time decreases, you may forget and just have wasted time waiting on a build.

It's definitely not optimal, and this is an edge-case related, but I'm not sure a better way to fix it since the build happens on the node side.

jwoertink:
Just thinking out loud here... What if the js build system just notified crystal when it was done, and crystal then booted at that point instead of "are you done yet? are you done yet? are you done yet?". [mix](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/event-hooks.md) has callbacks as well as [gulp](https://gulpjs.com/docs/en/getting-started/async-completion). So probably pretty common to say "I'm done building". Maybe it's as simple as...

```javascript
mix.build_stuff.then(function () { exec('lucky boot') })
```

paulcsmith:
@jwoertink Could be done, but time to get a working server would be slower since it'd have to wait on assets before it even attempts compilation. Also relying on mix or some other asset manager adds some coupling that would be unexpected (someone removes Webpack to use Gulp and wonders why the server won't even compile)

I believe the way to do this is to do "on-demand" asset checking. So instead of calling `AssetHelper.load_manifest` it would load the manifest when it couldn't find one (so first run it would always load the manifest). It would also reload the manifest if it wasn't in there (in case a new build hasn't finished). The idea would be that we'd do the retry at the macro level and not in the load manifest level.

This is all probably way poorly explained. I've experimented with it a bit I just need to do it. Been trying to get the new website (and new Lucky version) wrapped up though

jwoertink:
That makes sense. I don't really like the whole coupling thing. Especially with an API not using a manifest anyway. But this PR is currently garbage since it doesn't actually work how I need. I think I'm gonna close this PR, and we can revisit it later. The issue is still open, so we can keep track of methods to solve this

Issue #593, opened 2018-08-30
Make it easier for people to contribute


paulcsmith:
Inspired by @robacarp's post: 
https://gitter.im/luckyframework/Lobby?at=5b8751609c71d363c14b2bba

### Links

* https://solnic.codes/2018/08/29/a-quick-recap-and-plans-for-the-future/
* http://jessenoller.com/blog/2015/9/27/a-lot-happens

### Goals

* Make it easier to share the workload
* Make it easier for people to get started in open source
* Make it clear why some contributions won't be accepted, or why it will take longer to review
* Automate common steps (like new releases) so it is easier for myself or anyone else to keep things moving forward
* Help build a community that communicates clearly, humbly, and respectfully
* Help people feel comfortable making suggestions, and responding to suggestions kindly.

### For all Lucky libraries

- [x] Add `scripts/setup` scripts that make it easy to get started
- [x] Add `scripts/test` scripts that run any necessary tests (`crystal spec`, `shards build`, check formatting)
- [x] Start asking for inline docs. Add `PULL_REQUEST.md` template that guides users to write docs, run `bin/test`, check formatting.
- [ ] Add `bin/prep_release` scripts to all Lucky libraries. This would make it easier to bump versions by updating shard.yml, version.cr, anything in the README and then tag it with the correct version number
- [x] Me: Start adding more details when opening PRs and issue that talk about *why* I'm doing things

### Improve the trickier parts of Lucky

- [ ] Lucky CLI templates are very confusing right now. This may required improving teeplate to support conditional files/folders. Right now we have folders all over because that is not supported. It makes it hard to tell what a project will actually look like
- [x] LuckyRecord see if we can remove or extract some of the macros to nicely named methods
- [ ] LuckyRecord form code is pretty complex in some spots. Needs to be simplified

 More to come

### Not sure where...

- [ ] Talk about why certain contributions take longer to review
- [ ] Talk about why certain features (Windows support, MySQL/sqlite support) are not done and likely won't be done for awhile

More to come :)

ajwann:
@paulcsmith I would like to create the ```test``` scripts. Is a bin directory with an executable test file that runs those three commands all that you are looking for? Is it safe to assume that all lucky libraries would need to have those three commands run?

paulcsmith:
@ajwann that would be wonderful. I don’t think all of them need to run shards build but it wouldn’t hurt. 

And yes an executable that does that would be great!

ajwann:
[...]

PR #742, opened 2019-03-26
Create PULL_REQUEST_TEMPLATE

jwoertink:
## Purpose
This adds in a new PULL_REQUEST_TEMPLATE which should help people with contributing. Also helps to close out #593

## Description
A pull request template lets us standardize the PR process and helps newcomers to check. It also helps me remember to run the formatter because I forget EVERY time.

## Checklist
* [x] - All specs pass locally with `crystal spec/`
* [x] - All specs are formatted with `crystal tool format spec/`
* [x] - All code is formatted with `crystal tool format src/`
* [x] - Lucky builds on docker with `./scripts/setup`
* [x] - All builds and specs pass on docker with `./scripts/test`

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -8,7 +8,6 @@ Link to specific lines of code, or examples if you need to.
 
 ## Checklist
 * [ ] - An issue already exists detailing the issue/or feature request that this PR fixes
-* [ ] - All specs pass locally with `crystal spec/`
 * [ ] - All specs are formatted with `crystal tool format spec src`
 * [ ] - Inline documentation has been added and/or updated
 * [ ] - Lucky builds on docker with `./scripts/setup`
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,16 @@
+## Purpose
+Describe the feature or issue and link to the related issue. 
+If no issue has been opened about this, be sure to open an issue first to discuss the need for this PR.
+
+## Description
+Please include any relevant code samples or screen shots that may help to overview of this PR.
+Link to specific lines of code, or examples if you need to.
+
+## Checklist
+* [ ] - All specs pass locally with `crystal spec/`
+* [ ] - All specs are formatted with `crystal tool format spec/`
+* [ ] - All code is formatted with `crystal tool format src/`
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+Link to specific lines of code, or examples if you need to.
+
+## Checklist
+* [ ] - All specs pass locally with `crystal spec/`
+* [ ] - All specs are formatted with `crystal tool format spec/`
+* [ ] - All code is formatted with `crystal tool format src/`
+* [ ] - Inline documentation has been added and/or updated
+* [ ] - An issue already exists detailing the issue/or feature request that this PR fixes
+* [ ] - Lucky builds on docker with `./scripts/setup`
+* [ ] - All builds and specs pass on docker with `./scripts/test`
```
paulcsmith: Rather than have this + `crystal spec` could you leave just this one? Since if it doesn't pass then it won't pass on CI. What do you think?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+## Purpose
+Describe the feature or issue and link to the related issue. 
+If no issue has been opened about this, be sure to open an issue first to discuss the need for this PR.
+
+## Description
+Please include any relevant code samples or screen shots that may help to overview of this PR.
+Link to specific lines of code, or examples if you need to.
```
paulcsmith: I love this and the purpose :+1:

PULL_REQUEST_TEMPLATE.md
jwoertink: I just assumed if local is macOS and docker, then you're guaranteed to have ran it on 2 OS. Maybe that doesn't matter in this case?

paulcsmith: Oh I see. I think doing just `script/test` should cover it. For this project it is fairly simple to set up but for other Lucky projects like Avram it is a bit trickier. So asking people to set up locally and with Docker seems like too much. Especially since in the end someone could be developing on Linux anyway, in which case we still wouldn't know if it works on macOS. So far I haven't seen any Linux <-> Mac issues though so I think it is ok to stick with the Docker build

jwoertink: fair enough!

PR #748, opened 2019-04-06
Strict Transport Security changes

paulcsmith:

### DIFF
--- a/spec/lucky/force_ssl_handler_spec.cr
+++ b/spec/lucky/force_ssl_handler_spec.cr
@@ -54,23 +54,25 @@ describe Lucky::ForceSSLHandler do
       context.response.headers["Location"]?.should be_nil
     end
 
-    it "enables HSTS with 1 year max age and subdomains by default" do
+    it "allows setting Strict-Transport-Security" do
       context = build_context(path: "/path")
 
-      run_force_ssl_handler context
-
-      context.response.headers["Strict-Transport-Security"].should eq "max-age=31536000; includeSubDomains"
-    end
-
-    it "allows setting hsts max-age to 180 days with no subdomains" do
-      context = build_context(path: "/path")
-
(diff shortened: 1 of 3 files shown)


spec/lucky/force_ssl_handler_spec.cr
```diff
@@ -54,26 +54,28 @@ describe Lucky::ForceSSLHandler do
       context.response.headers["Location"]?.should be_nil
     end
 
-    it "enables HSTS with 1 year max age and subdomains by default" do
+    it "allows setting Strict-Transport-Security" do
```
paulcsmith: I think using the term `Strict-Transport-Security` makes this easier to understand/google. `hsts` looks odd when configuring the handler

spec/spec_helper.cr
```diff
@@ -41,7 +41,6 @@ end
 
 Lucky::ForceSSLHandler.configure do |settings|
   settings.enabled = true
-  settings.hsts = {max_age: 365.days, include_subdomains: true}
```
paulcsmith: Leave off by default to make sure it really is optional

src/lucky/force_ssl_handler.cr
```diff
@@ -12,25 +12,24 @@
...
 # Lucky::ForceSSLHandler.configure do |settings|
 #   settings.redirect_status = 303
 #   settings.enabled = false
-#   settings.hsts = {max_age: 18.weeks, include_subdomains: false}
+#   settings.strict_transport_security = {max_age: 1.year, include_subdomains: true}
 # end
 # ```
 class Lucky::ForceSSLHandler
   include HTTP::Handler
-  alias Hsts = NamedTuple(max_age: Time::Span, include_subdomains: Bool)
```
paulcsmith: The alias was nice, but it made it hard to tell what the setting expected when configuring it. It just says `Hsts`. I was hoping Crystal would also expand the alias and tell you what the NamedTuple was but it doesn't :(

PR #760, opened 2019-04-11
Added some logging when the fallback route picks up a request.

### LINKED ISSUE #758: fallback route needs some logging
I had an issue where I had these duplicate SQL calls, and couldn't figure out where it was coming from. Turned out that since I didn't have a favicon.ico file, lucky used the fallback route which in my case says render the home page for some SPA stuff, this is where the second set of queries came from. Adding a favicon fixed that.

Now when it logged
> GET /favicon.ico

There wasn't really anything that told me it was going to use the fallback routing. This would be a good place to do 

```
> GET /favicon.ico
> Route /favicon.ico not found, using fallback route
```

Or something along those lines.

jwoertink:
## Purpose
fixes #758 

This PR adds in a single line of logging to the fallback action. This lines up with the already existing log for normal actions.

## Description
If you use the `fallback` action, when reading your log output, there's not currently a way to tell if an action was caught by the fallback. In my case, the `/favicon.ico` call was caught by the fallback, but I had no way to tell. This will output

```
web | GET /favicon.ico
web | Handled by fallback in SPA::Index
```

## Checklist

I can add some docs to this handler, but since it's mostly private methods, not sure if that matters in this case..

### DIFF
--- a/src/lucky/route_not_found_handler.cr
+++ b/src/lucky/route_not_found_handler.cr
@@ -1,6 +1,6 @@
 # This HTTP::Handler takes in the current `context`,
 # then checks to see if a `fallback_action` has been defined to render that action first.
-# If no fallback has been defined, then raise a `Lucky::RouteNotFoundError` exception.
+# If no fallback has been defined, then it will raise a `Lucky::RouteNotFoundError` exception.
 #
 # This handler should be used after the `Lucky::RouteHandler`.
 #
@@ -11,7 +11,7 @@ class Lucky::RouteNotFoundHandler
 
   def call(context)
     if has_fallback?(context)
-      Lucky.logger.debug("Handled by fallback in #{fallback_action.name.to_s.colorize.bold}")
+      Lucky.logger.debug(handled_by_fallback: fallback_action.name.to_s)
       fallback_action.new(context, {} of String => String).perform_action
     else
       raise Lucky::RouteNotFoundError.new(context)
--- a/src/lucky/route_not_found_handler.cr
+++ b/src/lucky/route_not_found_handler.cr
@@ -4,6 +4,7 @@ class Lucky::RouteNotFoundHandler
 
   def call(context)
     if has_fallback?(context)
+      Lucky.logger.debug("Handled by fallback in #{fallback_action.name.to_s.colorize.bold}")
       fallback_action.new(context, {} of String => String).perform_action
     else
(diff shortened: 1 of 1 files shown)


paulcsmith:
I need to review later, but Lucky should have a static file for this by default https://github.com/luckyframework/lucky_cli/blob/master/src/browser_app_skeleton/public/favicon.ico

Did that not get used? Or maybe this was a project created after that was added?

jwoertink:
We're dealing multitenancy, so each site has it's own favicon that has to be loaded. That just wasn't hooked up yet.

paulcsmith:
Ahh that makes sense 👍

jwoertink:
Ok, I added some documentation to the handler. Might as well start filling in docs where I can.

src/lucky/route_not_found_handler.cr
```diff
@@ -1,9 +1,17 @@
+# This HTTP::Handler takes in the current `context`,
+# then checks to see if a `fallback_action` has been defined to render that action first.
+# If no fallback has been defined, then raise a `Lucky::RouteNotFoundError` exception.
```
paulcsmith: "then raise" -> "then it will raise"

src/lucky/route_not_found_handler.cr
```diff
@@ -1,9 +1,17 @@
...
+# This handler should be used after the `Lucky::RouteHandler`.
+#
+# See `Lucky::Routeable.fallback` for implementing the `fallback_action`.
 class Lucky::RouteNotFoundHandler
   include HTTP::Handler
   class_property fallback_action : Lucky::Action.class | Nil
 
   def call(context)
     if has_fallback?(context)
+      Lucky.logger.debug("Handled by fallback in #{fallback_action.name.to_s.colorize.bold}")
```
paulcsmith: This will automatically be bolded, etc when passed in as data. That way it can be formatted as JSON, pretty formatted, etc.

```
Lucky.logger.debug(handled_by_fallback: fallback_action.name.to_s)
```

jwoertink: I actually modeled this after [this line](https://github.com/luckyframework/lucky/blob/master/src/lucky/route_handler.cr#L9) which does the same. Should that one be changed too then? (separate PR)

paulcsmith: Ah yes it should also change!

PR #761, opened 2019-04-15
refactor CookieOverflow

skojin:
Check cookie size when set it, not in middleware

## Purpose
Instead of check cookie size at end of request in middleware check it when set it.

## Description

Two props:
1. more simple code
2. error stack trace will show place where bad cookie was set

## Extra 

Rails do in same way, check when cookie is set. One difference rails check size only of encrypted/signed cookies. Plain cookies size not checked.

In this commit I still check size of all cookies, encrypted and raw.
Possible we not need to check size of raw cookies.


paulcsmith:
Thank you so much for handling this! I'll merge for now and see if I can test it out and make sure it all works in production apps. It can be tricky to handle edge cases in cookie :P

@jwoertink would you mind giving this a try one some of your projects to make sure cookies are still working correctly?

PR #764, opened 2019-04-16
Upgrade Crystal 0.28.0

bcardiff:
I found the following changes needed to upgrade to the upcoming Crystal 0.28.0.

If some macro `compare_versions` are added it should be possible to support current versions as well.

Feel free to extract these changes and close the PR if you prefer to handle the migration in a specific way.

### DIFF
--- a/src/lucky/params.cr
+++ b/src/lucky/params.cr
@@ -367,8 +367,9 @@ class Lucky::Params
     multipart_params = {} of String => String
     multipart_files = {} of String => Lucky::UploadedFile
     body_io = IO::Memory.new(body)
-    boundary =
-      HTTP::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
+
+    boundary = MIME::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
+
     HTTP::FormData.parse(body_io, boundary.to_s) do |part|
       case part.headers
       when .includes_word?("Content-Disposition", "filename")
--- a/src/lucky/support/message_verifier.cr
+++ b/src/lucky/support/message_verifier.cr
@@ -47,7 +47,7 @@ module Lucky
     end
 
     private def generate_digest(data) : String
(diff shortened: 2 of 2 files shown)


bcardiff:
Note that db related shards are also in the process of been updated for Crystal 0.28.0.

src/lucky/params.cr
```diff
@@ -367,8 +367,9 @@ class Lucky::Params
     multipart_params = {} of String => String
     multipart_files = {} of String => Lucky::UploadedFile
     body_io = IO::Memory.new(body)
-    boundary =
-      HTTP::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
+
+    boundary = MIME::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
```
jwoertink: Does this need to be required? `require "mime"`

bcardiff: It could. `MIME` ended up been available because it is required in the `StaticFileHandler` of the stdlib.

jwoertink: hmm interesting. Yeah, the `Lucky::StaticFileHandlers` inherits from the `HTTP::StaticFileHandler` which does require mime already. Not sure why the CI said mime is undefined...

bcardiff:
Oh, that’s because 0.28 hasn’t been released yet :-)

jwoertink:
hahah 😅 my bad.

paulcsmith:
Closing in favor of https://github.com/luckyframework/lucky/pull/766 which updates the Dockerfile so it passes in CI. Thanks so much for this @bcardiff. It's so helpful that you help us get ready for new releases ❤️

bcardiff:
FYI the crystal-pg/mysql/sqlite3 shards are already released with 0.28.0 support

paulcsmith:
Wonderful :D Thanks for the update

PR #769, opened 2019-04-24
removed Lucky::Action::Status in favor of new built in HTTP::Status enum

### LINKED ISSUE #747: Update Lucky to use new HTTP::Status
Crystal just had [this commit](https://github.com/crystal-lang/crystal/commit/b42631e9b3c62621f42832755427d530c6e452d2) which adds in an `HTTP::Status` enum. Lucky [has this defined](https://github.com/luckyframework/lucky/blob/master/src/lucky/action.cr#L20), but it's missing my favorite status 418. I think we should just use the built in crystal one after crystal 0.28.0 is released.

The cool thing too, is the next version of crystal will give us compile time deprecation notices. I haven't tried this, but I think we could do something like:

```crystal
@[Deprecated("Lucky::Action::Status deprecated, please use HTTP::Status instead")]
alias Status = HTTP::Status
```

jwoertink:
## Purpose
fixes #747 

## Description
This PR removes the `Lucky::Action::Status` in favor of the built in `HTTP::Status` since the built in includes more statuses, and will be easier to maintain plus consistency. 

Pros:
* Consistent with other frameworks that will use this
* Less code in Lucky
* Maintenance moved to language
* This enum includes statuses missing from Lucky::Action
* All symbol style calls still work as normal
* Less typing when used outside of action

Cons:
* Major breaking change
* Requires crystal 0.28+ to work
* No slow deprecation release really possible with how enums work
* More typing if used in an action (HTTP::Status::CREATED vs Status::Created)

## Checklist

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -39,7 +39,7 @@ end
 
 class Rendering::JSON::WithTypedStatus < Lucky::Action
   get "/foo" do
-    json({name: "Paul"}, status: Status::Created)
+    json({name: "Paul"}, status: HTTP::Status::CREATED)
   end
 end
 
@@ -57,7 +57,7 @@ end
 
 class Rendering::HeadOnly::WithTypedStatus < Lucky::Action
   get "/foo" do
-    head status: Status::NoContent
+    head status: HTTP::Status::NO_CONTENT
   end
 end
 
@@ -81,7 +81,7 @@ end
 
 class Rendering::Text::WithTypedStatus < Lucky::Action
   get "/foo" do
-    text "Anything", status: Status::Created
+    text "Anything", status: HTTP::Status::CREATED
   end
 end
 
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -37,12 +37,6 @@ class Rendering::JSON::WithStatus < Lucky::Action
   end
 end
 
-class Rendering::JSON::WithTypedStatus < Lucky::Action
-  get "/foo" do
-    json({name: "Paul"}, status: HTTP::Status::CREATED)
-  end
-end
-
 class Rendering::JSON::WithSymbolStatus < Lucky::Action
   get "/foo" do
     json({name: "Paul"}, status: :created)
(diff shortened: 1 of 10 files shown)


spec/lucky/action_rendering_spec.cr
```diff
@@ -81,7 +81,7 @@ end
 
 class Rendering::Text::WithTypedStatus < Lucky::Action
   get "/foo" do
-    text "Anything", status: Status::Created
+    text "Anything", status: HTTP::Status::CREATED
```
paulcsmith: Could you switch these all to symbols? I think that is the way we should use these from now on both in the source and in the guides

src/lucky/action.cr
```diff
@@ -17,65 +17,6 @@ abstract class Lucky::Action
...
-    ServiceUnavailable            = 503
-    GatewayTimeout                = 504
-    HTTPVersionNotSupported       = 505
-    VariantAlsoNegotiates         = 506
-    InsufficientStorage           = 507
-    LoopDetected                  = 508
-    NotExtended                   = 510
-    NetworkAuthenticationRequired = 511
-  end
+  # `Lucky::Action::Status` has been removed in favor of
```
paulcsmith: I think this can be removed and instead added to the CHANGELOG or UPGRADE_NOTES

src/lucky/redirectable.cr
```diff
@@ -45,12 +45,12 @@ module Lucky::Redirectable
   # Redirect to the given path, with a human friendly status
   #
   # ```crystal
-  # redirect to: "/users", status: Status::MovedPermanently
+  # redirect to: "/users", status: HTTP::Status::MOVED_PERMANENTLY
```
paulcsmith: Thoughts on changing these docs to use the symbol version?

jwoertink: I think that's fine. The guides could mention that these map to the HTTP::Status and link to those docs, and just teach people to use the symbols instead. I'll remove all of the enum constants and replace with symbols.

jwoertink:
Ok, I removed most of the places we were passing in the enum in favor of just the symbol. Some spots still require the enum as doing `HTTP::Status::PERMANENT_REDIRECT.value` which I think is fine in that case. But let me know if you have any other thoughts on it.

PR #772, opened 2019-05-08
remove file extenions from path params

### LINKED ISSUE #771: Unable to add file extension on routes that end with a path variable
Defining an action like this one fails to compile when the route ends in a variable and have a file extension:

```crystal
  get "/reports/:report_name.csv" do
    text "test"
  end
end
```

```text
Syntax error in expanded macro: add_route:27: expecting token ')', not '.'

      report_name.csv,
                 ^
```

jwoertink:
## Purpose
fixes #771

## Description
If you have a route that ends in a path param as well as using a file extension, you'll get a compile time error showing a syntax error. This is because it assumed the whole path param included the file extension, then tried to make a method call `def whatever.txt`. This PR removes the file extensions if they exist, even if they are doubled up like `.xml.rss`. 

```crystal
class Reports < Lucky::Action
  get "/reports/:report_id.csv" do
    text "1,2,3"
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/action_route_params_spec.cr
+++ b/spec/lucky/action_route_params_spec.cr
@@ -1,9 +1,15 @@
-require "../../spec_helper"
+require "../spec_helper"
 
 include ContextHelper
 
 private class TestAction < Lucky::Action
   get "/test/:param_1/:param_2" do
+    text "test"
+  end
+end
+
+private class TestWithFileExtensionAction < Lucky::Action
+  get "/test/:param_1.csv" do
     text "test"
   end
 end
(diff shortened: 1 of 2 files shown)


spec/lucky/action_route_params_spec.cr
```diff
@@ -17,3 +23,11 @@ describe "Automatically generated param helpers" do
     typeof(action.param_2).should eq String
   end
 end
+
+describe "Testing params that end with file extension" do
+  it "generates helpers for the param name" do
+    action = TestWithFileExtensionAction.new(build_context, {"param_1" => "param_1_value"})
+    action.param_1.should eq "param_1_value"
+    typeof(action.param_1).should eq String
```
paulcsmith: Could you test that the output is the correct string and not just the right type? Just to make sure it works correctly. Might be worth doing in the one above too just to be sure

jwoertink:
Ok, after some discussion, this PR won't really work out, and my alternate solution would be too complicated. Going back to the drawing board, and will submit something better later that's less complicated.

Issue #759, opened 2019-04-11
Checking a request type in actions


jwoertink:
When you're inside an action, you have access to [these methods](https://github.com/luckyframework/lucky/blob/master/src/lucky/content_type_helpers.cr). One issue though is that these currently look for `Content-Type` header only. This works great in an API where you require a `Content-Type` to be passed, but from my recent findings, browsers don't pass this header with the request. That header is more for the response.

This means, doing this fails:

```crystal
route do
  if html?
    render(Home::Index)
  else
    raise Lucky::RouteNotFoundError.new(context)
  end
end
```

When I visit this page in both firefox, and chrome, here's the `headers` I have available:
```
HTTP::Headers{"Host" => "localhost:3001", "User-Agent" => "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:66.0) Gecko/20100101 Firefox/66.0", "Accept" => "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8", "Accept-Language" => "en-US,en;q=0.5", "Accept-Encoding" => "gzip, deflate", "Referer" => "http://localhost:3001/about", "DNT" => "1", "Connection" => "keep-alive", "Cookie" => "_session=abc123...", "Upgrade-Insecure-Requests" => "1", "Pragma" => "no-cache", "Cache-Control" => "no-cache"}
```

My thought is we can look at the accept header first then fallback to content-type after then finally an empty string. However, in doing this, it changes the logic a bit of how these are determined. If you look at my `Accept` header above, HTML is `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`. This means more of a regex approach would be needed. The other thing is I feel like with an html request, `html?` and `xml?` should both return true, but for an xml request `html?` would be false, and `xml?` would be true.

paulcsmith:
I didn't quite have time to read this all (feeling sick) but I have some code from another project= that should handle this I think

```
module Lucky::VerifyContentType
  macro included
    before verify_content_type
  end

  abstract def accepted_content_types

  def verify_content_type
    content_type = request.headers["Content-Type"]?

    if content_type.nil? || content_type == "*/*"
      content_type = accepted_content_types.first
      context.request.headers["Content-Type"] = content_type
    end

    if accepted_content_types.any? { |accepted_type| content_type.starts_with?(accepted_type) }
      continue
    else
[...]

PR #778, opened 2019-05-16
New fancy Lucky::RouteHandler

### LINKED ISSUE #771: Unable to add file extension on routes that end with a path variable

jwoertink:
## Purpose
fixes #771 

## Description
Lucky currently has some issues when it comes to handling routes that have a file extension. You can define a route like `get "/thing.html"`, and everything is fine, but if your route ends in a path param like `get "/thing/:id.json` then it throws an exception. 

Handling this becomes very tricky, and leads to a lot of gsubs and such. 

This PR moves the handling of this to a handler instead. If a request comes in with a file extension, we can now check the mime types for that extension, strip it off, and match against the path as defined without that extension. This gives us several new cool options. 

```crystal
  get "/some-action/:id" do
    if html?
     text "html request"
    elsif json?
     json("json")
    else
      text "no clue"
    end
  end
end
```

Now we make a call to `/some-action/123.html`, and it responds with "html request", or `/some-action/123.json`. Our paths can be defined with no file extension, and we can catch the request based on the requests Content-Type header. 

This also adds in a custom config which allows you to specify your own custom mime types in case [this list](https://crystal-lang.org/api/0.28.0/MIME.html) isn't enough for what you need. 

2 things to note:

1. Following this PR, there will need to be an update to the cli adding in a `config/mime_types.cr` file which would let you add your custom configs, as well as register the mime types. I wasn't sure of a better place for this to go.

2. Lucky has [several helper methods](https://github.com/luckyframework/lucky/blob/master/src/lucky/content_type_helpers.cr) that look at the request headers Content-Type like `html?`. Browsers don't send this by default as described in https://github.com/luckyframework/lucky/issues/759. So there is an extra spec in here that I commented out because it doesn't give me the response I would expect. 

## Checklist
* [x] - Use in production apps for a week with no issues 😉


jwoertink:
That's strange.. not sure why that failed...

spec/lucky/action_spec.cr
```diff
@@ -253,6 +253,18 @@ describe Lucky::Action do
...
   end
 
+  describe "when route is called with a file extension" do
+    it "still matches and responds" do
+      response = Tests::Show.new(build_context(path: "/tests/1.html"), params).call
+      response.body.should eq "test"
+    end
+
+    it "can be called with any weird extension and still match" do
+      response = Tests::Show.new(build_context(path: "/tests/1.js.php.erb"), params).call
```
edwardloveall: what a file...

spec/lucky/route_handler_spec.cr
```diff
@@ -0,0 +1,87 @@
+require "../spec_helper"
+
+include ContextHelper
+
+class SampleAction::Index < Lucky::Action
+  get "/sample-action" do
+    if html?
+      text "html test"
+    else
+      text "not html"
```
edwardloveall:
```suggestion
      text "something else"
```
This is a nit but it might read better

paulcsmith:
One concern I have is that this might cut off extensions when it is not expected. 

For example /users/foo@bar.com

Or /sites/foobar.com

Would the extension be cut off when accessing those params?

spec/lucky/route_handler_spec.cr
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+end
+
+describe Lucky::RouteHandler do
+  describe "when it finds the SampleAction" do
+    it "renders the html request" do
```
edwardloveall:
```suggestion
    it "using the Accept header" do
```

src/lucky/route_handler.cr
```diff
@@ -1,15 +1,36 @@
...
 class Lucky::RouteHandler
   include HTTP::Handler
 
+  Habitat.create do
+    setting mime_extensions : Hash(String, String) = {} of String => String
+  end
+
   def call(context)
+    original_path = context.request.path
+    format = get_extension(context.request.path)
```
edwardloveall: Is `format` a reserved crystal word or is github just messing up? If so, maybe use `extension` instead?

jwoertink:
In the case of something like `/users/jeremy@lucky.org`, yes, the `.org` would be cut off. I can see this being an issue, so that's valid. I'll see if I can add a spec to test for something like that. I see this as an edge case where you wouldn't define your route like `/users/:domain.com`, so we may be able to do something like if the mime type doesn't exist, then just pass it on through as is assuming it's html.. 

That would in turn create a new edge case where you do `/users/jeremy@lucky.org.json`, and expect a json result... but at that point I would suggest the user not do that 😛

paulcsmith:
Maybe we have a macro like `ignore_format_extension` or something like that. You can include that in your actions and it will use the full path. Might mean we need to store the original path in the context in an `original_path` variable or something. So that we can get at it when parsing params in Lucky actions

Whatever we decide I think it should be explicit opt-in or opt-out. If we get too clever with the mime types  I think it could be confusing. It could also  introduce bugs. Where if you add a new mime type it could mess up an existing route param. That could be bad

paulcsmith:
Also, I haven’t looked at the code yet but we should consider doing a Logger.warning that says the format was set and to what. So it’s clear something is happening. 

Maybe you already did that though...if so 🎉:D

src/lucky/route_handler.cr
```diff
@@ -1,15 +1,36 @@
...
+    context.request.path = context.request.path.gsub(format.to_s, "") if format
     handler = Lucky::Router.find_action(context.request)
     if handler
+      if format && (content_type = MIME.from_extension?(format))
+        Lucky.logger.debug({path_extension: format, new_content_type: content_type})
```
paulcsmith: Ah nice! Maybe we make it a warning so it is yellow?

paulcsmith: And maybe include “old_content_type” ? So it’s clear it is changing and how?

paulcsmith: Or maybe just “content_type_changed_to” a bit longer but I think clearer that it has been changed because of the extension

src/lucky/route_handler.cr
```diff
@@ -1,15 +1,36 @@
...
   include HTTP::Handler
 
+  Habitat.create do
+    setting mime_extensions : Hash(String, String) = {} of String => String
+  end
+
   def call(context)
+    original_path = context.request.path
+    format = get_extension(context.request.path)
+    context.request.path = context.request.path.gsub(format.to_s, "") if format
```
paulcsmith: The gain makes me nervous because it might overwrite query params or other parts of the path as well. Could you add a spec for that to make sure that doesn’t happen?

src/lucky/route_handler.cr
jwoertink: no, github is just rendering wrong.

jwoertink:
Looks like I'll have to merge in master now, but this last commit adds in a spec to check that we can still pass things like emails for path params. The only weird thing is since it's passed separately, it feels a little like cheating the spec. Maybe there's a different way to test that stuff?

If y'all think of any other specs I should write, let me know.

jwoertink:
Gonna close this out for now. I actually deleted this branch by accident anyway 😅 but there's been lots of changes since this, and we will need to look at an alternate way to handle file extensions at the end of a route with a route param.

PR #781, opened 2019-05-19
distance_of_time_in_words returns a string instead of text element

### LINKED ISSUE #780: Consider returning a string from time_ago_in_words instead of a text element
I was trying to make this code recently:

```crystal
time(
  time_ago_in_words(post.created_at),
  class: "created_at",
  datetime: post.created_at.to_rfc3339,
)
```

However, `time_ago_in_words` returns an `IO` object not a string. If I cast it to a string:

```crystal
time(
  time_ago_in_words(post.created_at).to_s,
  class: "created_at",
  datetime: post.created_at.to_rfc3339,
)
```

it compiles, but it spits out the HTML of the whole page! I believe this is because it's combining with the IO that renders the page so when you turn it to a string you get the whole page's html output. Not ideal 😉 .

What if instead we don't wrap `time_ago_in_words` in a `text` element? That way it can be used in a `time` element, or an `a`, or whatever!

edwardloveall:
## Purpose
closes #780 

## Description
Returning a text element (which uses the `IO` class) can cause weird bugs, like the entire HTML source of the page being printed out into the page. This alters the `distance_of_time_in_words` method to return a string instead which can be inserted into a `text` element, as an attribute of a different element, or even outside of a view entirely!

## Checklist


paulcsmith:
I think before merging this in wed also want to change a bunch of the other helpers methods that write directly to the IO. Otherwise some will work differently than others. Stuff like pluralize https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr

Some use raw so it doesn’t make sense to convert. Those should be documented that they write directly to IO though and should probably use a Nil return type so people don’t try to use them

edwardloveall:
Yeah, makes sense. I'll do that in a bit 👍

paulcsmith:
Nice branch name 😃

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### Changes since v0.13
 
 - Change `Lucky::BaseApp` to `Lucky::BaseAppServer`
+- `time_ago_in_words` returns a `String` instead of an `text` element (IO)
```
jwoertink: 0.14 has already been released, so this would more likely go under `### Changes since v0.14`. However, with that said, maybe all changelog changes can just go in a single commit? Kind of a pain to update all at once, but it helps to know what docs will need to be updated and such..

edwardloveall:
I made some progress on this. However, I'm not sure how to approach `truncate` because of its block overload. I'm not sure how to get it to run the block and then safely add the result onto the end of the truncated string. Here's my test case from `crystal play`:

```crystal
require "./src/lucky"
require "./spec/support/context_helper"
include ContextHelper

class TextHelperTestPage
  include Lucky::HTMLPage

  def render
  end
end

view = TextHelperTestPage.new(build_context)

view.link "#" do
  text = view.truncate("Here is a long test and I need a continue to read link", length: 27) do
    view.span "Continue"
  end.to_s
  view.text text
end.to_s
```

This returns some extra HTML because it's grabbing the existing HTML in the truncate method. I'd love an assist.

paulcsmith:
I think that is a pretty tricky edge case. Maybe for now we make the block version a different method. Like raw_truncate or something. What do you think?

edwardloveall:
Yeah I like that. There's a private method called `truncate_text` that I think will do what we want. I'll make that public and write some docs.

edwardloveall:
Cool, should be ready to go unless you want me to move the changelog stuff.

src/lucky/page_helpers/text_helpers.cr
```diff
@@ -28,22 +71,54 @@ module Lucky::TextHelpers
...
+  #   "You're such a nice and attractive person.",
+  #   phrases: ["nice", "attractive"],
+  #   highlighter: "<strong>\\1</strong>"
+  # )
+  # ```
+  # outputs:
+  # ```html
+  # You're such a <strong>nice</strong> and <strong>attractive</strong> person.
+  # ```
   def highlight(text : String, phrases : Array(String | Regex), highlighter : Proc | String = "<mark>\\1</mark>")
```
paulcsmith: Actually, could you make the return types `: Nil` for any of the methods that are meant to write HTML directly to the IO so that it won't ever accidentally be used as a `String`? Might also be worth adding a note like: `This method writes HTML directly to the page. It does not return a String` or something like that. Thoughts?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### Changes since v0.13
 
 - Change `Lucky::BaseApp` to `Lucky::BaseAppServer`
+- Many text helpers now return a `String` instead of appending to the view (`cycle`, `excerpt`, `highlight`, `pluralize`, `time_ago_in_words`, `to_sentence`, `word_wrap`) [#781](https://github.com/luckyframework/lucky/pull/781)
```
paulcsmith: Might be worth putting in bold `**Breaking change** - Many text...` since this will definitely break stuff using it

edwardloveall:
@paulcsmith done in 537ee85

paulcsmith:
Perfect! Thank you

PR #783, opened 2019-05-25
Update shard.yml and version.cr to v0.14.0 and crystal 0.28.0

edwardloveall:
I saw that you had [updated the version file to 0.14.0](https://github.com/luckyframework/lucky/commit/5d0d9cf2fd7e68baf018cc2e91eafcd527b9853a) but it looks like it never got merged into master. Was this a mistake?

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,7 +1,7 @@
 name: lucky
-version: 0.13.3
+version: 0.14.0
 
-crystal: 0.27.2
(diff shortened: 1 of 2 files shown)


paulcsmith:
No it is because 0.14 was based on the 0.13.x branch and just made a few changes to work with Crystal 0.28. So it isn’t meant to be in master. Master will be 0.15 once it is ready

PR #788, opened 2019-06-03
Add explicit require of ./lucky/route_helper before ./lucky/*

bcardiff:
Due to a bug present up to Crystal 0.28.0, this require was somehow not needed. But it will be needed starting on Crystal 0.29.0.


src/lucky.cr
```diff
@@ -15,6 +15,7 @@ require "./lucky/exceptions"
 require "./lucky/response"
 require "./lucky/cookies/*"
 require "./lucky/secure_headers/*"
+require "./lucky/route_helper"
```
jwoertink: So why does this one specifically need to be required where there's a ton of other files that do not? Is this because of an order where the RouteHelper is called before some others?

bcardiff: Yes. Some file in `./lucky/*` that is before route_helper uses `: RouteHelper` explicitly. Both, `redirectable.cr` and `tags/form_helpers.cr` match that. Not sure if one or both were the culprit.

paulcsmith:
Thanks @bcardiff  :D

bcardiff:
FYI you might be willing to wait for ameba to release a 0.29.0 compatible version.

PR #795, opened 2019-06-07
Adding support for an AssetHost

### LINKED ISSUE #794: Asset Host Option
Lucky::Server.settings should have an asset_host option that can be configured mainly for development. The default value can be an empty string so that it doesn't break exiting or new apps. That value can be configured for production to point to a remote cdn or something else.

russ:
Added an asset_host option to the Lucky::Server.settings
That setting is prepended to the asset path. It can be configured in the
application the same way host and port are.

## Purpose
Adding an asset_host option provides a reasonable way to configure production apps to use a cdn for assets.

fixes #794 

## Description
```
Lucky::Server.configure do |settings|
  if Lucky::Env.production?
    settings.asset_host = "production.com"
  else
    settings.asset_host = "/"
  end
end
```

## Checklist
* [NA] - Inline documentation has been added and/or updated

### DIFF
--- a/src/lucky/asset_helpers.cr
+++ b/src/lucky/asset_helpers.cr
@@ -14,7 +14,7 @@ module Lucky::AssetHelpers
 
     {% if path.is_a?(StringLiteral) %}
       {% if Lucky::AssetHelpers::ASSET_MANIFEST[path] %}
-        {{ Lucky::AssetHelpers::ASSET_MANIFEST[path] }}
+        Lucky::Server.settings.asset_host + {{ Lucky::AssetHelpers::ASSET_MANIFEST[path] }}
       {% else %}
         {% asset_paths = Lucky::AssetHelpers::ASSET_MANIFEST.keys.join(",") %}
         {{ run "../run_macros/missing_asset", path, asset_paths }}
@@ -48,7 +48,7 @@ module Lucky::AssetHelpers
   def dynamic_asset(path)
     fingerprinted_path = Lucky::AssetHelpers::ASSET_MANIFEST[path]?
     if fingerprinted_path
-      fingerprinted_path
+      Lucky::Server.settings.asset_host + fingerprinted_path
     else
       raise "Missing asset: #{path}"
     end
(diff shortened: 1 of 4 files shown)


jwoertink:
Looks like the failure is related to ameba updating to only support crystal 0.29.0, and Lucky running on an earlier version.

russ:
Yeah, that is a versioning error. Should Lucky be locked to a specific version of Ameba or should this pr wait until Lucky is upgraded to 0.29.0?

jwoertink:
I think we need to lock all shards down to very specific versions. It's a bit more restrictive, but there's nothing worse than running `shards update`, and everything is borked due to some random update. (e.g. this...)

paulcsmith:
Yeah I think this is a great idea. Feel free to lock ameba to an older version in shard.yml to get this passing. Will try to take a closer look soon!

src/lucky/asset_helpers.cr
```diff
@@ -14,7 +14,7 @@ module Lucky::AssetHelpers
 
     {% if path.is_a?(StringLiteral) %}
       {% if Lucky::AssetHelpers::ASSET_MANIFEST[path] %}
-        {{ Lucky::AssetHelpers::ASSET_MANIFEST[path] }}
+        Lucky::Server.settings.asset_host + {{ Lucky::AssetHelpers::ASSET_MANIFEST[path] }}
```
jwoertink: Can you look at a spec for this? Maybe in https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/spec/lucky/asset_helpers_spec.cr? Habitat has a `temp_config` method so you can set an asset_host to some production URL, and then test that the asset path has the whole thing.

russ:
I added some specs around that. Still failing on the ameba version though.

jwoertink:
Just lock it to 0.9.1 here https://github.com/luckyframework/lucky/blob/master/shard.yml#L63. Then the next version we roll out, we'll update that to 0.10.0

PR #798, opened 2019-06-08
Shards build --production

paulcsmith:
This way we don't load in dev dependencies


paulcsmith:
Do not delete this branch. It is needed so people can use 0.14.1

PR #801, opened 2019-06-13
Fix HSTS with Time::MonthSpan and release 0.15.1

paulcsmith:
For some reason Int#year and Int#month return a Time::MonthSpan. This
made it so setting the max_age to `1.year` did not work.

This commit makes it so Time::MonthSpans are accepted and converted to a
regular Time::Span for use as the max_age

cc @jwoertink 

I'm gonna merge and tag this so it works with the example in the generated config/server.cr file. but wanted to let you know in case you run into it

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,5 +1,5 @@
 name: lucky
-version: 0.15.0
+version: 0.15.1
 
 crystal: 0.29.0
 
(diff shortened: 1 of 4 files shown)


jwoertink:
Ah. I wonder if that was a crystal change recently? Good catch on that.

paulcsmith:
Yeah I think that might be the case. I think I get the reasoning behind it but it seems weird to me. I might open an issue about it. It seems like those methods should return the same type

jwoertink:
Working with Time in crystal is actually pretty frustrating.

PR #809, opened 2019-06-16
number_to_currency now returns a String instead of writing to the view

### LINKED ISSUE #790: number_to_currency still writes to IO
[this PR](https://github.com/luckyframework/lucky/pull/781) changed a few of the text helpers to return a string as opposed to writing directly to the view IO. The rule is that if the method returns HTML, it should render directly to the view, and if it only returns a plain string, then the method should just return the string.

Looks like [number_to_currency](https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/number_to_currency.cr#L33) is still calling `text` which renders to the view IO, but there's no reason (that I can see) that it should.

jwoertink:
## Purpose
fixes #790

## Description
In a recent PR, most of the page helpers were converted to return a String instead of writing to the view via `text`. The rule is now if the method generates HTML, it writes to the view, otherwise it should return a String. This method was missed during that commit.

## Checklist

Specs pass when settings avram back to 0.9.4. Since Avram renamed Field and Form stuff recently, and Lucky pulls from Avram master, specs fail.

### DIFF
--- a/spec/lucky/number_to_currency_spec.cr
+++ b/spec/lucky/number_to_currency_spec.cr
@@ -11,44 +11,44 @@ end
 describe Lucky::NumberToCurrency do
   describe "number_to_currency" do
     it "accepts Float" do
-      view.number_to_currency(29.92).to_s.should eq(HTML.escape("$29.92"))
+      view.number_to_currency(29.92).should eq(HTML.escape("$29.92"))
     end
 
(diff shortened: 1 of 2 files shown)


edwardloveall:
This is all fine, lucky just needs to be updated: https://github.com/luckyframework/lucky/pull/810

PR #811, opened 2019-06-18
[WIP] More reliable and configurable asset helpers

### LINKED ISSUE #717: Allow customizing manifest retry timeout

paulcsmith:
Closes #717

Note the ENV vars must be available at compile time. This means setting
them in a .env file will not work. Instead add them to your command in
Procfile.dev.

    web: ASSET_HELPER_TIMEOUT_IN_SECONDS=15 lucky watch

## Purpose

Previously the timeout was hardcoded. For asset compilation that takes a long time this was not fast enough.

It also only looked for a manifest but not a particular file. The problem with that is that you may have an old manifest that is in the process of being updated. That means Lucky incorrectly thinks the asset doesn't exist, when in reality the asset compilation process just hasn't finished and therefore hasn't updated the manifest.

Instead we look for particular assets. So even if the manifest exists it will retry until the new asset is there or it times out

## Checklist


paulcsmith:
This still needs some work to fix a couple errors (like if the manifest doesn't exist. But from what I can tell the core part works. It is configurable, a bit more reliable, and the "did you mean" works for compile time assets.

src/run_macros/get_fingerprinted_asset.cr
```diff
@@ -0,0 +1,64 @@
...
+      raise_missing_manifest_error
+    end
+  end
+
+  private def manifest_exists?
+    File.exists?(Lucky::AssetManifest::MANIFEST_PATH)
+  end
+
+  private def max_retries : Int32
+    seconds_to_wait = ENV["ASSET_HELPER_TIMEOUT_IN_SECONDS"]?.try(&.to_i) || 5
```
paulcsmith: In the future I have an idea where we can configure this just like our other code (Habitat in a config file) but that requires reworking how tasks work and probably won't be done for awhile. So for now an ENV should do the trick

src/run_macros/get_fingerprinted_asset.cr
```diff
@@ -0,0 +1,64 @@
...
+    (seconds_to_wait / RETRY_AFTER).to_i
+  end
+
+  private def raise_missing_manifest_error
+    puts "Manifest at #{Lucky::AssetManifest::MANIFEST_PATH} does not exist".colorize(:red)
```
paulcsmith: This is not currently working

src/lucky/asset_manifest.cr
```diff
@@ -0,0 +1,27 @@
...
+    INSTANCE.to_h[path]?
+  end
+
+  def self.non_fingerprinted_paths : Array(String)
+    INSTANCE.to_h.keys
+  end
+
+  # :nodoc:
+  def to_h : Hash(String, String)
+    @_to_h ||= begin
```
paulcsmith: Cached so we only parse the manifest once at runtime

spec/spec_helper.cr
```diff
@@ -3,8 +3,6 @@ require "../src/lucky"
 require "../tasks/**"
 require "./support/**"
 
-Lucky::AssetHelpers.load_manifest
```
paulcsmith: No longer needed :D

paulcsmith:
Hmm I’m guessing that the “did you mean” happened because it timed out before the asset was in the manifest, but by the time it generated the message it was there again.

I would try a pretty conservative 30s and see if that consistently works. You want to have a padded timeout so if it takes a bit longer it still waits.

Can you let me know if 30s constantly works?

jwoertink:
I set to 35, and it still fails. Even with all of the assets completing within 19s. I can't can't get it to boot with `lucky dev` anymore. I have to run the two commands in separate tabs. 

```
web: ASSET_HELPER_TIMEOUT_IN_SECONDS=35 lucky watch --reload-browser
assets: yarn watch
```

```
Finished 'clean' after 151 ms
Finished 'watch' after 206 ms
Finished 'default' after 236 ms
Finished 'fonts' after 219 ms
Finished 'vendored_js' after 310 ms
Finished 'js' after 2.44 s
Finished 'fontawesome' after 2.42 s
Finished 'images' after 2.63 s
Finished 'css' after 19 s
```

paulcsmith:
Ok back to the drawing board lol. Thanks for giving it a try. I'll try to figure out what's up

jwoertink:
I'm doing some debugging with it. I'll let you know if I find where it's getting hung up

paulcsmith:
Closing for now. Will revisit later with something that works :)

PR #812, opened 2019-06-18
Update UPGRADE_NOTES.md

KCErb:


UPGRADE_NOTES.md
```diff
@@ -22,7 +22,10 @@ brew upgrade lucky
 
 - Run `shards update`
 
-- Rename `src/server.cr` to `src/start_server.cr`. 
+- Rename `src/server.cr` to `src/start_server.cr`.
+- Edit `src/start_server.cr` by changing 
+    * `app` to `app_server` and `App` to `AppServer`.
+    * change the `puts` to output `"Listening on http://#{app_server.host}:#{app_server.port}"`
```
paulcsmith: This message should probably be removed actually. There is a new message that takes the place of this one and is much easier to read

KCErb: You mean the puts line? If I understand you correctly, in a future release you'll remove this puts line so may as well remove now in the upgrade notes?

paulcsmith: Even in the current release it is removed most of the time. It is wrapped in this if https://github.com/luckyframework/lucky_cli/blob/6796508e6a03e11bd4e1789cbe3d22dc5109d792/src/web_app_skeleton/src/start_server.cr.ecr#L9

I think it should be completely removed actually but I just need to test it out. For the upgrade notes I think we can say to remove it though.

KCErb:
Hmm, I just used the github gui to submit these changes. I'm guessing that's built against my fork which is behind. Could that be causing the travis failure here?

paulcsmith:
No, I just broke it with a recent commit. Oops :) I'll fix it. But no worries about CI. It's the README so I feel comfortable merging with a red CI :D

KCErb:
Oh ok, it also looks like github fast forwarded my fork for me which is cool.

paulcsmith:
Nice! :D GitHub is awesome

PR #816, opened 2019-06-18
Remove a redundant 'if'

smadeja:
`text` is a `String` and won't ever be falsey.

## Checklist


jwoertink:
Makes sense to me. Thanks!

PR #818, opened 2019-06-19
reverts returning a String for the highlight

### LINKED ISSUE #792: highlight no longer writes to the IO
The [highlight](https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr#L101) method used to write the page, and then that got updated to not; however, that was a mistake we over looked 😅 

That should go back to writing directly to the page since it outputs HTML.

jwoertink:
## Purpose
fixes #792 

## Description
In https://github.com/luckyframework/lucky/pull/781 a few methods were changed to return a String instead of writing directly to the view. However, some of those methods return HTML and would still need to be passed in to `raw` to work. So this reverts this method to write back to the view.

## Checklist

### DIFF
--- a/src/lucky/page_helpers/text_helpers.cr
+++ b/src/lucky/page_helpers/text_helpers.cr
@@ -76,6 +76,7 @@ module Lucky::TextHelpers
   # This can be customized with the `highlighter` argument.
   #
   # **Note: This method writes HTML directly to the page. It does not return a
+  # String**
   #
   # ```crystal
   # highlight("Crystal is type-safe and compiled.", phrases: ["type-safe", "compiled"])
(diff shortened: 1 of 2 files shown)


src/lucky/page_helpers/text_helpers.cr
```diff
@@ -75,6 +75,8 @@ module Lucky::TextHelpers
   # `phrases` array. The default is to wrap each with the `<mark>` element.
   # This can be customized with the `highlighter` argument.
   #
+  # **Note: This method writes HTML directly to the page. It does not return a
```
edwardloveall:
```suggestion
  # **Note: This method writes HTML directly to the page. It does not return a
  # String.**
```

jwoertink: ha! Good catch.

PR #820, opened 2019-06-19
separating the helper methods

### LINKED ISSUE #791: Splitting up page helpers
We have some page helpers like `truncate_text` that just return a string, and some like `simple_format` that write to page directly. 

I think it would be good to split these out in to two separate modules so we can easily distinguish which ones do which. It will help documenting them a little easier too.

jwoertink:
## Purpose
fixes #791 

## Description
Now that we have a distinction between helper methods that return `String` VS ones that actually return `Nil` because they write to the view directly, we can separate these methods out in to their own modules. This will make it easier for people to decipher which ones do what. As well as documentation.

## Checklist

### DIFF
--- a/src/lucky/page_helpers/html_helpers.cr
+++ b/src/lucky/page_helpers/html_helpers.cr
@@ -1,0 +1,159 @@
+# These helper methods will write directly to the view.
+module Lucky::HTMLHelpers
+  # Shortens text after a length point and inserts content afterward
+  #
+  # **Note: This method writes HTML directly to the page. It does not return a
+  # String.**
+  #
+  # This is ideal if you want an action associated with shortened text, like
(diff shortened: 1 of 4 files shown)


src/lucky/page_helpers/html_helpers.cr
```diff
@@ -0,0 +1,159 @@
+# These helper methods will write directly to the view.
+module Lucky::HTMLHelpers
```
paulcsmith: What do you think about HTMLTextHelper? So it’s clear it’s for text?

jwoertink: lol. I originally had that, and then second guessed myself. I'll go back to that.

PR #821, opened 2019-06-20
moving ContentTypeHelpers to RequestTypeHelpers. 

### LINKED ISSUE #759: Checking a request type in actions

jwoertink:
## Purpose
Fixes #759 

## Description
The `ContentTypeHelpers` has a method `html?` which checks the `Content-Type` header for "text/html". The problem is, at least with Firefox and Chrome, no `Content-Type` header is actually sent. So if your code relied on code like

```crystal
get "/" do
  if html?
    render HomePage
  else
    raise "only html here"
  end
end
```

Your app would always raise in a browser request. This does a few things

* Adds in a fallback check for the `Accept` header in case the `Content-Type` is blank
* Rename `ContentTypeHelpers` to `RequestTypeHelpers` since it checks more than just ContentType

## Checklist


paulcsmith:
Github hates me this week. I can’t merge this no matter what I try 😂 

795BD863-910E-43AC-862A-D2CF1D588A99

Could you merge this?

paulcsmith:
Github hates me this week. I can’t merge this no matter what I try 😂 

795BD863-910E-43AC-862A-D2CF1D588A99

Could you merge this?

paulcsmith:
And when I tried to comment lol

55AEF569-02A6-4753-9A3B-E0A826115DCE

Looks like eventually it went through though...I hope

jwoertink:
You win! You found the github bugs

PR #832, opened 2019-06-27
Added in memoize.

### LINKED ISSUE #35: Expose and memoize in one go
```crystal
expose :users, UserQuery.all, memoize_as: UserQuery

# or
expose_memoized users : UserQuery, UserQuery.all
```

jwoertink:
## Purpose
fixes #35

## Description
This PR gives a new macro called `memoize`. The purpose of the `memoize` macro is to allow you to define a method in your `Lucky::Action` class that may be an expensive or long running task, and have the value of that method only calculated once. 

```crystal
  memoize def current_user : User
    UserQuery.first
  end
 
  expose current_user

  get "/some-action" do
    render UserPage
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/memoize_spec.cr
+++ b/spec/lucky/memoize_spec.cr
@@ -2,76 +2,32 @@ require "../spec_helper"
 
 include ContextHelper
 
-class OnlyMemoize < Lucky::Action
-  @i = 0
-  memoize def long_task : Int32
-    proc = ->{ @i += 1 }
-    proc.call
+private class ObjectWithMemoizedMethods
+  include Lucky::Memoizable
+  getter times_method_1_called = 0
+  getter times_method_2_called = 0
+
+  memoize def method_1 : String
+    @times_method_1_called += 1
+    "method_1"
   end
 
-  get "/memoize/one" do
-    long_task
-    text long_task.to_s
-  end
-end
-
-class MultiMemoize < Lucky::Action
-  memoize def task_one : String
-    "running task_one"
-  end
-
-  memoize def task_two : Time
-    Time.utc
-  end
-
-  get "/memoize/multi" do
-    task_one
-    task_two
-    text "Still works"
-  end
-end
-
-class MemoizeOnPage < Lucky::Action
-  @seed = 0
-  memoize def big_data : Int32
-    proc = ->{ @seed += 1 }
-    proc.call
-  end
-
-  expose big_data
-
-  get "/memoze/on-page" do
-    render MemoizePage
-  end
-end
-
-class MemoizePage
-  include Lucky::HTMLPage
-
-  needs big_data : Int32
-
-  def render
-    h1 @big_data.to_s
-    h2 @big_data.to_s
-    h3 @big_data.to_s
-    h4 @big_data.to_s
-    h5 @big_data.to_s
+  memoize def method_2 : Int32
+    @times_method_2_called += 1
+    5
   end
 end
 
 describe "memoizations" do
   it "only calls the long_task once" do
-    response = OnlyMemoize.new(build_context, params).call
-    response.body.should contain "1"
-  end
+    object = ObjectWithMemoizedMethods.new
 
-  it "has no conflicts when memoizing numerous methods" do
-    response = MultiMemoize.new(build_context, params).call
-    response.body.should contain "Still works"
-  end
+    object.method_1.should eq "method_1"
+    2.times { object.method_1 }
+    object.times_method_1_called.should eq 1
 
-  it "can expose and pass memoized method to page" do
-    data = MemoizeOnPage.new(build_context, params).big_data
-    MemoizePage.new(build_context, big_data: data).render.to_s.should contain %(<h5>1</h5>)
+    object.method_2.should eq 5
+    9.times { object.method_2 }
+    object.times_method_2_called.should eq 1
   end
 end
--- a/spec/lucky/memoize_spec.cr
+++ b/spec/lucky/memoize_spec.cr
@@ -1,0 +1,77 @@
+require "../spec_helper"
+
+include ContextHelper
+
+class OnlyMemoize < Lucky::Action
+  @i = 0
+  memoize def long_task : Int32
+    proc = ->{ @i += 1 }
+    proc.call
+  end
+
+  get "/memoize/one" do
+    long_task
+    text long_task.to_s
+  end
+end
+
+class MultiMemoize < Lucky::Action
+  memoize def task_one : String
+    "running task_one"
+  end
+
+  memoize def task_two : Time
+    Time.utc
+  end
+
+  get "/memoize/multi" do
+    task_one
+    task_two
+    text "Still works"
+  end
+end
+
+class MemoizeOnPage < Lucky::Action
+  @seed = 0
+  memoize def big_data : Int32
+    proc = ->{ @seed += 1 }
+    proc.call
+  end
+
+  expose big_data
+
+  get "/memoze/on-page" do
+    render MemoizePage
+  end
+end
+
+class MemoizePage
+  include Lucky::HTMLPage
+
+  needs big_data : Int32
+
+  def render
+    h1 @big_data.to_s
+    h2 @big_data.to_s
+    h3 @big_data.to_s
+    h4 @big_data.to_s
+    h5 @big_data.to_s
+  end
+end
+
+describe "memoizations" do
+  it "only calls the long_task once" do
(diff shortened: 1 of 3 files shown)


src/lucky/memoizable.cr
```diff
@@ -0,0 +1,38 @@
...
+  # The `memoize` method will raise a compile time exception if you forget to include
+  # a return type for your method, or if your return type is a `Union`.
+  macro memoize(method_def)
+    {% raise "Return type must not be a Union" if method_def.return_type.is_a?(Union) %}
+    {% raise "You must include a return type for memoize methods" if method_def.return_type.is_a?(Nop) %}
+
+    @__{{ method_def.name }} : {{ method_def.return_type }}? = nil
+
+    def _{{ method_def.name }}{% if method_def.args.size > 0 %}({{ method_def.args }}){% end %} : {{ method_def.return_type }}
+      {{ method_def.body }}
```
robacarp: wow, cool.

src/lucky/memoizable.cr
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def {{ method_def.name }}{% if method_def.args.size > 0 %}({{ method_def.args }}){% end %} : {{ method_def.return_type }}
```
robacarp: the parameters list is a bit verbose...not to be too meta, but should that (could that?) be memoized in macrospace?

jwoertink: 🤷‍♂ no clue. Have any pseudo code example of what you're thinking?

bdtomlin:
@jwoertink,

Sorry to just pop in here on this, but I saw this and was definitely interested in the functionality.

It seems like a more complete implementation of memoization using a hash would be needed.

This implementation would give what I would think to be unexpected results. After the first call, all subsequent calls to the memoized method would return the original value even when different args are given.

Consider the following example:

```crystal
class BrowserAction
  memoize def calculate_data(month : String) : ReportData
    # some heavy task to return data
  end
end

calculate_data("November") #=> {NovemberData}
calculate_data("December") #=> {NovemberData}
calculate_data("January") #=> {NovemberData}
```

Again, sorry for lurking :) I discovered Lucky very recently and I'm very excited about what you guys are doing.

jwoertink:
Ah, I see what you're saying. I actually didn't go through to test with args. I was on the fence of even adding it in for this reason. I have an idea though that I'll give a shot, and I'll use your example as my test subject. 

Also, welcome to Lucky @bdtomlin! Thanks for the help 😄

jwoertink:
Ok, so looking at how the rails memoize method worked, it took a list of Symbols https://apidock.com/rails/ActiveSupport/Memoizable/memoize `memoize :big_data, :long_task, :whatever`. 

So the big question here is, does this one take args or not? @paulcsmith @edwardloveall 
Also curious what everyone else's thoughts are? @robacarp @bdtomlin 

My vote is no args allowed, but what do y'all think?

paulcsmith:
Yeah I think no args. Doing it with args is a bit too complex and I’m not sure is the best use of memoize. I think it should be reserved for very simple but costly calculations. Too much complexity would likely introduce bugs. Either because the memoize macro didn’t handle it right, or because it’s not what the user expected.

bdtomlin:
I think no args is fine, it avoids a lot of unnecessary complications. Raising when memoizing a method with args seems like the right solution to me in this case.

src/lucky/memoizable.cr
robacarp: I was thinking something like this:

```crystal
{% 
if method_def.args.size > 0
  rendered_args = method_def.args
else
  rendered_args = ""
end
%}

def {{ method_def.name }} {{ rendered_args }} : {{ method_def.return }}
end

def _{{ method_def.name }} {{ rendered_args }} : {{ method_def.return }}
end
```

jwoertink: Ah. We decided to just drop args anyway 😂

spec/lucky/memoize_spec.cr
```diff
@@ -0,0 +1,77 @@
...
+
+  it "has no conflicts when memoizing numerous methods" do
+    response = MultiMemoize.new(build_context, params).call
+    response.body.should contain "Still works"
+  end
+
+  it "can expose and pass memoized method to page" do
+    data = MemoizeOnPage.new(build_context, params).big_data
+    MemoizePage.new(build_context, big_data: data).render.to_s.should contain %(<h5>1</h5>)
+  end
```
paulcsmith: This spec uses actions and pages, but I think memoization can apply to any object so you could probably add a fake class here that includes memoizable. You could also hook into the instance to clarify what we're testing. I'm finding it hard to explain so I'll try to write a bit of (untested) code. Something like this:

```crystal
private class ObjectWithMemoizedMethods
  include Lucky::Memoizable
  getter times_method_1_called = 0
  getter times_method_2_called = 0
  
  # Also do this for method_2
  memoize def method_1 : String
    @time_method_1_called += 1 
     "method_1"
  end
end

it "memoizes the method" do
  object = ObjectWithMemoizedMethods.new

  # Do same for method_2
  object.method_1.should eq "method_1"
  2.times { object.method_1 }
  object.times_method_1_called.should eq 1
end
```

If this works then it should be usable in pages, actions and anywhere!

Which makes me think...should we add an charm for `Object` that includes this by default so you can memoize outside actions out of the box? No need to block this PR but something we can think about adding in the future

src/lucky/memoizable.cr
```diff
@@ -0,0 +1,40 @@
...
+  # a return type for your method, or if your return type is a `Union`.
+  macro memoize(method_def)
+    {% raise "Return type of memoize method must not be a Union" if method_def.return_type.is_a?(Union) %}
+    {% raise "You must define a return type for memoize methods" if method_def.return_type.is_a?(Nop) %}
+    {% raise "Memoize methods can not be defined with arguments" if method_def.args.size > 0 %}
+
+    @__{{ method_def.name }} : {{ method_def.return_type }}? = nil
+
+    # no cache
+    def _{{ method_def.name }} : {{ method_def.return_type }}
```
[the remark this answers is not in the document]
jwoertink: We could just call `method_def.body` below, but having a way to access a non-cached version would be helpful in debugging and such. The rails memoize actually had the same.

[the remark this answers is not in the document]
paulcsmith: Ahhh that makes sense! Ok one more thought then and feel free not to implement because I'm not sure about it. What do you think about `{{ method_def.name }}__uncached`?

spec/lucky/memoize_spec.cr
jwoertink: ok, I can dig it. 👍

paulcsmith:
One other thing. This can be added later, but it might be worth beefing this up so it also caches nil. Right now if the result is nil it will keep re-running the method. Instead we could another instance var that is a `Bool = false`. And the memoize body would be something like this:

```crystal
@{{ method_def.anem}}_memoized : Bool = false
def {{ method_def.name }}
  if @{{ method_def.name}}_memoized
    # return the memoized result
  else
    # store memoized_result
    # Set memoization status to memoized
    @{{ method_def.name}}_memoized = true
  end
end
```

jwoertink:
I originally wanted to allow for `nil` values, but it makes it a lot harder with this compiler. If we do that, then we would need some wonky logic to say if it's not nillable, then call `.not_nil!` so we don't always have to typecast the method (i.e. `current_user.as(User).email`). But then if it is nillable as a Union, you'd have to typecast everywhere you expect it to not be nil. 

Maybe we can plan that out a bit better for a later PR?

paulcsmith:
Yeah let's tackle it in another PR. I think it is doable, but can be done later. This is a great feature as it is!

spec/lucky/memoize_spec.cr
```diff
@@ -0,0 +1,33 @@
...
+
+  memoize def method_2 : Int32
+    @times_method_2_called += 1
+    5
+  end
+end
+
+describe "memoizations" do
+  it "only calls the long_task once" do
```
paulcsmith: Maybe `only call the method body once`?

src/lucky/memoizable.cr
```diff
@@ -0,0 +1,40 @@
+module Lucky::Memoizable
+  # Caches the return value of the method. Helpful for expensive methods that are called
+  # more than once.
```
paulcsmith: I am fairly certain the docs only take the first line into account for the summary. If that's the case (can check with `crystal docs` then I think this should go on one line

paulcsmith:
Aw yeah! 🎉

PR #835, opened 2019-07-01
Draft the beefed up responsive meta tag

### LINKED ISSUE #828: Beef up responsive meta tag
We have this cool method 

https://github.com/luckyframework/lucky/blob/master/src/lucky/tags/specialty_tags.cr#L40

But the viewport meta tag has a lot [more options](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag#Viewport_basics). We should have this method take some options to update things like `user-scalable`, `minimum-scale`, and `maximum-scale`.

smadeja:
Closes #828. The docs will have to be updated, I wanted to get some feedback first.

## Checklist

### DIFF
--- a/src/lucky/tags/specialty_tags.cr
+++ b/src/lucky/tags/specialty_tags.cr
@@ -39,7 +39,7 @@ module Lucky::SpecialtyTags
   # details.
   def responsive_meta_tag(**options)
     options = {width: "device-width", initial_scale: "1"}.merge(options)
-    meta name: "viewport", content: build_content_attr_value(options)
+    meta name: "viewport", content: build_viewport_properties(options)
   end
 
   # Adds *string* directly to the rendered HTML with no escaping.
@@ -76,7 +76,7 @@ module Lucky::SpecialtyTags
     view
   end
 
-  private def build_content_attr_value(options)
+  private def build_viewport_properties(options)
     String.build do |attrs|
       options.each_with_index do |key, value, index|
         attrs << ", " if index > 0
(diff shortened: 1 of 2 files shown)


src/lucky/tags/specialty_tags.cr
```diff
@@ -74,4 +75,14 @@ module Lucky::SpecialtyTags
     how_many.times { raw("&nbsp;") }
     view
   end
+
+  private def build_content_attr_value(options)
```
paulcsmith: Could you name this something like `build_responsive_meta_tag_content_attr`. So that it is clear it is for the responsive_meta_tag? We may need something like this in the future for other methods, but until then I think being specific will add clarity

src/lucky/tags/specialty_tags.cr
```diff
@@ -37,8 +37,9 @@ module Lucky::SpecialtyTags
   # See the [MDN's documentation on the meta
   # tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta) for
   # details.
```
jwoertink: Can you update these docs to point to https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag ? That way it's easier to see what options can be passed in to this.

paulcsmith: Great idea!

src/lucky/tags/specialty_tags.cr
smadeja: Would `build_viewport_properties` be ok? If so, than I think it's ready.

PR #838, opened 2019-07-02
New XML rendering option

### LINKED ISSUE #803: XML rendering support
We already have `json`, so why not just add a nice helper method for `xml`? 

Right now you can do

```crystal
Lucky::TextResponse.new(context, content_type: "text/xml; charset=utf-8", body: string status: 200)
```

but I'm thinking we do something like

```crystal
private def xml(body : String, content_type : String = "text/xml",  status : Int32? = nil) : Lucky::TextResponse
    Lucky::TextResponse.new(context, content_type, body, status: status)
end
```

https://github.com/luckyframework/lucky/blob/master/src/lucky/renderable.cr#L164

The only thing we would need to consider is if that `content_type` should go there. Following along with other examples, it shouldn't; however, XML content types could be `application/rss+xml` if you're building RSS, or you may want to add the charset. Maybe a better solution is all the others allow you to override the content_type?

jwoertink:
## Purpose
fixes #803

## Description
This adds in a new action render method `xml` similar to `json` and `text`. This also refactors the methods a bit to proxy down to a `send_response` method which actually handles it. I also noticed a possible bug where settings a status on json requests was probably a little weird. 

## Checklist

Site note, I wonder if these methods shouldn't be private so we can add docs to some of them and have it generated with `crystal docs`?

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -154,6 +154,10 @@ module Lucky::Renderable
     plain_text(body, status: status.value)
   end
 
+  private def text(*args, **named_args)
+    {% raise "'text' in actions has been renamed to 'plain_text'" %}
+  end
+
   @[Deprecated("`render_text deprecated. Use `plain_text` instead")]
   private def render_text(*args, **named_args) : Lucky::TextResponse
     plain_text(*args, **named_args)
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -142,24 +142,25 @@ module Lucky::Renderable
     file(path, content_type, disposition, filename, status.value)
   end
 
-  private def send_response(body : String, content_type : String, status : Int32? = nil) : Lucky::TextResponse
+  private def send_text_response(body : String, content_type : String, status : Int32? = nil) : Lucky::TextResponse
     Lucky::TextResponse.new(context, content_type, body, status: status)
   end
 
-  private def text(body : String, status : Int32? = nil) : Lucky::TextResponse
-    send_response(body, "text/plain", status)
+  private def plain_text(body : String, status : Int32? = nil) : Lucky::TextResponse
+    send_text_response(body, "text/plain", status)
   end
 
-  private def text(body : String, status : HTTP::Status) : Lucky::TextResponse
-    text(body, status: status.value)
+  private def plain_text(body : String, status : HTTP::Status) : Lucky::TextResponse
+    plain_text(body, status: status.value)
   end
 
+  @[Deprecated("`render_text deprecated. Use `plain_text` instead")]
   private def render_text(*args, **named_args) : Lucky::TextResponse
-    text(*args, **named_args)
+    plain_text(*args, **named_args)
   end
 
   private def head(status : Int32) : Lucky::TextResponse
-    send_response(body: "", content_type: "", status: status)
+    send_text_response(body: "", content_type: "", status: status)
   end
 
   private def head(status : HTTP::Status) : Lucky::TextResponse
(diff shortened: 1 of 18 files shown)


src/lucky/renderable.cr
```diff
@@ -142,31 +142,43 @@ module Lucky::Renderable
     file(path, content_type, disposition, filename, status.value)
   end
 
+  private def send_response(body : String, content_type : String, status : Int32? = nil) : Lucky::TextResponse
```
paulcsmith: Maybe rather than `send_response` we could do `text_response` so it is clear it is for text_responses? What do you think?

src/lucky/renderable.cr
```diff
@@ -142,31 +142,43 @@ module Lucky::Renderable
...
     head(status.value)
   end
 
   private def json(body, status : Int32? = nil) : Lucky::TextResponse
-    Lucky::TextResponse.new(context, "application/json", body.to_json, status)
+    send_response(body.to_json, "application/json", status)
+  end
+
+  private def json(body, status : HTTP::Status) : Lucky::TextResponse
+    json(body, status: status.value)
```
paulcsmith: It looks like this one does not call `to_json` on the body, but probably should. Maybe there is a missing spec for it?

src/lucky/renderable.cr
jwoertink: I feel like `text_response` might get confused with `text` which is a response.... Also, while I'm in there, what's [this method](https://github.com/luckyframework/lucky/blob/master/src/lucky/renderable.cr#L153) for? Can we deprecate it? or is there a specific use?

src/lucky/renderable.cr
jwoertink: since `json` isn't typed, it just passes the `body` to the previous def `json` which calls `to_json`. If we call it here too, then it would `to_json.to_json`.

src/lucky/renderable.cr
paulcsmith: Yeah that's the old method. Can be deprecated. 

And yeah that could be confusing but `send_response` is kind of confusing since it doesn't send anything. So I'm not totally sure what to do here. Maybe we rename `text` to `plain_text` and then have `text_response`? I imagine that *might* clear things up but I'm not sure

jwoertink: How about `send_text_response`? I would be ok renaming `text` too only because we use `text` on Pages, and I don't know if people ever confuse those too.

paulcsmith: Yeah I like that! `send_text_response` and `text` -> `plain_text` to make it clearer

paulcsmith: Also if renaming `text` -> `plain_text` the generator for `lucky gen.action.browser` needs to use change too

src/lucky/renderable.cr
```diff
@@ -142,43 +142,44 @@ module Lucky::Renderable
...
-    send_response(body, "text/plain", status)
+  private def plain_text(body : String, status : Int32? = nil) : Lucky::TextResponse
+    send_text_response(body, "text/plain", status)
   end
 
-  private def text(body : String, status : HTTP::Status) : Lucky::TextResponse
-    text(body, status: status.value)
+  private def plain_text(body : String, status : HTTP::Status) : Lucky::TextResponse
+    plain_text(body, status: status.value)
   end
```
paulcsmith: Just one thing. Could you add a method that tells people this has been renamed?

```
def text(*args, **named_args)
  {% raise "'text' in actions has been renamed to 'plain_text'" %}
end
```

PR #839, opened 2019-07-02
Update spec to fail for with_defaults with append_class

skojin:
Update spec to fail for with_defaults with append_class when no default class

error
```
double splatting a union (NamedTuple(field: Avram::PermittedAttribute(String)) | NamedTuple(field: Avram::PermittedAttribute(String), class: String)) is not yet supported
```


paulcsmith:
Ah! So it needs a default. Thank you for tracking that down. I really appreciate it

paulcsmith:
Thanks for reporting this @skojin. Fix is in https://github.com/luckyframework/lucky/pull/842

PR #840, opened 2019-07-05
passing in context to forgery meta tags. 

### LINKED ISSUE #831: csrf_meta_tags issue with context
The `Shared::LayoutHead` component needs the `context` when it's mounted; however, if you look at the component class, there's no mention of actually using the `@context`, so at first glance it seems it's not needed. 

Then if you remove the needs, and don't include it, you get an error saying it can't determine the type of `@context` when passing to [this method](https://github.com/luckyframework/lucky/blob/master/src/lucky/tags/forgery_protection_helpers.cr#L23). 

This can lead to some confusion (as it did for me just now when migrating to new class based components from older modules). My first thought was "oh, components need to have context passed always?" Then I read the error a little more and dug into the code and saw it's because the `csrf_meta_tags` method requires it. Maybe this is just documentation? Maybe we pass context in to that method? Or maybe it's just me and not really a thing? lol

### LINKED ISSUE #836: Pass in context to both CSRF methods
https://github.com/luckyframework/lucky/blob/master/src/lucky/tags/forgery_protection_helpers.cr should have the context passed in instead of using the instance var directly. See https://github.com/luckyframework/lucky/issues/831

jwoertink:
## Purpose
fixes #836 (I looked at the wrong one. #831 was already closed 😛 )

## Description
This PR fixes a little confusion. The `Shared::LayoutHead` component needs the context when it's mounted; however, if you look at the component class, there's no mention of actually using the @context, so at first glance it seems it's not needed. By passing it in to the method, we expose where it's used so it doesn't seem so "magical".

## Checklist

### DIFF
--- a/src/lucky/tags/forgery_protection_helpers.cr
+++ b/src/lucky/tags/forgery_protection_helpers.cr
@@ -5,10 +5,10 @@ module Lucky::ForgeryProtectionHelpers
   # `Lucky::FormHelpers#form_for`. It creates a hidden input with the CSRF
   # token. THis ensures that the form is safe. If you try to submit a form
   # without a CSRF token it will fail with a 403 forbidden status code.
-  def csrf_hidden_input
+  def csrf_hidden_input(context : HTTP::Server::Context)
     input type: "hidden",
       name: ProtectFromForgery::PARAM_KEY,
-      value: ProtectFromForgery.get_token(@context)
+      value: ProtectFromForgery.get_token(context)
   end
 
   # Meta tags used for submitting AJAX links and forms
@@ -16,10 +16,10 @@ module Lucky::ForgeryProtectionHelpers
   # These tags are automatically added to MainLayout when generating a new
   # project. They are used by Rails UJS to safely submit forms and non-GET AJAZ
   # requests
-  def csrf_meta_tags
+  def csrf_meta_tags(context : HTTP::Server::Context)
     meta name: "csrf-param",
       content: ProtectFromForgery::PARAM_KEY
     meta name: "csrf-token",
-      content: ProtectFromForgery.get_token(@context)
+      content: ProtectFromForgery.get_token(context)
   end
 end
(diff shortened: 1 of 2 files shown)


src/lucky/tags/forgery_protection_helpers.cr
```diff
@@ -5,21 +5,21 @@ module Lucky::ForgeryProtectionHelpers
   # `Lucky::FormHelpers#form_for`. It creates a hidden input with the CSRF
   # token. THis ensures that the form is safe. If you try to submit a form
   # without a CSRF token it will fail with a 403 forbidden status code.
-  def csrf_hidden_input
+  def csrf_hidden_input(context : HTTP::Server::Context)
```
paulcsmith: It looks like this will need to be changed in https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/src/lucky/tags/form_helpers.cr#L9

which also makes me wonder if we should pass in the context for this. That would mean that we'd need to pass in the context to `form_for` so that it could then get passed to `csrf_hidden_input`. That seems like a bit of a pain. Any ideas? Maybe we leave this one as it was?

jwoertink: hmm.. Well my first concern is why didn't this get caught when I ran the specs? Shouldn't the compiler have thrown a fit? Secondly, that does make things a bit of a pain. My first thought was just pass in `@context`, but then it becomes magical again just in a different spot. We could leave it off, but that's also a little confusing. 3 months down the road someone will look at the source code and assume that's a mistake. I'll forget, and then merge that mistake lol.

jwoertink: Maybe I inadvertently created an issue that wasn't really an issue because I was reading too much in to the code? 🤔 We could just close this PR, and wait for someone else to bring it up?

paulcsmith: I'm not sure why it didn't catch it locally. Travis caught it so that is super weird. Maybe some caching issue?

And yeah this is a tricky one...I'm almost thinking we keep things as-is so that the API is similar.

Another thought though...we could see what it would take to always pass the `context` to all components. I'm not sure how that'd work or if that'd be a pain to test but that might be a way around it too.

paulcsmith: Yeah let's keep it as-is for now. But if we can think of a good way to clean this up w.o introducing other issues I'm down!

PR #858, opened 2019-07-31
Upgrade to Crystal v0.30.0

bcardiff:
A `Crystal::VERSION` check around URI.encode_www_form should be added if you want lucky to be warning free but also keep working in 0.29.0.

You can double-check in crystal-lang/crystal#7997 which encode/decode method applies better.

A couple of dependencies need to be updated also as stated in https://github.com/bcardiff/lucky/commit/053b5591493bcc71b7328b85e7dd6134e931b807 after each of them are released. Do not use my forks as dependencies.

- [ ] https://github.com/luckyframework/lucky_cli/pull/379
- [x] https://github.com/luckyframework/avram/pull/176
- [x] https://github.com/paulcsmith/cry/pull/2
- [x] https://github.com/luckyframework/dexter/pull/8

There is one fix I notice along the way regarding `distance_in_days`.

Feel free to use as-is or close and use as a reference.

Thanks for all the specs that helped us discovering https://github.com/crystal-lang/crystal/issues/8020 before releasing.

### DIFF
--- a/spec/lucky/error_handling_spec.cr
+++ b/spec/lucky/error_handling_spec.cr
@@ -11,7 +11,7 @@ end
 private class ShortAndStoutError < Exception
   include Lucky::HttpRespondable
 
-  def http_error_code
+  def http_error_code : Int32
     418
   end
 end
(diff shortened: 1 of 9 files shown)


jwoertink:
Awesome! Thanks for all your efforts!

paulcsmith:
Thanks :D I removed the CLI and other mosop deps from paulcsmith/cry since we weren't using those libs anyway.

Rebased and added a new commit in #859

PR #869, opened 2019-08-10
Revamp mime type negotiation and verification 

### LINKED ISSUE #822: Change how we handle Accepts and Content-Type
Right now the param parser correctly parses the Content-Type https://github.com/luckyframework/lucky/blob/b5d6056dec881bdc9400bab35890af97f19e3956/src/lucky/params.cr#L384-L394

But when check what the request wants in return we incorrectly use Content-Type but we should instead use `Accept`.

I think we can keep the same `html?|json?|etc` methods, but they should all use the `Accept` header instead of `Content-Type`

## Prior art

* https://blog.bigbinary.com/2010/11/23/mime-type-resolution-in-rails.html
* https://til.hashrocket.com/posts/02e3ee8fd1-how-rails-responds-to-
* https://hexdocs.pm/phoenix/1.4.9/Phoenix.Controller.html#accepts/2
* https://github.com/rails/rails/blob/d1ffe59ab5fd6e811833c127d43b32e87b5d7131/actionpack/lib/action_dispatch/http/mime_negotiation.rb
* https://webmasters.stackexchange.com/questions/31212/difference-between-the-accept-and-content-type-http-headers
*  [List of default Accept values - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values)
* https://blog.arkency.com/2016/03/creating-new-content-types-in-rails-4-dot-2/

paulcsmith:
**Merge AFTER v0.17.0 is out**. We already have a big release. Let's not add one more thing at the last minute :P

Closes #822 

A lot of the info is in #822, but the gist is that Lucky was incorrectly checking the content-type header when it should have checked the accept header.

This goes a bit further:

* Allows verifying that the action accept the given format
* Sets a default format if none is provided
* Lots of (hopefully) helpful errors.

Once this is merged we'll need to add 'accepted_formats' in LuckyCLI's `BrowserAction` and `ApiAction`

### DIFF
--- a/spec/lucky/mime_type_spec.cr
+++ b/spec/lucky/mime_type_spec.cr
@@ -1,0 +1,62 @@
+require "../spec_helper"
+
+include ContextHelper
+
+describe Lucky::MimeType do
+  describe "known_formats" do
+    it "returns :html even though there is no mapping" do
+      Lucky::MimeType.known_formats.should contain(:html)
+    end
+  end
+
+  describe "determine_clients_desired_format" do
+    it "returns the format for the 'Accept' header if it is an AJAX request" do
+      format = determine_format("accept": "application/json", "X-Requested-With": "XmlHttpRequest")
+      format.should eq(:json)
+    end
+
+    it "returns the format for the 'Accept' header if it is not the default browser header " do
+      format = determine_format("accept": "application/json")
+      format.should eq(:json)
+    end
+
+    it "returns :ajax if it is an AJAX request and no 'Accept' header is given" do
+      format = determine_format("X-Requested-With": "XmlHttpRequest")
+      format.should eq(:ajax)
+    end
+
+    describe "when the 'Accept' header is the default browser header, and it is not an AJAX request" do
+      it "returns :html if :html is an accepted format" do
+        default_browser_header = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+        format = determine_format(default_format: :csv, headers: {"accept": default_browser_header}, accepted_formats: [:html, :csv])
+        format.should eq(:html)
+      end
+
+      it "returns the 'default_format' if :html is NOT an accepted format" do
+        default_browser_header = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+        format = determine_format(default_format: :csv, "accept": default_browser_header)
+        format.should eq(:csv)
+      end
+    end
+
+    it "falls back to 'default_format' if no accept header and not AJAX" do
+      format = determine_format(default_format: :csv)
+      format.should eq(:csv)
+    end
+  end
+end
+
+private def determine_format(default_format = :ics, **headers)
+  determine_format(default_format, headers, accepted_formats: [] of Symbol)
+end
+
+private def determine_format(default_format, headers, accepted_formats)
+  headers = headers.to_h.transform_keys(&.to_s.as(String))
+  request = build_request
+  request.headers.merge!(headers)
+  Lucky::MimeType.determine_clients_desired_format(
+    request,
+    default_format: default_format,
+    accepted_formats: accepted_formats
+  )
+end
--- a/spec/lucky/request_type_helper_spec.cr
+++ b/spec/lucky/request_type_helper_spec.cr
@@ -2,77 +2,77 @@ require "../spec_helper"
 
 private class FakeAction
   include Lucky::RequestTypeHelpers
+  default_format :my_default_format
 
-  class_property headers : HTTP::Headers = HTTP::Headers.new
+  property context : HTTP::Server::Context = ContextHelper.build_context
+  class_property _accepted_formats = [] of Symbol
 
-  def headers
-    self.class.headers
-  end
+  delegate request, to: context
 end
 
 describe Lucky::RequestTypeHelpers do
-  it "works for JSON" do
-    set_header "Content-Type", "application/json" do
-      FakeAction.new.json?.should be_true
-    end
-
-    set_header "Content-Type", "not/json" do
-      FakeAction.new.json?.should be_false
+  it "determines the format from 'Accept' header correctly" do
+    Lucky::MimeType.accept_header_formats.each do |header, format|
+      override_accept_header header do |action|
+        action.accepts?(format).should be_true
+      end
     end
   end
 
-  it "works for HTML" do
-    set_header "Content-Type", "text/html" do
-      FakeAction.new.html?.should be_true
-    end
-
-    set_header "Content-Type", "not/html" do
-      FakeAction.new.html?.should be_false
-    end
-
-    set_header "Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" do
-      FakeAction.new.html?.should be_true
+  it "uses the default if no header is set" do
+    override_accept_header "" do |action|
+      action.accepts?(:my_default_format).should be_true
     end
   end
 
-  it "works for AJAX" do
-    set_header "X-Requested-With", "XMLHttpRequest" do
-      FakeAction.new.ajax?.should be_true
-    end
-
-    set_header "X-Requested-With", "not ajax" do
-      FakeAction.new.ajax?.should be_false
+  it "doesn't use the default if header is given" do
+    override_accept_header "application/json" do |action|
+      action.accepts?(:my_default_format).should be_false
     end
   end
 
-  it "works for XML" do
-    set_header "Content-Type", "application/xml" do
-      FakeAction.new.xml?.should be_true
-    end
-
-    set_header "Content-Type", "application/xhtml+xml" do
-      FakeAction.new.xml?.should be_true
-    end
-
-    set_header "Content-Type", "not/xml" do
-      FakeAction.new.xml?.should be_false
+  it "raises if the format is unknown" do
+    expect_raises Lucky::UnknownAcceptHeaderError do
+      override_accept_header "wut" do |action|
+        action.accepts?(:blow_up)
+      end
     end
   end
 
-  it "works for plain text" do
-    set_header "Content-Type", "text/plain" do
-      FakeAction.new.plain?.should be_true
-    end
+  it "checks if client accepts JSON" do
+    override_format :json, &.json?.should(be_true)
+    override_format :foo, &.json?.should(be_false)
+  end
 
-    set_header "Content-Type", "text/plain; charset=UTF8" do
-      FakeAction.new.plain?.should be_true
-    end
+  it "checks if client accepts HTML" do
+    override_format :html, &.html?.should(be_true)
+    override_format :foo, &.html?.should(be_false)
+  end
+
+  it "checks if client accepts AJAX" do
+    override_format :ajax, &.ajax?.should(be_true)
+    override_format :foo, &.ajax?.should(be_false)
+  end
+
+  it "checks if client accepts XML" do
+    override_format :xml, &.xml?.should(be_true)
+    override_format :foo, &.xml?.should(be_false)
+  end
+
+  it "checks if client accepts plain text" do
+    override_format :plain_text, &.plain_text?.should(be_true)
+    override_format :foo, &.plain_text?.should(be_false)
   end
 end
 
-private def set_header(key, value)
-  FakeAction.headers[key] = value
-  yield
-ensure
-  FakeAction.headers = HTTP::Headers.new
+private def override_format(format : Symbol?)
+  action = FakeAction.new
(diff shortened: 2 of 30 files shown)


src/lucky/verify_accepts_format.cr
```diff
@@ -0,0 +1,78 @@
...
+
+  macro accepted_formats(formats)
+    {% if !formats.is_a?(ArrayLiteral) %}
+      {% raise "#{@type} 'accepted_formats' should be an array of Symbols. Example: [:html, :json]" %}
+    {% end %}
+
+    {% if formats.size == 1 %}
+      accepted_formats {{ formats }}, default: {{ formats.first }}
+    {% else %}
+      {% formats.raise "#{@type} must pass a default to 'accepted_formats'. Example: accepted_formats [:html, :json], default: :html" %}
```
edwardloveall: I'm confused. If this gets raised when the `formats.size` isn't `1`, wouldn't that raise on your suggestion here which has two formats?

```crystal
accepted_formats [:html, :json], default: :html
```

src/lucky/mime_type.cr
```diff
@@ -0,0 +1,94 @@
...
+        :ajax
+      else
+        default_format
+      end
+    end
+
+    private def from_accept_header(accept : String) : Symbol?
+      Lucky::MimeType.accept_header_formats.find do |accept_header_substring, format|
+        accept.includes?(accept_header_substring)
+      end.try(&.[1])
```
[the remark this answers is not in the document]
jwoertink: wouldn't `.first?` be the same as `[0]?`? If the array has 2 elements, then it would be `.last?`. But also, if that `find` block returned `nil`, then calling it directly on it would fail. 

`.try(&.[1])` isn't the "prettiest", but I think it's correct here.

spec/lucky/request_type_helper_spec.cr
```diff
@@ -2,77 +2,75 @@ require "../spec_helper"
...
-    set_header "Content-Type", "application/json" do
-      FakeAction.new.json?.should be_true
-    end
-
-    set_header "Content-Type", "not/json" do
-      FakeAction.new.json?.should be_false
+  it "determines the format from 'Accept' header correctly" do
+    Lucky::MimeType.accept_header_formats.each do |header, format|
+      override_accept_header header do |action|
+        action.accepts?(format)
```
jwoertink:
```suggestion
        action.accepts?(format).should be_true
```

spec/lucky/verify_accepts_format_spec.cr
```diff
@@ -0,0 +1,78 @@
...
+include ContextHelper
+
+private class ActionThatAcceptsHtml < Lucky::Action
+  accepted_formats [:html], default: :html
+
+  get "/test" do
+    plain_text "yay"
+  end
+
+  property clients_desired_format : Symbol = :foo
```
jwoertink: Is this supposed to be `_clients_desired_format` from the HTTP::Server::Context?

src/lucky/errors.cr
```diff
@@ -13,4 +13,47 @@ module Lucky
...
+      accept_header = request.headers["accept"]?
+      super <<-TEXT
+      Lucky couldn't figure out what format the client accepts.
+
+          The client's Accept header: '#{accept_header}'
+
+      You can teach Lucky how to handle this header:
+
+          # Add this in config/mime_types.cr
+          Lucky::MimeType.register "#{accept_header}", :custom_format
```
jwoertink: Should this also mention adding to the `accepted_formats`? Or is that not required?

jwoertink: oh, wait, I think I see. The below error would mention that for you.

src/lucky/mime_type.cr
```diff
@@ -0,0 +1,94 @@
...
+  register "application/atom+xml", :atom
+  register "application/x-yaml", :yaml
+  register "text/yaml", :yaml
+  register "text/csv", :csv
+  register "text/css", :css
+  register "text/calendar", :ics
+  register "text/plain", :plain_text
+  register "multipart/form-data", :multipart_form
+  register "application/x-www-form-urlencoded", :url_encoded_form
+
```
jwoertink: Does `:html` need to be in this list? Also, maybe `"text/javascript", :js`?

src/lucky/mime_type.cr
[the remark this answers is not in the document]
edwardloveall: Oh pfff. You're totally right. Should have woken up a bit before I reviewed this.

src/lucky/verify_accepts_format.cr
jwoertink: I think `formats` in this case is `[ [:html, :json], default: :html]`(ish)... So, if you had just `[:html, :json]` then it would fail because it's missing the `default`. However, if you passed `:html, :json` it wouldn't throw the error, but might still cause some issues. Maybe there's a better way to check this?

paulcsmith: There are 2 macros here. One has just one arg. That one can only accept a single element array. Anything else means you need to use the other macro below that sets an explicit default

Does that make sense?

edwardloveall: Ah ha! Yes I think I was thinking it would count the number of formats not the count of arguments passed. I.e. it's counting there is one argument, not one formats.

spec/lucky/request_type_helper_spec.cr
paulcsmith: Good idea :D

spec/lucky/request_type_helper_spec.cr
```diff
@@ -2,77 +2,75 @@ require "../spec_helper"
...
-    set_header "Content-Type", "not/html" do
-      FakeAction.new.html?.should be_false
+  it "uses the default if no header is set" do
+    override_accept_header "" do |action|
+      action.accepts?(:my_default_format).should be_true
     end
 
-    set_header "Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" do
-      FakeAction.new.html?.should be_true
+    # Doesn't use the default if one is given
```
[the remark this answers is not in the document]
paulcsmith: Good idea

spec/lucky/verify_accepts_format_spec.cr
paulcsmith: No this override the whole method. We're basically mocking it out but without a mocking library

src/lucky/errors.cr
paulcsmith: Yeah exactly if they don't add it they'll get the error below 👍

src/lucky/mime_type.cr
paulcsmith: `:html` is not really reliable since browser use a super weird default header (I think). It seems like what other frameworks do is they fallback to HTML if no accept header is given and :html is the default. But I will research this more to make sure we are doing this right.

paulcsmith: I like the js idea. I'll add that

src/lucky/verify_accepts_format.cr
paulcsmith: It is counting that there is one format, but *only* if you are using the macro with one argument. If you are then you can only use 1 format. If  you use the one argument macro with multiple formats then we don't want to "guess" what the correct format is, we need to know the default. If you give it an empty array of formats then we also can't guess what the default tis so you need to tell Lucky explicitly.

If you look at the other macro below it does not check the number of formats. So you can use multiple or an empty array to allow all formats

src/lucky/mime_type.cr
paulcsmith: Phoenix docs: https://hexdocs.pm/phoenix/1.4.9/Phoenix.Controller.html#accepts/2

> It is important to notice that browsers have historically sent bad accept headers. For this reason, this function will default to "html" format whenever:
>
> * the accepted list of arguments contains the "html" format
>
> * the accept header specified more than one media type preceded or followed by the wildcard media type "*/*"

I think this will work if you have `accepted_formats [:html, :json], default: :html` *but* if you do something like `accepted_formats [:html, :json], default: :json` then `:html` will never match...I wonder how to handle this case. I wonder if we should force defaulting to `:html` if you accept it? Or maybe we need to do something in Lucky where if there is a `*/*` and there is no accept header and `:html` is accepted we use that?

This PR has been hard lol

jwoertink: Yeah, I've seen the same with HTML. I think that being the default fallback is fine since this is a web framework, and html is at least the 60% use case 😂

jwoertink: Browsers will also send `text/html` inside of the accept headers. If you have default `:json`, but the accept header has `*/*` and `text/html`, then it's html.

paulcsmith: But only if you also accept `:html`. If you only accept `:json` then you'd want to default to `:json` and not assume `:html`. That's part of why this is such a PITA. So many edge cases. Lol

I think what we can do is if you have `:html` in accepted formats, you *must* have `:html` as the default. I think this makes sense anyway since if you set `:json` or something, how would a use using a browser request html instead?

It's a little weird, but I think it makes sense to just force that and raise a nice error telling people why that is if they try to default to something different. Thoughts?

jwoertink: Just to be clear, we're talking about when a request comes in that doesn't have any accept header? Like `curl http://localhost:3001/posts.json`. We have `[:html, :json]`, but if `:html` is default, then that's what would be returned here? That actually might be ok. If we do that, then [this pr](https://github.com/luckyframework/lucky/pull/778) can be re-worked a little to see that `.json` file extension, set the proper accept header, and then the issue wouldn't matter at that point.

paulcsmith: Let's say you have an `ApiAction` with `accepted_formats [:json]`

If we made it so that the browser header sets html or `text/html` sets html, then when you visit any ApiAction from the browser it would throw an error. You *could* use the trailing format, but I think as a user I would expect/want it to just give me JSON since that's the only thing it accepts.

So we can't just set `:html` format if it includes `text/html` or `*/*`. We have to fallback to the default if the header is a browser default `*/*` (in this case, :json).

However that breaks down if someone does `accepted_formats [:html, :json], default: :json` because then the browser headers will be `*/*` and would set the format to `:json`. Meaning the format would never be `:html` :( So what I think we should do is force the default to `:html` if you accept `:html`. The downside is if you curl that endpoint without a trailing format or a `application/json` you'd get back HTML. I think that is fine. You should specify if you want something other than the default

Does that make sense? Did I understand your question/thoughts correctly?

paulcsmith: I'm glad we're talking about this because it is a huge PITA and hard to get right :P

paulcsmith: Actually I'm prob overthinking this. We should just do what Phoenix does. If we have a browser default header, and the action accepts :html, then we use :html. Otherwise use the default

jwoertink: Ah, yeah. I've never had to hit an api from the browser, but that brings up a good point. It could be more than just json. You might hit a CSV endpoint from the browser and such. I don't really see another way around aside from your suggestion. Let's just roll with that for now, and if we come up with a better idea after playing with it for a bit, then we can always make updates with new releases.

paulcsmith:
Ok I believe I got everything. I also updated :html handling to mimic what Phoenix does. Will hold off on merging until after v0.17.0 is out. That way we have time to make sure we've handled all the edge cases first :D

paulcsmith:
Actually, I wonder if `accepted_formats` is a good term. Because really it means that the action responds with those list of formats. It doesn't necessarily mean it can parse that format. For example `accepted_formats [:ics]` means that the server can respond with an ics response, but it doesn't mean if you *send* an ics text that it will know what to do with it.

Does that make sense? Maybe it doesn't matter...I'm fine with this if it makes sense, but if you can think of something better, LMK

Maybe:

* `responds_with [:html, :json], default: :json`?

spec/lucky/mime_type_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+      format = determine_format("X-Requested-With": "XmlHttpRequest")
+      format.should eq(:ajax)
+    end
+
+    describe "when the 'Accept' header is the default browser header, and it is not an AJAX request" do
+      it "returns :html if :html is an accepted format" do
+        default_browser_header = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
+        format = determine_format(default_format: :csv, headers: {"accept": default_browser_header}, accepted_formats: [:html, :csv])
+        format.should eq(:html)
+      end
```
paulcsmith: Added this new spec

PR #875, opened 2019-08-23
Improve BaseHTTPClient

paulcsmith:
This can be used for request specs. This will be used for the default auth specs when generating an API app with Auth. 

I imagine this could be used for other HTTP Clients as well, but for now the main purpose is for testing Lucky apps

### DIFF
--- a/spec/lucky/base_http_client_spec.cr
+++ b/spec/lucky/base_http_client_spec.cr
@@ -1,6 +1,6 @@
 require "../spec_helper"
 
-server = TestServer.new(6226)
+server = TestServer.new(test_server_port)
 
 spawn do
   server.listen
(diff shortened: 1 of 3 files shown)


spec/lucky/base_http_client_spec.cr
```diff
@@ -10,58 +10,120 @@ Spec.before_each do
...
-
-        response.body.should eq "world"
-
-        request = server.last_request.not_nil!
-        params = request.query.not_nil!
-        params.should eq HTTP::Params.encode({"foo" => "bar"})
+  describe "headers" do
+    it "sets headers and allows chaining" do
+      with_fake_server(path: "/hello", response_body: "world") do
+        response = MyClient.new
```
jwoertink: is `response` used?

paulcsmith: It isn't! Nice catch

spec/lucky/base_http_client_spec.cr
```diff
@@ -10,58 +10,120 @@ Spec.before_each do
...
         end
       end
     end
-
   {% end %}
 end
 
+private def with_fake_server(path : String, response_body : String)
+  TestServer.route(path: path, response_body: response_body)
+  Lucky::Server.temp_config(host: "localhost", port: 6226) do
```
edwardloveall: Any reason behind the `6226` port? Could we name it in a constant? Does it need to be configurable?

spec/lucky/base_http_client_spec.cr
```diff
@@ -10,58 +10,120 @@ Spec.before_each do
...
+  describe "exec" do
+    describe "with Lucky::Action class" do
+      it "uses the method and path" do
+        with_fake_server(path: "/hello", response_body: "world") do
+          response = MyClient.new.exec(HelloWorldAction)
+
+          request = server.last_request
+          request.path.should eq "/hello"
+          request.method.should eq("POST")
+          request.body.not_nil!.gets_to_end.should eq(NamedTuple.new.to_json)
```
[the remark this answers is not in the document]
paulcsmith: The `request`  is an `HTTP::Request` from Crystal so I can't easily change that without monkey patching it. Since end users won't be checking the request body I don't think it is worth monkey patching it.

I do  like the idea about just using a string! That's a lot clearer

spec/lucky/base_http_client_spec.cr
paulcsmith: Extracted a private method. It's the port  of the test server

src/lucky/base_http_client.cr
```diff
@@ -1,28 +1,100 @@
...
     end
-    @client.get(path, headers: headers)
-  end
+  {% end %}
 
-  {% for method in [:put, :patch, :post] %}
+  # Set headers for requests
+  #
+  # ```
+  # # `content_type` will be normalized to `content-type`
```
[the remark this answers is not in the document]
paulcsmith: Headers are case insensitive! https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive So  technically it does  not capitalize them, but it does change the underscore to a hyphen. That's done  by Crystal :D 

Should  I add a note that headers are case insensitive to the docs?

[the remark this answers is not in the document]
paulcsmith: However it doesn't appear to change the underscore to a hyphen:

```
HTTP::Headers{"Foo" => "bar",
 "accept" => "text/csv",
 "accept-encoding" => "gzip, deflate",
 "content-length" => "2",
 "content_type" => "application/json",
 "host" => "localhost:6226",
 "user-agent" => "Crystal"}
```

I think most implementations of header parsers convert underscore to  hyphen  (like Crystal's does) but I think we should do it before setting just in case :)

https://github.com/crystal-lang/crystal/blob/5e6a1b672a8b06d2a819210b9905b806ec3d7c71/src/http/headers.cr#L3-L4

src/lucky/base_http_client.cr
```diff
@@ -1,28 +1,104 @@
...
+  #
+  # AppClient.new
+  #   .accept_plain_text
+  #   .get("/some-path")
+  # ```
+  def headers(**header_values)
+    @client.before_request do |request|
+      header_values = header_values
+        .to_h
+        .transform_keys(&.gsub("_", "_"))
```
edwardloveall: Did you mean

```suggestion
        .transform_keys(&.gsub("_", "-"))
```

paulcsmith: Yeah :D And I had to tweak this a bit because there was a compile error

paulcsmith:
*Now* I think it's good to go 🤣

PR #877, opened 2019-08-24
Added support for date and time input fields. 

### LINKED ISSUE #802: Support for Time fields in forms
It would be handy to have an input helper for Time fields. Currently it must be handled using a standard text input field. Being able to do this:

~~~crystal
field(f.last_updated) { |i| time_input i }
~~~

Would be ideal.

Thanks for all your hard work on Lucky, and congrats on shipping `0.15`!

jwoertink:
## Purpose
Fixes #802

## Description
This PR adds support for [date field](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) and [time field](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time)

## Checklist

### DIFF
--- a/spec/lucky/input_helpers_spec.cr
+++ b/spec/lucky/input_helpers_spec.cr
@@ -35,6 +35,15 @@ class InputTestForm
       param_key: "user"
     )
   end
+
+  def joined_at
+    Avram::PermittedAttribute(Time?).new(
+      name: :joined_at,
+      param: nil,
+      value: nil,
+      param_key: "user"
+    )
+  end
 end
 
 private class TestPage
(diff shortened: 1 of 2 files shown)


spec/lucky/input_helpers_spec.cr
```diff
@@ -224,6 +233,26 @@ describe Lucky::InputHelpers do
...
+  end
+
+  it "renders date inputs" do
+    view.date_input(form.joined_at).to_s.should contain <<-HTML
+    <input type="date" id="user_joined_at" name="user:joined_at" value="">
+    HTML
+
+    view.date_input(form.joined_at, min: "2019-01-01", max: "2019-12-31").to_s.should contain <<-HTML
+    <input type="date" id="user_joined_at" name="user:joined_at" value="" min="2019-01-01" max="2019-12-31">
+    HTML
```
paulcsmith: I think this should also have a spec for when a value is on the attribute. So we know that the HTML value is set correctly. Right now. This may mean implementing a custom version of the method but I’m not sure

Issue #878, opened 2019-08-25
Bash in built-in tasks ?


waghanza:
Hi @paulcsmith,

The `shebang` in 
https://github.com/luckyframework/lucky/blob/95361062b9ec143356add4714d12e9eff6005e61/script/precompile_tasks#L1
assume `bash` interpreter is installed

This break compilation, for example on **alpine** ?

Are `bash` feature really necessary here ?

Regards,

jwoertink:
Hey, thanks for reporting!

We actually did try to remove the bashisms in a few of the projects [ref1](https://github.com/luckyframework/lucky_cli/pull/362#issuecomment-514257906), [ref2](https://github.com/luckyframework/avram/pull/142). Taking a look at that script, it looks like there's no pipes being used, so the `pipefail` is pointless in this case; however, there are other cases where it's not. 

It's possible to replace the pipefail with a bashless alternative, but it leaves us open to possible issues with regards to needing a way to run specs against custom shell functions that are complex. 

I would love to be fully shell compliant on all scripts. If you have some suggestions on how we can do this, we'd love to hear it.

j8r:
That's not that problematic, usually `set  -eu` is fine. There are mitigations like checking the return value of the pipe command.
If you really want to know the exit core of cmd1 in `cmd1 | cmd2`, there are POSIX alternatives.

jwoertink:
@j8r The only POSIX alternative to `-o pipefail` I found was in [this answer](https://stackoverflow.com/questions/13084352/how-to-implement-set-o-pipefail-in-a-posix-way-almost-done-expert-help-nee), but that's not a solution we want to implement. Do you happen to know a better alternative? We'd like to move it over on all our scripts, but so far that's the only major thing that's holding us back. 

Thanks for the help!

j8r:
No @jwoertink, sorry, What scripts really need `pipefail`?
In my experience, this can often be mitigated by other means, like checking the content of the variable. And also this is not rare that a failing command make the one after the pipe fails, too.

PR #879, opened 2019-08-25
Make shell scripts POSIX-compliant

j8r:
## Purpose
Fixes https://github.com/luckyframework/lucky/issues/878
It converts bash scripts to POSIX compliant ones, in order to run them in any other OS without bash like Alpine (even if we can install it).

## Description
Bashisms were removed or converted to POSIX-compliant syntax.
I have also added minor improvements, mainly stylistic (but not opinionated). I can explain if there are questions about them.

## Checklist

### DIFF
--- a/script/precompile_tasks
+++ b/script/precompile_tasks
@@ -1,13 +1,9 @@
-#! /bin/bash
-
-# Exit if any subcommand fails
-set -e
-set -o pipefail
+#!/bin/sh -e
 
 # Precompile tasks if not in production and not in CI
 # Also allow skipping precompilation with an ENV var
 if [ "$LUCKY_ENV" != "production" ] && [ -z "$CI" ] && [ -z "$SKIP_LUCKY_TASK_PRECOMPILATION" ]; then
-    shards build && \
-    mkdir -p ../../bin && \
-    cp -r $(pwd)/bin $(pwd)/../..
+    shards build
+    mkdir -p ../../bin
+    cp -r $PWD/bin $PWD/../..
 fi
--- a/script/setup
+++ b/script/setup
@@ -1,8 +1,4 @@
-#! /bin/bash
-
-# Exit if any subcommand fails
-set -e
-set -o pipefail
+#!/bin/sh -eu
 
 if ! command -v docker-compose > /dev/null; then
     printf 'Docker and docker-compose are not installed.\n'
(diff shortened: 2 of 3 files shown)


script/precompile_tasks
```diff
@@ -1,13 +1,9 @@
-#! /bin/bash
-
-# Exit if any subcommand fails
-set -e
-set -o pipefail
+#!/bin/sh -e
```
edwardloveall: I assume this is doing more or less what the above two lines are doing, which is ensuring that if a line (or part of a line) fails, it will exit. I'm surprised `set -e` doesn't work as it seems to be in the POSIX standard (I believe [this](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html) is the relevant link for the flags) but it this works everywhere then I'm fine with it. 🙂

script/precompile_tasks
```diff
@@ -1,13 +1,9 @@
...
 
 # Precompile tasks if not in production and not in CI
 # Also allow skipping precompilation with an ENV var
 if [ "$LUCKY_ENV" != "production" ] && [ -z "$CI" ] && [ -z "$SKIP_LUCKY_TASK_PRECOMPILATION" ]; then
-    shards build && \
-    mkdir -p ../../bin && \
-    cp -r $(pwd)/bin $(pwd)/../..
+    shards build
+    mkdir -p ../../bin
+    cp -r $PWD/bin $PWD/../..
```
edwardloveall: Again, I thought `pwd` was a standard POSIX thing: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pwd.html but also, if it works more places, then 👍

script/precompile_tasks
edwardloveall: Oh, and I guess switching from `bash` to `sh`

script/precompile_tasks
j8r: Yes it is. This change isn't for compatibility, it just avoid to call the `pwd` binary since the `$PWD` variable already exists. That's a minor optimization I've seen.

script/precompile_tasks
j8r: We can do both yes, I was changing the shebang so I added the option there.
Did you found that `set -e` and `sh -e` are different, can you point me the line?

edwardloveall: From what I read, I think you're right that they are the same. I was asking because I wanted to know if I was missing something. You had mentioned POSIX compliance and I read that these are both compliant so I wanted to make sure I wasn't missing anything. Thanks for the confirmation 😄

PR #888, opened 2019-09-04
Add primary key and timestamps to migration

paulcsmith:
Fix migration for gen.resource.browser


paulcsmith:
Thanks for the review! I found this while working on a POC for Fusion :) We'll see how this goes

PR #890, opened 2019-09-04
Accept every integer type on the #pluralize method

### LINKED ISSUE #887: Pluralize doesn't take Int64
If you use [select_count](https://github.com/luckyframework/avram/blob/master/src/avram/queryable.cr#L142), it will return `Int64`, but passing that to `pluralize` will throw 

> no overload matches 'SomePage#pluralize' with types Int64, String

https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr#L89

My guess is we should probably also support `Int16`. Maybe we just have a few overloads that convert the type, and pass to a main method?

vnbrs:
## Purpose
This PR makes the `#pluralize` method to accept every integer subtype (BigInt, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8).

## Description
This changes the `#pluralize` method signature to `Int`.

## Checklist

Closes #887

### DIFF
--- a/spec/lucky/text_helpers/pluralize_spec.cr
+++ b/spec/lucky/text_helpers/pluralize_spec.cr
@@ -5,6 +5,7 @@ describe Lucky::TextHelpers do
     it "pluralizes words" do
       view.pluralize(1, "count").should eq "1 count"
       view.pluralize(2, "count").should eq "2 counts"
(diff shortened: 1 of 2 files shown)


paulcsmith:
Looks great! Thanks for tackling this

PR #903, opened 2019-09-16
Revamp error handler/action

### LINKED ISSUE #868: Improve error handling
Supersedes #500 

https://gist.github.com/paulcsmith/7f9099ee46f41dc3bdfaa62ab1070ebd

## Problem

* A decent amount of duplication: https://github.com/luckyframework/lucky_cli/blob/d439b9c7e3c9e8817a25cf7c5fd66e03ec7646d9/src/web_app_skeleton/src/actions/errors/show.cr.ecr#L1
* No easy way to report errors
* Sometimes  we don't want to debug certain errors. Right now there is not a built-in/easy way to do that

## Solution

```crystal
class Errors::Show < Lucky::ErrorAction
  default_format :html
  dont_report Lucky::RouteNotFoundError 
  
  # This handles a lot of the default Lucky errors, and any errors you add `Lucky::RenderableException` to
  def render(error : Lucky::RenderableException)
    if html?
      render Errors::ShowPage, status: error.status_code, title: error.error_message
    else
      json ErrorSerializer.new(message: error.error_message), status: error.status_code
    end
  end
  
  def render(error : Exception)
    if html?
      render Errors::ShowPage, status: 500, title: "Something went wrong"
    else
      json ErrorSerializer.new(message: "Something went wrong"), status: 500
    end
  end
  
  def report(error : Exception)
    # Send an email, use Rollbar, etc.
  end
end
```

## `Lucky::RenderableException`

You can extend classes with this module to automatically have Lucky handle them

```crystal
class MyException
  include Lucky::RenderableException

  def status_code
    418
  end
  
  def error_message
    "I am a teapot!"
  end
end
```

paulcsmith:
Part of #868. See that issue for details


spec/lucky/error_handling_spec.cr
```diff
@@ -2,80 +2,137 @@ require "../spec_helper"
...
+    setting output : IO = IO::Memory.new
   end
-end
 
-private class InvalidParam < Lucky::Exceptions::InvalidParam
-end
-
-private class FakeErrorAction < Lucky::ErrorAction
-  def handle_error(error : FakeError) : Lucky::Response
+  def render(error : CustomError) : Lucky::Response
```
paulcsmith: Renamed `handle_error` to `render`. I think this is simpler and reads better

spec/lucky/error_handling_spec.cr
```diff
@@ -2,80 +2,137 @@ require "../spec_helper"
...
     head status: 404
   end
 
-  def handle_error(error : Exception) : Lucky::Response
-    plain_text "Oops"
+  def render(error : Exception) : Lucky::Response
+    plain_text "This is not a debug page", status: 500
+  end
+
+  def report(error : Exception)
```
paulcsmith: Added an abstract `report` method for reporting methods however you want!

src/lucky/error_handler.cr
```diff
@@ -16,28 +17,9 @@ class Lucky::ErrorHandler
...
-      status: 500,
-      content_type: "text/html",
-      body: Lucky::ExceptionPage.for_runtime_exception(context, error).to_s
-    )
-  end
-
   private def call_error_action(context : HTTP::Server::Context, error : Exception) : HTTP::Server::Context
-    status_code = status_code_by_error(error)
-    context.response.status_code = status_code
+    settings.logger.error(exception: error.inspect_with_backtrace)
```
paulcsmith: Always log the exception

spec/lucky/error_handling_spec.cr
```diff
@@ -2,80 +2,137 @@ require "../spec_helper"
...
-describe Lucky::ErrorHandler do
-  it "does nothing if no errors are raised" do
-    error_handler = Lucky::ErrorHandler.new(action: FakeErrorAction)
-    error_handler.next = ->(_ctx : HTTP::Server::Context) {}
+describe "Error handling" do
+  describe "reporting" do
+    it "calls the report method on the error action" do
+      io = IO::Memory.new
+      FakeErrorAction.temp_config(output: io) do
+        handle_error(error: CustomError.new) do |_context, _output|
```
jwoertink: Should this be `render` too?

jwoertink: oh, nevermind. I see this method from below.

spec/lucky/error_handling_spec.cr
jwoertink: Does this cause any issues with calling the `render` method from actions? I guess it shouldn't if their argument types are different.

paulcsmith: I think it will work (or it wouldn't compile) but this could be kind of confusing or lead to unexpected compile errors since there could be tons of different overloads. I think for now I'll rename it  `render_error`

I've been thinking of renaming `render` to `html` or something. Since everything else (like `json`) has the type in the name. E.g. `html Users::IndexPage, users: UserQuery.new`

jwoertink: I like the idea of using `html`. It's an easy change.

paulcsmith: Done! I like it. And should be a pretty easy thing to upgrade

PR #905, opened 2019-09-16
Rename render -> html

### LINKED ISSUE #904: Rename render -> html
`render` is only ever used for HTML and it is the only rendering method that says nothing about its type.

Let's rename it to `html`

paulcsmith:
Closes #904

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -1,9 +1,9 @@
 module Lucky::Renderable
   # Render a page and pass it data
   #
-  # `render` is used to pass data to a page. Each key/value pair must match up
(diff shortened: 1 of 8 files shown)


src/lucky/renderable.cr
```diff
@@ -43,7 +43,7 @@ module Lucky::Renderable
   #   end
   # end
   # ```
-  macro render(page_class = nil, **assigns)
```
jwoertink: Should we have this anyway to just raise an exception to say it's been renamed?

paulcsmith: Good point :)

PR #917, opened 2019-09-18
v0.18.0 UPGRADE_NOTES

jwoertink:
## Purpose
Update UPGRADE_NOTES for the next release

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -85,7 +85,7 @@ brew upgrade lucky
       json ErrorSerializer.new(message: message, details: details, param: param), status: status
     end
 
-    private def report(error : Exception)
+    private def report(error : Exception) : Nil
       # Send to Rollbar, send an email, etc.
     end
   end
@@ -106,6 +106,7 @@ brew upgrade lucky
   end
   ```
 </details>
+- Add: `require "src/serializers/base_serializer"` to your `src/app.cr`
 - Optional: Update all serializers to inherit from `BaseSerializer`. Also merge Show/Index serializers in to a single file now.
 <details>
   <summary>src/serializers/</summary>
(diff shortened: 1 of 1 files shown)


UPGRADE_NOTES.md
```diff
@@ -1,3 +1,97 @@
...
+  ```crystal
+  abstract class BaseSerializer < Lucky::Serializer
+    def self.for_collection(collection : Enumerable, *args, **named_args)
+      collection.map do |object|
+        new(object, *args, **named_args)
+      end
+    end
+  end
+  ```
+</details>
```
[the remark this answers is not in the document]
paulcsmith: Also worth noting that is optional. They can keep them as-is but may want to convert them to match the newest style (and because it is just way nicer to use :P)

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,132 @@
+## Upgrading from 0.17 to 0.18
+
+- Ensure you've upgraded to crystal ????
```
jwoertink: This depends on if we release this before the next version of crystal. This current release doesn't require the next crystal (yet).

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,151 @@
...
+    end
+
+    private def report(error : Exception)
+      # Send to Rollbar, send an email, etc.
+    end
+  end
+  ```
+</details>
+- Rename: `title` to `message` in `src/pages/errors/show_page.cr`.
+- Add: `BaseSerializer` to `src/serializers/`.
```
paulcsmith: Also just realized  you need to require  this in src/app.cr

https://github.com/luckyframework/lucky_cli/blob/e3a98ed6d39663c24b82a0d8d71d1758cfdede84/src/web_app_skeleton/src/app.cr.ecr#L16

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,151 @@
...
+    private def error_html(message : String, status : Int)
+      context.response.status_code = status
+      html Errors::ShowPage, message: message, status: status
+    end
+
+    private def error_json(message : String, status : Int, details = nil, param = nil)
+      json ErrorSerializer.new(message: message, details: details, param: param), status: status
+    end
+
+    private def report(error : Exception)
```
paulcsmith: This needs  a  return type  of `Nil` to remove warnings on 0.31.1

PR #918, opened 2019-09-23
Update to Crystal 0.31.0

bcardiff:
Commits are self-descriptive.
The changes are compatible with Crystal 0.30.0 and 0.31.0.

The workaround for the http server lifecycle is a compromise. I would suggest to handle the lifecycle of the server in a different way. There are changes in how specs are run in Crystal 0.31.0 that prevent the former `at_exit` to work as it used to. Ref: https://github.com/crystal-lang/crystal/pull/8125#issuecomment-533296357

### DIFF
--- a/spec/lucky/base_http_client_spec.cr
+++ b/spec/lucky/base_http_client_spec.cr
@@ -128,6 +128,33 @@ private def test_server_port
   6226
 end
 
-at_exit do
-  server.close
-end
+# There is no hook for after the specs.
+# As a workaround we can just let the server be closed when the program finished
+# or monkey match Spec.run
+#
+# at_exit do
+#   server.close
+# end
+
+# monkey patch that requires renaminc server variable to SERVER constant
+#
+# private def spec_after_all
+#   puts "spec_after_all"
+#   SERVER.close
+# end
+
+# module Spec
+#   def self.run
+#     start_time = Time.monotonic
+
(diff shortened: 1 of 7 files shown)


spec/lucky/base_http_client_spec.cr
```diff
@@ -128,6 +128,33 @@ private def test_server_port
...
+
+# module Spec
+#   def self.run
+#     start_time = Time.monotonic
+
+#     at_exit do
+#       run_filters
+#       root_context.run
+#     ensure
+#       spec_after_all
```
jwoertink: @bcardiff I tried to use this patch in another project, but it blows up with this same error https://github.com/crystal-lang/crystal/issues/8228

```crystal
SERVER = TestServer.new(3002)

private def spec_after_all
  LuckyFlow.shutdown
  SERVER.close
end

module Spec
  def self.run
    start_time = Time.monotonic

    at_exit do
      run_filters
      root_context.run
    ensure
      spec_after_all
      elapsed_time = Time.monotonic - start_time
      root_context.finish(elapsed_time, @@aborted)
      exit 1 unless root_context.succeeded && !@@aborted
    end
  end
end
```

paulcsmith:
Closed by https://github.com/luckyframework/lucky/pull/926 which uses 0.31.1's new `Spec.after_suite! :D

PR #919, opened 2019-09-24
[DO NOT MERGE] Jaw/v0.18.0

jwoertink:
Basically #918 but branched from 85ccd15393aa17324e56c4b9bdc80b40ac35978a

### DIFF
--- a/src/lucky/version.cr
+++ b/src/lucky/version.cr
@@ -1,3 +1,3 @@
 module Lucky
-  VERSION = "0.17.0"
+  VERSION = "0.18.0"
 end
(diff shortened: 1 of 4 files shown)


src/lucky/version.cr
```diff
@@ -1,3 +1,3 @@
 module Lucky
-  VERSION = "0.17.0-rc1"
+  VERSION = "0.17.0"
```
paulcsmith: Should this be 0.18?

jwoertink: 🤦‍♂

PR #932, opened 2019-10-04
Add some more sample code to upgrade notes

paulcsmith:


paulcsmith:
I also removed the details/summary because it was messing up the formatting and was kind of hard to read because they weren't indented like lists

PR #936, opened 2019-10-07
Add a few details to upgrade notes for 0.18

oneiros:
I just updated an app to 0.18 using the upgrade notes and noticed some missing info. I tried to incorporate what I think was missing here.

1. I added a few missing version requirements. Updating only lucky is not enough.
2. I tried to render the info about the new `require` for `BaseSerializer` more precisely. I am pretty sure the order of the statements matter. Also, existing `require`s looked a bit differently, so I decided to stay consistent.
3. I had to make a few updates to spec helper files. It actually took me a long time to figure out why specs were failing, so hopefully this is helpful to others.


paulcsmith:
Gonna merge this and figure out the CI issues later

Issue #937, opened 2019-10-08
Make `lucky exec -r -b 1` the default


paulcsmith:
Make it easy to re-run the same file and modify it. Easier to make that the default and delete the code, than have to remember to use those args

https://gitter.im/luckyframework/Lobby?at=5d9b99f9eac5612d22d0f581

PR #938, opened 2019-10-09
changed defaults for lucky exec to make it work more like a repl

bdtomlin:
## Purpose
To make lucky exec work more like a repl by changing the default options.

## Description
Addresses #937.

Note: the print_help_or_call method is overridden because because it hijacks the help message.

## Checklist

### DIFF
--- a/spec/tasks/exec_spec.cr
+++ b/spec/tasks/exec_spec.cr
@@ -3,7 +3,7 @@ require "../spec_helper"
 describe Lucky::Exec do
   it "runs the editor" do
     with_test_template do
-      Lucky::Exec.new.call(["-e", %(echo '5 + 5' >)])
+      Lucky::Exec.new.call(["-o", "-e", %(echo '5 + 5' >)])
 
       newest_code.should eq <<-CODE
       5 + 5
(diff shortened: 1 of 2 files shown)


tasks/exec.cr
```diff
@@ -12,21 +12,25 @@ class Lucky::Exec < LuckyCli::Task
     setting template_path : String = "#{__DIR__}/exec_template.cr.template"
   end
 
+  def print_help_or_call(args : Array(String), io : IO = STDERR)
```
paulcsmith: Nice catch. Could you add  a comment about why this is here so people in the future know what's up?

spec/tasks/exec_spec.cr
```diff
@@ -3,7 +3,7 @@ require "../spec_helper"
 describe Lucky::Exec do
   it "runs the editor" do
     with_test_template do
-      Lucky::Exec.new.call(["-e", %(echo '5 + 5' >)])
+      Lucky::Exec.new.call(["-o", "-e", %(echo '5 + 5' >)])
```
paulcsmith: Just to make these options clearer, could you switch these to the full args?

```suggestion
      Lucky::Exec.new.call(["--once", "--ediitor", %(echo '5 + 5' >)])
```

tasks/exec.cr
bdtomlin: Good idea. Done.

spec/tasks/exec_spec.cr
bdtomlin: Yep, that's better. I didn't use the suggested change due to a typo, but the update has been made.

PR #939, opened 2019-10-09
Adjust lucky exec comments

paulcsmith:
There is no config/exec.cr file (yet) so I removed it

Added a few comments on general usage too


paulcsmith:
Would love some thoughts on wording or if there is something else that would be useful to add

PR #941, opened 2019-10-09
Explain when code runs

paulcsmith:


tasks/exec_template.cr.template
```diff
@@ -2,6 +2,8 @@ require "../../src/app.cr"
 
 # You have access to all your app's code here.
 #
+# This fill will be executed when the editor is closed.
```
jwoertink: Is the `fill` a typo? Or does that mean something?

paulcsmith:
Supposed to be file :P

paulcsmith:
Updated :)

PR #943, opened 2019-10-10
Add missing inline docs for time helpers

vlazar:
I've noticed time helpers are missing inline docs (#275).


paulcsmith:
Looks great! Thank you

PR #968, opened 2019-11-15
Tone down the watcher error

paulcsmith:
It took way too much attention away from the real error.

Make it subtle and less wordy.

Screen Shot 2019-11-15 at 5 56 54 PM

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -166,15 +166,16 @@ module LuckySentry
     end
 
     private def print_error_message
-      puts "There was a problem compiling. Watching for fixes...".colorize(:red)
       if successful_compilations.zero?
         puts <<-ERROR
 
-        Try this...
+        #{"---".colorize.dim}
 
-          #{green_arrow} If you haven't done it already, run #{"script/setup".colorize(:green)}
-          #{green_arrow} Run #{"shards install".colorize(:green)} to ensure dependencies are installed
-          #{green_arrow} Ask for help in #{"https://gitter.im/luckyframework/Lobby".colorize(:green)}
+        Feeling stuck? Try this...
+
+          ▸  Run setup: #{"script/setup".colorize.bold}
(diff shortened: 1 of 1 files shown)


tasks/watch.cr
```diff
@@ -166,15 +166,16 @@ module LuckySentry
     end
 
     private def print_error_message
-      puts "There was a problem compiling. Watching for fixes...".colorize(:red)
```
paulcsmith: Yeah I know. There's a giant compiler error 🤣

tasks/watch.cr
```diff
@@ -166,15 +166,16 @@ module LuckySentry
...
       if successful_compilations.zero?
         puts <<-ERROR
 
-        Try this...
+        #{"---".colorize.dim}
```
paulcsmith: Divider so it is separate from the Crystal compile error

tasks/watch.cr
```diff
@@ -166,15 +166,16 @@ module LuckySentry
...
 
-          #{green_arrow} If you haven't done it already, run #{"script/setup".colorize(:green)}
-          #{green_arrow} Run #{"shards install".colorize(:green)} to ensure dependencies are installed
-          #{green_arrow} Ask for help in #{"https://gitter.im/luckyframework/Lobby".colorize(:green)}
+        Feeling stuck? Try this...
+
+          ▸  Run setup: #{"script/setup".colorize.bold}
+          ▸  Reinstall shards: #{"rm -rf lib bin && shards install".colorize.bold}
```
paulcsmith: Sometimes cached binaries and shards cause issues during Crystal upgrades

tasks/watch.cr
jwoertink: ooo nice

paulcsmith:
Here's what it used to look like

Screen Shot 2019-11-15 at 6 03 27 PM

Issue #978, opened 2019-11-25
generate_needy_initializer does not allow a default value of false


wezm:
I tried to define a field:

```crystal
  needs required : Bool = false
```

but I get the error:

```
There was a problem expanding macro 'generate_needy_initializer'

Code in macro 'finished'

 1 | generate_needy_initializer
     ^
Called macro defined in lib/lucky/src/lucky/html_builder.cr:46:3

 46 | macro generate_needy_initializer

Which expanded to:

 > 15 |
 > 16 |
 > 17 |            @required : Bool,
                   ^
Error: argument must have a default value
```

It would appear that this line does not handle `false`:

https://github.com/luckyframework/lucky/blob/33606a60927df013f997f19dca3dc88439cf46e9/src/lucky/html_builder.cr#L52

bdtomlin:
@jwoertink I have a few minutes this afternoon if you would like a PR for this. Just want to make sure no one else is working on it since it's a quickie:)

jwoertink:
@bdtomlin Have at it! I'm definitely not working on a solution for this, so that'd be great.

paulcsmith:
This has been on my todo list for a long time so your help would be greatly appreciated @bdtomlin :D

wezm:
Fixed by #979

Issue #804, opened 2019-06-13
Allow optional parameters in components


watzon:
Currently any `needs` parameter that gets included in a Component makes that a required parameter when using `mount`. It would be nice if there was either

- a `wants` macro that defines an optional parameter
- the ability to have it determine that any nillable `needs` parameter is optional

### Example

```crystal
needs user : User?

# or

wants user : User
```

paulcsmith:
I think `needs user : User?` is good so we don't have to introduce a new concept. This kind of matches the behavior of `getter user : User?` where the default is `nil` even though you don't need to specify it. Thank you for opening this!

watzon:
No problem :)

PR #979, opened 2019-11-25
allow false as default arg to html page

bdtomlin:
## Purpose
Bug when using needs macro with boolean and false default value #978 
Also added a few tests to keep things honest down the road.

## Description
The source of the problem: https://github.com/luckyframework/lucky/blob/33606a60927df013f997f19dca3dc88439cf46e9/src/lucky/html_builder.cr#L52
This line accounts for nil, but not for false.

## Checklist

### DIFF
--- a/spec/lucky/html_page_spec.cr
+++ b/spec/lucky/html_page_spec.cr
@@ -62,6 +62,26 @@ class InnerPage < MainLayout
   end
 end
 
+class LessNeedyDefaultsPage < MainLayout
+  needs a_string : String = "string default"
+  needs bool : Bool = false
+  needs nil_default : String? = nil
+
+  def inner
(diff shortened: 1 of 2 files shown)


spec/lucky/html_page_spec.cr
```diff
@@ -62,6 +62,26 @@ class InnerPage < MainLayout
   end
 end
 
+class LessNeedyDefaultsPage < MainLayout
```
paulcsmith: Great name :D

bdtomlin:
@paulcsmith I'll definitely take a look at #804, if not today, within the next day or so.

paulcsmith:
Awesome! Thanks @bdtomlin

Issue #154, opened 2017-10-26
Gzip files in StaticHandler


paulcsmith:

bobzoller:
working on it ^^^

paulcsmith:
@bobzoller Thanks so much. This would be an awesome addition. I'm thinking it would look something like Amber's https://github.com/amberframework/amber/blob/master/src/amber/router/pipe/static.cr#L80 implementation. We don't need the etag stuff yet, but it should be added eventually.

I think their implementation covers a lot of cases so it would likely be quite similar. I think you used `CompressHandler` in the other PR. That only compresses the HTML/JSON response (I think) so I don't think that will work, but I'm not sure. One way to try is to add the handler to https://github.com/luckyframework/web/blob/master/src/server.cr and then run it with `crystal src/server.cr` and open `localhost:8080/tasks` and see what things look like. There is a sample app in `src/app` where you can add stuff to play around with while testing out the handler to make sure it gzips correctly.

I hope that helps!

s0kil:
Will this also include Brotli compression?

jwoertink:
I'm wondering if this should just use the built in crystal handler https://github.com/crystal-lang/crystal/blob/master/src/http/server/handlers/compress_handler.cr

s0kil:
@paulcsmith Ali recently wrote a [binding to brotli algo](https://github.com/naqvis/brotli.cr)

bdtomlin:
@jwoertink, @paulcsmith,

I think the built in crystal handler is not an option for a couple of reasons.

1. I tested it in development and it works great if all of your files are tiny, but on medium sized files it completely chokes and the compression never completes.

2. Even if the performance was within reason, it would still mean that you have to recompress the file on every uncached request.

What do you think about using the [compression-webpack-plugin](https://github.com/webpack-contrib/compression-webpack-plugin)

Then, create a `Lucky::CompressHandler` that gets added to the middleware before `Lucky::StaticFileHandler`. The `Lucky::CompressHandler` would basically check for three conditions:
1. The file is js, css, or svg
2. Accept-Encoding includes gzip
3. The .gz file exists

If the conditions are met, we just read the .gz file and send it.

If none of those conditions are met it just falls through to the current static file handler.

I'd love to hear more of your thoughts on this.

I have a working prototype stuffed in my `app_server.cr` file right now. I just copy/pasted Crystal's StaticFileHandler and did some quick mods to make it work:

```crystal
class Lucky::CompressHandler
[...]

PR #982, opened 2019-12-05
Static gzip

bdtomlin:
## Purpose
Adding gzipping for static assets #154 

## Description
This is the handler to take care of serving static compressed files that were pre compressed by something like `compression-webpack-plugin`

## Checklist
* [x] - An issue already exists detailing the issue/or feature request that this PR fixes
* [x] - All specs are formatted with `crystal tool format spec src`
* [ ] - Inline documentation has been added and/or updated
* [x] - Lucky builds on docker with `./script/setup`
* [x] - All builds and specs pass on docker with `./script/test`

PR #983, opened 2019-12-06
Add gzip for assets as well as server responses

bdtomlin:
## Purpose
Gzip Lucky::TextResponse if it's enabled #154

## Description
@paulcsmith Lucky::TextResponse was definitely the right place to do this. The implementation was trivial. Once the static compression PR and this PR are merged, I will add a config file to the lucky_cli repo that enables this in new apps and contains comments that explain how it works.

## Checklist


src/lucky/text_response.cr
```diff
@@ -1,4 +1,13 @@
+{% if !flag?(:without_zlib) %}
+  require "gzip"
+{% end %}
+
 class Lucky::TextResponse < Lucky::Response
+  Habitat.create do
+    setting gzip_enabled : Bool = false
+    setting gzip_content_types : Array(String) = %w(text/html text/javascript application/json text/plain application/xml text/csv)
```
paulcsmith: I'm thinking these settings might go perfectly in https://github.com/luckyframework/lucky/blob/master/src/lucky/server.cr#L6 with some of the other settings.

Then my second suggestion is to share the `gzip_enabled` setting with the `StaticCompressionHandler` in https://github.com/luckyframework/lucky_cli/pull/442/files rather than having 2 different settings. I assume if people turn it off for one it is because they either don't want gzip for anything, or they have nginx or something else that is compressing both responses.

If enough people want them to be separate we can separate them later, but I think trying to have on setting would be great if we can get away with it

spec/lucky/text_response_spec.cr
```diff
@@ -30,13 +30,27 @@ describe Lucky::TextResponse do
...
+      Lucky::TextResponse.temp_config(gzip_enabled: true) do
+        output = IO::Memory.new
+        context = build_context_with_io(output)
+        context.request.headers["Accept-Encoding"] = "gzip"
+
+        print_response(context, status: 200, body: "some body")
+        context.response.close
+
+        context.response.headers["Content-Encoding"].should eq "gzip"
+        output.to_s.ends_with?("some body").should be_false
```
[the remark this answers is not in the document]
bdtomlin: tldr; I'm making this change :)
I sure thought the response was not always the same. I tested that initially and thought I was getting different output for the same string. 

I figured I'd go double check before officially making that claim and apparently I was wrong. I must have had a typo or something before. I tested this again a bunch of times but every time got the same output given the same string, so I have changed the test.

src/lucky/text_response.cr
bdtomlin: Yep, this sounds good. I'll also change #982 to use this setting.

I'll also expand the content types a little and use them for the static handler and text response. The static handler is currently still using file extensions.

paulcsmith: Sounds great!

bdtomlin: Is it ok to merge my changes for #982 into this PR? That would make it a little easier to deal with the shared settings as well as working with Lucky Cli.

paulcsmith: Yep! I think that's a good idea

spec/lucky/static_compression_handler_spec.cr
```diff
@@ -0,0 +1,107 @@
...
+
+describe Lucky::StaticCompressionHandler do
+  it "calls next when not enabled" do
+    context = build_context(path: PATH)
+    context.request.headers["Accept-Encoding"] = "gzip"
+    next_called = false
+
+    call_handler_with(context) { next_called = true }
+
+    next_called.should be_true
```
paulcsmith: I believe this spec and the next one should also assert that the content was not gzipped by doing something like: `context.response.headers["Content-Encoding"].should_not eq "gzip"`

spec/lucky/static_compression_handler_spec.cr
```diff
@@ -0,0 +1,107 @@
...
+
+  it "calls next when Accept-Encoding doesn't include gzip" do
+    Lucky::Server.temp_config(gzip_enabled: true) do
+      context = build_context(path: PATH)
+      context.request.headers["Accept-Encoding"] = "whatever"
+      next_called = false
+
+      call_handler_with(context) { next_called = true }
+
+      next_called.should be_true
```
paulcsmith: This one should also assert the response isn't gzipped as mentioned above

spec/lucky/static_compression_handler_spec.cr
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  it "sends NOT_MODIFIED when file hasn't been modified" do
```
paulcsmith: Nice! I love that this is handled automatically 👍

spec/lucky/text_response_spec.cr
```diff
@@ -30,13 +30,29 @@ describe Lucky::TextResponse do
...
       context.response.status_code.should eq 200
     end
+
+    it "gzips if enabled" do
+      Lucky::Server.temp_config(gzip_enabled: true) do
+        output = IO::Memory.new
+        context = build_context_with_io(output)
+        context.request.headers["Accept-Encoding"] = "gzip"
+        expected_io = IO::Memory.new
+        Gzip::Writer.open(expected_io) { |gzw| gzw.print "some body" }
```
paulcsmith: Nicely handled 👍 One suggestion I have is to move this to verify phase on line 45 so it is closer to where it is used

spec/lucky/text_response_spec.cr
```diff
@@ -30,13 +30,29 @@ describe Lucky::TextResponse do
...
+
+        print_response(context, status: 200, body: "some body")
+        context.response.close
+
+        context.response.headers["Content-Encoding"].should eq "gzip"
+        output.to_s.ends_with?(expected_io.to_s).should be_true
+      end
+    end
```
paulcsmith: I think it might be good to have a spec asserting the response is not gzipped if it doesn't have a type supported by gzip. Like if you just did `foo/bar` as the content type it should not be gzipped. What do you think? Would that be worth testing?

src/lucky/server.cr
```diff
@@ -1,8 +1,27 @@
+# Class for configuring server settings
+#
+# The settings created here can be customized in each Lucky app by modifying them in your config/server.cr
```
paulcsmith: Nice!

src/lucky/static_compression_handler.cr
```diff
@@ -0,0 +1,87 @@
+# Middleware that serves static files that have been pre-compressed.
+# There can be multiple instances and the first in the middleware stack will take precedence.
+# For example, if you want to serve brotli compressed assets for browsers that support it and
+# serve gzip assets for those that don't you would do something like this in your middleware:
+#
+# ```
```
paulcsmith: May be worth putting a note that this is in `src/app_server.cr`

src/lucky/static_compression_handler.cr
```diff
@@ -0,0 +1,87 @@
...
+  private def cache_request?(context : HTTP::Server::Context, last_modified : Time) : Bool
+    # According to RFC 7232:
+    # A recipient must ignore If-Modified-Since if the request contains an If-None-Match header field
+    if if_none_match = context.request.if_none_match
+      match = {"*", context.response.headers["Etag"]}
+      if_none_match.any? { |etag| match.includes?(etag) }
+    elsif if_modified_since = context.request.headers["If-Modified-Since"]?
+      header_time = HTTP.parse_time(if_modified_since)
+      # File mtime probably has a higher resolution than the header value.
+      # An exact comparison might be slightly off, so we add 1s padding.
```
paulcsmith: Nice comments. Thanks for the explanation

src/lucky/static_compression_handler.cr
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    context.response.content_type = content_type
+    context.response.content_length = File.size(compressed_path)
+    File.open(compressed_path) do |file|
+      IO.copy(file, context.response)
+    end
+  end
+
+  private def should_compress?(file_path, content_type, compressed_path, request_headers)
```
paulcsmith:
```suggestion
  private def should_compress?(file_path, content_type, compressed_path, request_headers) : Bool
```

src/lucky/static_compression_handler.cr
bdtomlin: Good call.

spec/lucky/text_response_spec.cr
bdtomlin: I agree. I had that spec in the static compression handler, not sure why I didn't add it here. Probably got distracted :)

bdtomlin:
@paulcsmith, all the issues have been resolved. I appreciate your patience as I learn crystal and lucky!

spec/lucky/static_compression_handler_spec.cr
paulcsmith: It looks like this one still needs the not gzip check

bdtomlin:
There we go...

PR #993, opened 2019-12-23
Sorting the type declarations from needs so the nillable ones are las…

### LINKED ISSUE #992: Error: argument must have a default value when nilable needs property precedes non-nillable properties
I have a component:

```crystal
class Shared::LayoutHead < BaseComponent
  needs page_title : String
  needs app_js : Bool
  needs admin : Bool
  needs extra_css : String?
  # This is used by the 'csrf_meta_tags' method
  needs categories : CategoryQuery

  def render
    head do
      utf8_charset
      title "#{@page_title} – Read Rust"
      css_link asset("css/app.css"), data_turbolinks_track: "reload"
      js_link asset("js/app.js"), defer: "true", data_turbolinks_track: "reload" if @app_js

      if @admin
        css_link asset("css/admin.css"), data_turbolinks_track: "reload"
        js_link asset("js/admin.js"), defer: "true", data_turbolinks_track: "reload"
      end
      meta name: "turbolinks-cache-control", content: "no-cache"
      csrf_meta_tags
      responsive_meta_tag

      @categories.each do |category|
        tag "link", href: RssFeed::Show.with(category.slug).url, rel: "alternate", title: "Read Rust - #{category.name}", type: "application/rss+xml"
      end

      meta content: "Read Rust collects interesting posts related to the Rust programming language.", name: "description"
      css_link dynamic_asset(@extra_css), data_turbolinks_track: "reload" if @extra_css
    end
  end
end
```

This previously compiled on Lucky 0.18.0. On 0.18.2 the compiler reports this error. I'm using a picture here because useful information is lost without the colour and emphasised text.

Screenshot from 2019-12-23 11-25-17

The issue is [the addition of the inferred `nil` default value](https://github.com/luckyframework/lucky/pull/980) because now there is an argument with a default value that precedes ones without, which is not allowed (see [this test in the Crystal repo](https://github.com/crystal-lang/crystal/blob/1580298f62aee5e4664d87f8a890b0204baa09a7/spec/compiler/parser/parser_spec.cr#L1515)).

Before this change there was no default value on the `extra_css` parameter so it compiled ok. It's possible to work around the error by reordering the `needs` calls so that ones that are non-nil come first but this does not seem very intuitive. Plus, the error is extremely unhelpful as it does not point to the actual file where the error is happening. It wasn't until I discovered the `--error-trace` argument to the compiler that I was able to work out what was going on. This seems like it might be something that would trip up folks new to the framework. This is was the error is without `--error-trace`:

```
There was a problem expanding macro 'generate_needy_initializer'

Code in macro 'finished'

 1 | generate_needy_initializer
     ^
Called macro defined in lib/lucky/src/lucky/html_builder.cr:46:3

 46 | macro generate_needy_initializer

Which expanded to:

 > 35 |
 > 36 |
 > 37 |            @context : HTTP::Server::Context,
                   ^
Error: argument must have a default value
```

jwoertink:
## Purpose
Fixes #992

## Description
When you use `needs` that have both nilable and required values, the nilable ones had to come last otherwise there was a crystal compile error. This PR sorts the assignable type declarations putting the nilable ones last for the initialize method.

cc. @bdtomlin @wezm 

It'd be great if y'all could look over this and see if there's any other cases this solution doesn't cover. 

## Checklist

### DIFF
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -45,6 +45,7 @@ end
 
 class PageWithDefaultsFirst
   include Lucky::HTMLPage
+  needs required : String
   needs extra_css : String?
   needs status : String = "special"
   needs title : String
@@ -60,6 +61,6 @@ describe "Assigns within multiple pages with the same name" do
     PageTwo.new build_context, title: "foo", name: "Paul"
     PageThree.new build_context, name: "Paul", admin_name: "Pablo", title: "Admin"
     PageWithQuestionMark.new(build_context, signed_in?: true).perform_render.to_s.should contain("true")
-    PageWithDefaultsFirst.new(build_context, title: "foo").perform_render.to_s.should contain("special foo")
+    PageWithDefaultsFirst.new(build_context, required: "thing", title: "foo").perform_render.to_s.should contain("special foo")
   end
 end
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -1,5 +1,7 @@
 require "../spec_helper"
 
+include ContextHelper
+
 class BasePage
   include Lucky::HTMLPage
 
@@ -41,11 +43,23 @@ class PageWithQuestionMark
   end
 end
 
+class PageWithDefaultsFirst
+  include Lucky::HTMLPage
+  needs extra_css : String?
+  needs status : String = "special"
+  needs title : String
+
+  def render
+    text "#{@status} #{@title}"
+  end
+end
+
 describe "Assigns within multiple pages with the same name" do
   it "should only appear once in the initializer" do
     PageOne.new build_context, title: "foo", name: "Paul", second: "second"
     PageTwo.new build_context, title: "foo", name: "Paul"
     PageThree.new build_context, name: "Paul", admin_name: "Pablo", title: "Admin"
     PageWithQuestionMark.new(build_context, signed_in?: true).perform_render.to_s.should contain("true")
+    PageWithDefaultsFirst.new(build_context, title: "foo").perform_render.to_s.should contain("special foo")
   end
 end
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -46,7 +46,9 @@ end
 class PageWithDefaultsFirst
   include Lucky::HTMLPage
   needs required : String
+  needs nothing : Bool = false
   needs extra_css : String?
+  needs extra_html : String? = nil
   needs status : String = "special"
   needs title : String
 
--- a/spec/lucky/component_spec.cr
+++ b/spec/lucky/component_spec.cr
@@ -9,7 +9,6 @@ private class TestComponent < Lucky::BaseComponent
 end
 
 private class ComplexTestComponent < Lucky::BaseComponent
-  needs extra_css : String?
   needs title : String
 
   def render
--- a/spec/lucky/component_spec.cr
+++ b/spec/lucky/component_spec.cr
@@ -9,6 +9,7 @@ private class TestComponent < Lucky::BaseComponent
 end
 
 private class ComplexTestComponent < Lucky::BaseComponent
+  needs extra_css : String?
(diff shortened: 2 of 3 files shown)


spec/lucky/component_spec.cr
```diff
@@ -9,6 +9,7 @@ private class TestComponent < Lucky::BaseComponent
 end
 
 private class ComplexTestComponent < Lucky::BaseComponent
+  needs extra_css : String?
```
jwoertink: Adding this line threw the compilation error.

bdtomlin:
I don't think this goes far enough. The problem is not just with nilable types, there is still a problem if you have needs macro that has a default value before a needs statement that doesn't.

In other words the following will fail just due to the order of the needs statements:

```crystal
needs some_var : String = "whatever"
needs required_var : String
```

jwoertink:
Ah, good call. Luckily the implementation is easy now, so adding on to it shouldn't be too bad.

spec/lucky/assignable_spec.cr
```diff
@@ -41,11 +43,23 @@ class PageWithQuestionMark
   end
 end
 
+class PageWithDefaultsFirst
+  include Lucky::HTMLPage
+  needs extra_css : String?
```
paulcsmith: Could you add another needs above this one without a default and not nil? I’m not sure but I think that may show an issue

src/lucky/html_builder.cr
```diff
@@ -45,8 +45,9 @@ module Lucky::HTMLBuilder
 
   macro generate_needy_initializer
     {% if !@type.abstract? %}
+      {% sorted_assigns = ASSIGNS.sort_by { |dec| dec.type.resolve.nilable? || dec.value ? 1 : 0 } %}
```
paulcsmith: I think something like this may work and ensure that false and nil defaults work too:

decl.default || decl.default == nil == decl.default == false

I’m not sure but that’s my hunch

spec/lucky/assignable_spec.cr
paulcsmith: And maybe a default of false too

jwoertink: Adding a required one above this and the spec still passes.

src/lucky/html_builder.cr
jwoertink: If the type is `thing : String?`, then there's no value at all. It doesn't seem to even equate to `nil` from what I can tell. Without the resolve, it won't compile. But I'll add in the check for it having a value of nil and false.

spec/lucky/assignable_spec.cr
paulcsmith: Ok cool! What about the false default?

src/lucky/html_builder.cr
```diff
@@ -45,8 +45,11 @@ module Lucky::HTMLBuilder
 
   macro generate_needy_initializer
     {% if !@type.abstract? %}
+      {% sorted_assigns = ASSIGNS.sort_by { |dec|
+           dec.type.resolve.nilable? || dec.value || dec.value == nil || dec.value == false ? 1 : 0
```
paulcsmith: I think this will work without the resolve now. That’d also be preferable because resolve can lead to weird loading issues like we used to have in models where you had to require models depending on alphabetic order

paulcsmith: Actually maybe not...since that would break the type union when it’s nilable with no default. So I guess this has to stay :+1: my bad

paulcsmith:
I think this looks good. Just trying to think if there’s a way to do it without resolving the type. But we can try to optimize for that later!

LGTM

bdtomlin:
If you want to avoid resolve, you could use
```crystal
dec.type.stringify.ends_with?("?")
```

spec/lucky/assignable_spec.cr
jwoertink: Cool. I added a few more just for some added comfort. They look good!

jwoertink:
Sweet. Once these checks pass, I'll merge it in!

PR #996, opened 2019-12-28
Add more context to resource generator

paulcsmith:
## Purpose

Add more context to page titles and headers. Also uses id instead of the first column since the first column may not make any sense at all for a header/page title.

## Checklist


paulcsmith:
Thanks for the review!

PR #997, opened 2020-01-06
Fixing issue on gen.resource.browser with some column types

### LINKED ISSUE #995: Error with gen.resource.browser when using JSON
when I try to make a browser resource with a JSON::Any column I get an error.

```bash
fermi% lucky gen.resource.browser Foo bar:JSON::Any
Must provide valid columns for the resource: lucky gen.resource.browser Foo name:String
```

jwoertink:
## Purpose
Fixes #995 

## Description
Prior to this, the `gen.resource.browser` task would fail if you tried to generate a `JSON::Any` column or any of the `Array` column types. We were just doing a simple split on `:` which didn't really work for `JSON::Any`. 

## Checklist

### DIFF
--- a/spec/tasks/gen/resource_spec.cr
+++ b/spec/tasks/gen/resource_spec.cr
@@ -7,7 +7,7 @@ describe Gen::Action do
   it "generates actions, model, operation and query" do
     with_cleanup do
       Gen::Migration.silence_output do
-        io = generate Gen::Resource::Browser, "User", "name:String"
+        io = generate Gen::Resource::Browser, "User", "name:String", "signed_up:Time"
 
         should_create_files_with_contents io,
           "./src/actions/users/index.cr": "class Users::Index < BrowserAction",
@@ -32,28 +32,17 @@ describe Gen::Action do
           "./src/queries/user_query.cr": "class UserQuery < User::BaseQuery",
           "./src/operations/save_user.cr": "class SaveUser < User::SaveOperation"
         should_generate_migration named: "create_users.cr"
+        should_generate_migration named: "create_users.cr", with: "add signed_up : Time"
         io.to_s.should contain "at: #{"/users".colorize.green}"
       end
     end
   end
 
-  it "allows for all supported column types" do
+  it "displays an error when given a more complex type" do
     with_cleanup do
-      io = generate Gen::Resource::Browser,
-        "Alphabet",
-        "a:Bool",
-        "b:Int16",
-        "c:Int32",
-        "d:Int64",
-        "e:String",
-        "f:UUID",
-        "g:Time",
-        "h:Float64",
-        "i:JSON::Any",
-        "j:Array(String)"
-      io.to_s.should_not contain("Must provide valid columns")
-      should_generate_migration named: "create_alphabets.cr", with: "add i : JSON::Any"
-      should_generate_migration named: "create_alphabets.cr", with: "add j : Array(String)"
+      io = generate Gen::Resource::Browser, "Alphabet", "a:JSON::Any"
+
+      io.to_s.should contain("Other complex types can be added manually")
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/tasks/gen/resource_spec.cr
```diff
@@ -4,7 +4,7 @@ include CleanupHelper
 include GeneratorHelper
 
 describe Gen::Action do
-  it "generates actions, model, form and query" do
+  it "generates actions, model, operation and query" do
```
jwoertink: Might as well change this while I'm in here 😂

spec/tasks/gen/resource_spec.cr
```diff
@@ -37,6 +37,26 @@ describe Gen::Action do
...
+        "d:Int64",
+        "e:String",
+        "f:UUID",
+        "g:Time",
+        "h:Float64",
+        "i:JSON::Any",
+        "j:Array(String)"
+      io.to_s.should_not contain("Must provide valid columns")
+      should_generate_migration named: "create_alphabets.cr", with: "add i : JSON::Any"
+      should_generate_migration named: "create_alphabets.cr", with: "add j : Array(String)"
```
jwoertink: I wonder if there's a better way that I can check all of these columns and only run the expectation once? 🤔

tasks/gen/resource/browser.cr
```diff
@@ -134,12 +134,24 @@ class Gen::Resource::Browser < LuckyCli::Task
 
   private def columns_are_valid? : Bool
     column_definitions.any? && column_definitions.all? do |column_definition|
-      column_parts = column_definition.split(":")
-      column_name = column_parts.first
+      column_parts = parse_definition(column_definition)
+      column_name = column_parts.first?.to_s
```
jwoertink: This is needed because you can do `"".split(":").first` and get nil but `([] of String).first` throws an exception 😕

tasks/gen/resource/browser.cr
```diff
@@ -134,12 +134,24 @@ class Gen::Resource::Browser < LuckyCli::Task
...
       column_parts.size == 2 && column_name == column_name.underscore
     end
   end
 
+  private def parse_definition(column_definition : String) : Array(String)
+    matcher = /^([a-z_0-9]+):(Bool|Int16|Int32|Int64|String|UUID|Time|Float64|JSON::Any|Array\(\w+\))/
```
jwoertink: Regex is never pretty, but I'm not sure a cleaner way to do this. I originally wrote one that allowed for basically any constant, but I figure this might be a little more "type-safe"(ish) in the case that someone tries to do something like `x:Uint8` which we don't support. 

An alternative to parsing the definition this way could be to not allow `JSON::Any`, and just do something like `data:JSON`. But then we have to watch for that string specifically and append the `::Any` where necessary, and also add some sort of error message if someone does the full one. I feel like that would get just as messy and confusing though.

spec/tasks/gen/resource_spec.cr
paulcsmith: Nice!

jwoertink:
I'm down for that too. I forgot about creating the inputs. I guess we can say that the generators are only meant for a basic "getting started" point, and any complex thing needs to be done manually. We should probably do up a whole deal on the guides on generators and mention that. 

Ok, so I'll change this up to check specifically for the simple types that we support, and then any others will raise an error message. Do you think the error message should say something like "If you need to use JSON or Arrays, you will need to add those in manually"?

paulcsmith:
Yeah that’s a good idea. Maybe more generic though? Like “other complex types can be added manually” and maybe a link to a guide if you can think of a good one!

paulcsmith:
Also I think split takes a second argument that’s an int. So if you pass 2 it will stop splitting after the first colon. That’s how it is in ruby and hopefully in crystal too 😆

Issue #1007, opened 2020-02-11
Allow specifying columns for gen.model


wout:
A few times already I tried generating a model with columns. Something like:

```
lucky gen.model Address street:String number:String postcode:String city:String
```

But unlike `gen.resource.browser`, they are ignored. Is this intentionally? Because it feels a bit inconsistent.

paulcsmith:
Thanks for opening the issue  This is definitely not intentional. Just hasn’t been the biggest priority but would love it if this were supported like in the other generator!

wout:
I will have a look at it one of these days. I guess the whole columns part can be moved to a mixin so we can include it in both `Gen::Model` and `Gen::Resource::Browser`.

paulcsmith:
@wout Yeah I think that would work!

PR #1009, opened 2020-02-11
Add option for columns in model generator.

wout:
## Purpose
Adds the option to define columns with the `gen.model` generator, as discussed in #1007.

## Description
Basically, I moved some code around, extracted common behaviour into a mixin and dried up some parts.

To simplify sharing, I renamed the following methods:
- `model_name` -> `subject_name`
- `resource_name` -> `subject_name` 

## Checklist

### DIFF
--- a/spec/tasks/gen/model_spec.cr
+++ b/spec/tasks/gen/model_spec.cr
@@ -17,7 +17,7 @@ describe Gen::Model do
           "./src/models/customer.cr": "table"
         should_create_files_with_contents io,
           "./src/models/customer.cr": "class Customer < BaseModel",
-          "./src/operations/save_customer.cr": "class SaveCustomer < Customer::SaveOperation",
+          "./src/operations/save_customer.cr": "class SaveCustomer < Customer::SaveOperation\nend",
           "./src/queries/customer_query.cr": "class CustomerQuery < Customer::BaseQuery"
         should_generate_migration named: "create_customers.cr"
       end
@@ -35,6 +35,9 @@ describe Gen::Model do
 
         should_create_files_with_contents io,
           "./src/models/contact_info.cr": "table"
+        should_create_files_with_contents io,
+          "./src/models/contact_info.cr": "column name : String",
+          "./src/operations/save_contact_info.cr": "permit_columns name, contacted_at"
         should_create_files_with_contents io,
           "./src/models/contact_info.cr": "class ContactInfo < BaseModel",
           "./src/operations/save_contact_info.cr": "class SaveContactInfo < ContactInfo::SaveOperation",
--- a/spec/tasks/gen/model_spec.cr
+++ b/spec/tasks/gen/model_spec.cr
@@ -8,6 +8,26 @@ describe Gen::Model do
     with_cleanup do
       Gen::Migration.silence_output do
         io = IO::Memory.new
+        model_name = "Customer"
+        ARGV.push(model_name)
+
+        Gen::Model.new.call(io)
+
+        should_create_files_with_contents io,
+          "./src/models/customer.cr": "table"
+        should_create_files_with_contents io,
+          "./src/models/customer.cr": "class Customer < BaseModel",
+          "./src/operations/save_customer.cr": "class SaveCustomer < Customer::SaveOperation",
+          "./src/queries/customer_query.cr": "class CustomerQuery < Customer::BaseQuery"
+        should_generate_migration named: "create_customers.cr"
+      end
+    end
+  end
+
+  it "generates a model with columns" do
+    with_cleanup do
+      Gen::Migration.silence_output do
+        io = IO::Memory.new
         model_name = "ContactInfo"
         ARGV.push(model_name, "name:String", "contacted_at:Time")
 
@@ -26,14 +46,12 @@ describe Gen::Model do
   end
 
   it "displays an error when given a more complex type" do
-    with_cleanup do
-      io = IO::Memory.new
-      ARGV.push("Alphabet", "a:JSON::Any")
+    io = IO::Memory.new
+    ARGV.push("Alphabet", "a:JSON::Any")
 
-      Gen::Model.new.call(io)
+    Gen::Model.new.call(io)
 
-      io.to_s.should contain("Other complex types can be added manually")
-    end
+    io.to_s.should contain("Other complex types can be added manually")
   end
 
   it "displays an error if given no arguments" do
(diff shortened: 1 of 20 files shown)


paulcsmith:
Added an issue to the website to document the new options 👍  https://github.com/luckyframework/website/issues/211

spec/tasks/gen/model_spec.cr
```diff
@@ -9,21 +9,33 @@ describe Gen::Model do
       Gen::Migration.silence_output do
         io = IO::Memory.new
         model_name = "ContactInfo"
-        ARGV.push(model_name)
+        ARGV.push(model_name, "name:String", "contacted_at:Time")
```
paulcsmith: I think there should also be a test that just passes the model name and no columns, just to make sure that works too

spec/tasks/gen/resource_spec.cr
```diff
@@ -3,7 +3,7 @@ require "../../spec_helper"
 include CleanupHelper
 include GeneratorHelper
 
-describe Gen::Action do
+describe Gen::Resource::Browser do
```
paulcsmith: Nice catch!

tasks/gen/model.cr
```diff
@@ -3,8 +3,10 @@ require "teeplate"
 require "avram"
 require "./templates/model_template"
 require "wordsmith"
+require "./mixins/migration_with_columns"
 
 class Gen::Model < LuckyCli::Task
+  include Gen::Mixins::MigrationWithColumns
```
paulcsmith: Tiny thing, but could you add a newline after this? We typically put newlines after includes

tasks/gen/model.cr
```diff
@@ -24,59 +26,52 @@ class Gen::Model < LuckyCli::Task
 
     Example:
 
-      lucky gen.model Project
+      lucky gen.model Project title:String completed:Bool priority:Int32
```
paulcsmith: Love it!

spec/tasks/gen/model_spec.cr
wout: Will do!

tasks/gen/model.cr
wout: I did initially because I love the space, but then I removed it. I will add one again.

wout:
As for `subject_name`, I was a bit on the fence about whether to give priority to `model_name` or `resource_name`. So I decided to go with something new :joy: . I'll change it to `resource_name`, as it is in fact a resource.

spec/tasks/gen/model_spec.cr
wout: Good you noticed because that didn't work. It's fixed now.

tasks/gen/mixins/migration_with_columns.cr
```diff
@@ -0,0 +1,70 @@
...
+      [] of String
+    end
+  end
+
+  private def column_arguments? : Bool
+    !!ARGV[1]?
+  end
+
+  private def columns_are_valid?(required : Bool = false) : Bool
+    (!required || column_definitions.any?) && column_definitions.all? do |column_definition|
```
paulcsmith: Rather than passing in `required` true, you could change this to remove `column_definitions.any?` and instead do `column_definitions.all?`. That will return true if there are no columns so could be used in the model generator.

Then in the resource generator there could add the check there: `columns_are_valid? && column_definitions.any?`

That way `columns_are_valid?` does just one thing and the behavior is a bit clearer. Thoughts?

wout: You're right, that is a lot easier on the eyes.

wout:
> Also, does this change the model template to generate the columns in the table definition?

No. :blush: 

But it does now, as well as the `permit_columns` line on the generated operation.

tasks/gen/templates/model/operations/save_{{underscored_name}}.cr.ecr
```diff
@@ -1,2 +1,5 @@
 class Save<%= @name %> < <%= @name %>::SaveOperation
+<%- if @columns.any? -%>
+  permit_columns <%= @columns.map(&.name).join(", ") %>
```
paulcsmith: I am worried about security here. If someone generates a model with `admin` or `encrypted_password` or something, we wouldn't want to permit those. I think since we aren't generating forms and pages, it is ok to leave this out and users can add them depending on what they need. Also they may be creating things without params, in which case they don't need to permit_columns: `SaveUser.create(name: "John", email: "John@gmail.com")

One other idea would be to include this but comment it out and add an explaining comment:

```
# In order for clients to save to a column, you must permit them:
# permit_columns ....
```

Thoughts? cc @jwoertink @edwardloveall

jwoertink: Yeah, that makes sense. You don't want to generate leaky code and have the user not know. If we don't permit them, and they try to use it, they will be told right away. Then if those fields are there and just commented out with a note, that makes it even easier because they don't need to look up docs on what to do.

paulcsmith: Right on. We could also link to the docs on it! https://luckyframework.org/guides/database/validating-saving#perma-permitting-columns

I'm thinking the comment can be included either way, and the commented out `permit_columns` can be generated if there are columns passed to the generator.

**If no columns provided**

```crystal
# To save user provided params to a column, you must permit them
# https://luckyframework.org/guides/database/validating-saving#perma-permitting-columns
#
# permit_columns column_1, column_2
```

*If columns are provided**

```crystal
# To save user provided params to a column, you must permit them
# https://luckyframework.org/guides/database/validating-saving#perma-permitting-columns
#
# permit_columns {{ the ones passed in from the generator }}
```

I'm guessing this could be cleaned up by adding a `columns_list` method to the `ModelTemplate` that is something like: 

```crystal
example_columns = [
  ColumnDefinition.new("column_1", ""), 
  ColumnDefinition.new("column_2", "")
]
(columns.any? ? columns : example_columns).map(&.name).join(", ")

wout: Makes a lot of sense! I'll attend to it tomorrow.

wout: Good morning! That's done now in both the model and the resource generators.

paulcsmith:
Looks good!

PR #1010, opened 2020-02-11
Updating how custom tags take boolean attributes

### LINKED ISSUE #1001: Boolean attributes appended without a space in custom tag
Perhaps there's a better way to achieve this, but anyway... I was supplied the following markup to use for my web statistics:

```html
<script data-goatcounter="https://readrust.goatcounter.com/count" async src="//gc.zgo.at/count.js"></script>
```

Which after a few different attempts, I translated to:

```crystal
tag "script", "async", data_goatcounter: "https://readrust.goatcounter.com/count", src: "//gc.zgo.at/count.js" do
end
```

However when that's rendered it comes out as:

```html
<script data-goatcounter="https://readrust.goatcounter.com/count" src="//gc.zgo.at/count.js"async>
```

Note that `async` is joined to the `src`.

The issue seems to be in:

https://github.com/luckyframework/lucky/blob/e866dd93cff3e4aeafcd47fc3c6219aba87dd36b/src/lucky/tags/custom_tags.cr#L32

where it appends the `boolean_attrs` to the `tag_attrs` without a space in between.

jwoertink:
## Purpose
Fixes #1001

## Description
If you used a custom tag with a boolean attribute and a block, the boolean attributes had to be passed as a string as opposed to any other form which is an array of symbols. It also made the attributes butt up against the previous attr value which could lead to issues. 

This PR pushes the bool attr building down a tag level which fixes all that. 

## Checklist

### DIFF
--- a/spec/lucky/custom_tags_spec.cr
+++ b/spec/lucky/custom_tags_spec.cr
@@ -38,6 +38,11 @@ describe Lucky::CustomTags do
         page.text "content"
       end.to_s.should contain %(<foo-tag class="my-class">content</foo-tag>)
     end
+
+    view.tap do |page|
+      page.tag "script", [:async], data_counter: "https://counter.co", src: "count.js" do
+      end.to_s.should contain %(<script data-counter="https://counter.co" src="count.js" async></script>)
+    end
   end
 
   it "has a method for empty tags" do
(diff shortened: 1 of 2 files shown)


spec/lucky/custom_tags_spec.cr
```diff
@@ -38,6 +38,11 @@ describe Lucky::CustomTags do
         page.text "content"
       end.to_s.should contain %(<foo-tag class="my-class">content</foo-tag>)
     end
+
+    view.tap do |page|
```
paulcsmith: unnamed

PR #1014, opened 2020-02-21
permit_columns when generating a resource

paulcsmith:
This was causing an error because the resource generator generates field inputs so needs these columns to be permitted

So I removed the comment for the resource generator. The model generator still leaves it uncommented since it doesn't generate any inputs.

Here's the error:

```
In src/pages/comments/edit_page.cr:14:27
522
523 14 | mount Shared::Field.new(op.title), &.text_input(autofocus: "true")
524                          ^--
525Error: no overload matches 'Shared::Field(T).new' with type Avram::Attribute(String | Nil)+
526
527
```


wout:
Ok, so do we keep them commented out for the model generator?

paulcsmith:
@wout there were 2 separate template files so the one for models remains the same :D

PR #1018, opened 2020-02-29
Allow needs with metaclass

### LINKED ISSUE #1017: needs doesn't handle Metaclass
```crystal
needs x : String.class
```

This throws an error saying `undefined method resolve for Metaclass` which is here:

https://github.com/luckyframework/lucky/blob/2327ebdcad5ada30bea3fac19c29c49f1e56ef46/src/lucky/html_builder.cr#L49

I think we can just sort metaclasses first.... unless it's possible for them to be nilable?

```crystal
dec.type.is_a?(Metaclass) || dec.type.resolve.nilable? || ....
```

paulcsmith:
Closes #1017

This was an uncaught regression when we fixed sorting of needs.
Luckily the website uses a metaclass so we caught this before announcing
the new release :)

### DIFF
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -49,12 +49,18 @@ class PageWithDefaultsFirst
   needs nothing : Bool = false
   needs extra_css : String?
   needs extra_html : String? = nil
+  needs optional_metaclass : String.class | Nil
   needs status : String = "special"
   needs title : String
 
   def render
     text "#{@status} #{@title}"
   end
+end
+
+class PageWithMetaclass
+  include Lucky::HTMLPage
+  needs string_class : String.class
 end
 
 describe "Assigns within multiple pages with the same name" do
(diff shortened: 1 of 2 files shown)


spec/lucky/assignable_spec.cr
```diff
@@ -49,6 +49,7 @@ class PageWithDefaultsFirst
   needs nothing : Bool = false
   needs extra_css : String?
   needs extra_html : String? = nil
+  needs optional_metaclass : String.class | Nil
```
paulcsmith: Needed this to make sure it sorted with a nilable metaclass

src/lucky/html_builder.cr
```diff
@@ -46,7 +46,13 @@ module Lucky::HTMLBuilder
   macro generate_needy_initializer
     {% if !@type.abstract? %}
       {% sorted_assigns = ASSIGNS.sort_by { |dec|
-           dec.type.resolve.nilable? || dec.value || dec.value == nil || dec.value == false ? 1 : 0
+           has_explicit_value =
+             dec.type.is_a?(Metaclass) ||
+               dec.type.types.map(&.id).includes?(Nil.id) ||
```
paulcsmith: Removed resolve. This makes things a bit more reliable since it doesn't need to resolve the type at this point, it can be done later

PR #1020, opened 2020-03-03
Add Lucky::Paginator

paulcsmith:
This will make it easy to paginate with Lucky. We (and others) can extend this with new goodies since the paginator is decoupled from Avram. This works very very similarly to the Ruby Pagy gem: https://github.com/ddnexus/pagy but with some modifications to link generation and the provided methods.

Still to do:

- [x] Add specs (this was all tested in an app, but need some real unit tests)
- [x] Add specs for backend helpers 
- [x] Add specs for components
- [x] Add in SimpleNav component to get up and running super fast with a built-in component
- [x] Create templates for Bulma and Bootstrap pagination links to get up and running quickly
- [x] Add `series` to help generate gaps and page windows

Probably later on:

- [ ] Paginate array helpers
- [ ] Add pagination to resource generator by default! Helpful to have and shows people how to do it
- [ ] Paginate without count helpers

### DIFF
--- a/spec/paginator/backend_helpers_spec.cr
+++ b/spec/paginator/backend_helpers_spec.cr
@@ -1,0 +1,88 @@
+require "../spec_helper"
+
+class FakeUser < Avram::Model
+  table do
+  end
+
+  def self.database
+    UnusedDatabase
+  end
+
+  class BaseQuery
+    # Override so it doesn't hit the db and we get 2 pages of results
+    def select_count
+      50
+    end
+  end
+end
+
+class Paginatable
+  include Lucky::Paginator::BackendHelpers
+  include ContextHelper
+
+  def initialize(@page : String? = nil)
+  end
+
+  def call
+    paginate(FakeUser::BaseQuery.new)
+  end
+
+  def params
+    Params.new(@page)
+  end
+
+  def context
+    build_context
+  end
+
+  class Params
+    getter page
+
+    def initialize(@page : String?)
+    end
+
+    def get?(_value) : String?
+      @page
+    end
+  end
+end
+
+class PaginatableWithOverriddenMethods < Paginatable
+  def paginator_page : Int32
+    2
+  end
+
+  def paginator_per_page : Int32
+    10
+  end
+end
+
+describe Lucky::Paginator::BackendHelpers do
+  it "uses default of page 1 and per_page of 25" do
+    pages, records = Paginatable.new.call
+
+    pages.page.should eq(1)
+    pages.per_page.should eq(25)
+    pages.total.should eq(2)
+    records.query.offset.should eq(0)
+    records.query.limit.should eq(25)
+  end
+
+  it "uses the 'page' param if given" do
+    pages, records = Paginatable.new(page: "2").call
+
+    pages.page.should eq(2)
+    pages.per_page.should eq(25)
+    records.query.offset.should eq(25)
+    records.query.limit.should eq(25)
+  end
+
+  it "allows overriding 'paginator_page' and 'paginator_per_page'" do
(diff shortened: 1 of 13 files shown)


jwoertink:
This will be awesome!

wontruefree:
Cursor based pagination is becoming more popular.  What do you think about adding cursor based pagination support or an interface to implement it yourself?

paulcsmith:
@wontruefree Definitely in the pipeline! I prefer cursor based for a lot of stuff. Doesn't work for everything, but when it does it is awesome!

wontruefree:
Agreed it does not work for everything.  Also can add tons of complexity in edge cases.  that is the reason I mentioned just an interface for it.

spec/paginator/backend_helpers_spec.cr
```diff
@@ -0,0 +1,88 @@
...
+    pages.total.should eq(2)
+    records.query.offset.should eq(0)
+    records.query.limit.should eq(25)
+  end
+
+  it "uses the 'page' param if given" do
+    pages, records = Paginatable.new(page: "2").call
+
+    pages.page.should eq(2)
+    pages.per_page.should eq(25)
```
jwoertink: Just as a safety, should we still have `pages.total.should eq(2)`?

spec/paginator/backend_helpers_spec.cr
```diff
@@ -0,0 +1,88 @@
...
+    records.query.offset.should eq(25)
+    records.query.limit.should eq(25)
+  end
+
+  it "allows overriding 'paginator_page' and 'paginator_per_page'" do
+    pages, records = PaginatableWithOverriddenMethods.new.call
+
+    pages.page.should eq(2)
+    pages.per_page.should eq(10)
```
jwoertink: Same. Maybe having `pages.total.should eq(5)`?

src/lucky/paginator/backend_helpers.cr
```diff
@@ -0,0 +1,91 @@
...
+  # ```
+  def paginate(
+    query : Avram::Queryable(T),
+    per_page : Int32 = paginator_per_page
+  ) : Tuple(Paginator, Avram::Queryable(T)) forall T
+    pages = Paginator.new \
+      page: paginator_page,
+      per_page: per_page,
+      item_count: query.clone.reset_order.reset_limit.reset_offset.select_count,
+      full_path: context.request.resource
```
jwoertink: So with this, if you need to override what your page link paths are, you'd just pass in the string? `full_path: "/whatever"`

src/lucky/paginator/paginator.cr
```diff
@@ -0,0 +1,301 @@
...
+  #   def page_item(gap : Lucky::Paginator::Gap)
+  #     text ".."
+  #   end
+  # end
+  # ```
+  def series(
+    begin beginning : Int32 = 0,
+    left_of_current : Int32 = 0,
+    right_of_current : Int32 = 0,
+    end ending : Int32 = 0
```
jwoertink: I'm shocked that works.

paulcsmith: Hah yeah me too. Glad it did though!

src/lucky/paginator/backend_helpers.cr
paulcsmith: In this case you can't since the method is hardcoded to using the current path. We could add a setter for `full_path` or an option on the `paginate` method, but I think this is a good start for MVP since I think 99% of the time you want to paginate the same url

paulcsmith:
Added the extra checks in https://github.com/luckyframework/lucky/commit/f4df5a11c1ceced46249cea0c38ac321556d3efe

PR #1021, opened 2020-03-03
Upgrade notes for 0.18 -> 0.19

paulcsmith:
There may be more, but I think this is everything


paulcsmith:
Thanks for taking a look! Will merge it in

PR #1034, opened 2020-03-07
[Breaking change] generate getters for needs in pages

### LINKED ISSUE #853: Exposing params via `needs` is not congruent
In the Pages, we access params with "@" whereas in the Action, we access params without the leading "@".    This incongruence leads to extra effort on developer to remember when "@" is required and when not.

For example:

In the Action:
```crystal
class Charts::Folder < BrowserAction
  include Auth::AllowGuests

  get "/charts/:folder" do
    render FolderPage, folder: folder
  end
end
```

In the Page:
```crystal
class Charts::FolderPage < PublicLayout
  needs folder : String

  def content
    h1 "Folder #{@folder}"
    # ...
  end
end
```

The `needs` directive should expose params so as to not need the leading "@" and the Page definition would become:
```crystal
class Charts::FolderPage < PublicLayout
  needs folder : String

  def content
    h1 "Folder #{folder}"
    # ...
  end
end
```

paulcsmith:
Closes #853

The breaking change is that you can't use ? in 'needs' anymore. Also we
may overwrite a method by generating a getter but I think there is very
little change since most methods would be defined *after* needs

### DIFF
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -36,10 +36,10 @@ end
 
 class PageWithQuestionMark
   include Lucky::HTMLPage
-  needs signed_in? : Bool
+  needs signed_in : Bool
 
   def render
-    text @signed_in.to_s
+    text signed_in?.to_s
   end
 end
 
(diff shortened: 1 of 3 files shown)


jwoertink:
wow! This is awesome!

PR #1035, opened 2020-03-07
[Breaking change] Removing link overload that takes a String. 

### LINKED ISSUE #1026: Link with just text and html attrs compiles when it shouldn't
```crystal
link "Next Page", class: "link"
```

Compiles and sets the href to "Next Page" and sets no text. I think Crystal is using the version of `link` that was supposed to be passed a block. It should fail at compile time instead

### LINKED ISSUE #1033: Proposal to drop `link to: String`.
Right now our `link` helper method has a ton of overloads. [These 3](https://github.com/luckyframework/lucky/blob/master/src/lucky/tags/link_helpers.cr#L38-L50) allow you to just pass a plain string. 

I'm not sure if this is actually useful though, and could lead to possible errors if not used properly.

Consider this:

```crystal
link "Hello", to: "/hello", class: "hello"
# vs
a "Hello", href: "/hello", class: "hello"
```

The second way is actually 1 character smaller 😅 . The benefit you get from the `link` helper method is that you can pass an action right in, and if that action isn't a GET, you'll auto get a `data-method="whatever"` added. This is gives you some safety, and extra help.

When using just the string, you could accidentally do `link to: "/ hello"`. This would compile, but not work, so you never gained any benefit from it. Yeah, this would be the same issue with `a()`, but I think it's assumed that if you go with the HTML basics, you're left to your own code. 

I propose that we deprecate those, or just raise some compile time errors telling you to just use `a()` if you're going to use a string directly.

### Downsides

The route helper method `path` returns a string, so this would no longer work

```crystal
link to: Home::Index.path
```

That is a bummer, but in this case, I'd say `link to: Home::Index` is a better option anyway.

jwoertink:
## Purpose
Fixes #1033 
Fixes #1026 

## Description
This removes the overloads for the `link` helper that take a String for the path. One of the overloads was `link(to : String, **opts)` which meant you could define `link "Home"`, and it would be fine, but that shouldn't be. This also helps to keep links a bit more safe by using the Action class as the preferred method.

## Checklist

### DIFF
--- a/src/lucky/tags/link_helpers.cr
+++ b/src/lucky/tags/link_helpers.cr
@@ -36,16 +36,33 @@ module Lucky::LinkHelpers
   end
 
   def link(text, to : String, attrs : Array(Symbol) = [] of Symbol, **html_options)
-    a text, merge_options(html_options, {"href" => to}), attrs
+    {%
+      raise <<-ERROR
+      `link` no longer supports passing a String to `to`.
+      Use `a()` or pass an Action class instead.
+
+      Example:
+        a "Home", href: "/"
+
+        link "Home", to: Home::Index
+      ERROR
+    %}
   end
 
   def link(to : String, attrs : Array(Symbol) = [] of Symbol, **html_options)
-    a attrs, merge_options(html_options, {"href" => to}) do
-      yield
-    end
-  end
+    {%
+      raise <<-ERROR
+      `link` no longer supports passing a String to `to`.
+      Use `a()` or pass an Action class instead.
 
-  def link(to : String, attrs : Array(Symbol) = [] of Symbol, **html_options)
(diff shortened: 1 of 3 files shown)


src/lucky/tags/link_helpers.cr
```diff
@@ -36,16 +36,33 @@ module Lucky::LinkHelpers
...
+      Use `a()` or pass an Action class instead.
 
-  def link(to : String, attrs : Array(Symbol) = [] of Symbol, **html_options)
-    a attrs, merge_options(html_options, {"href" => to}) do
-      yield
-    end
+      Example:
+        a "Home", href: "/"
+
+        link "Home", to: Home::Index
```
paulcsmith: Niiice

PR #1038, opened 2020-03-10
Adding new CookieNotFoundError

### LINKED ISSUE #964: Cookie error exception classes
On this line (and a few above it)

https://github.com/luckyframework/lucky/blob/3745f4aafb651de5295442877f6ab1cbde00057a/src/lucky/cookies/cookie_jar.cr#L67

We raise just a string error. These really should be exception classes so that way we can add custom responses in our Error Show action.

```crystal
raise WhoStoleTheCookieFromTheCookieJar.new("No cookie for '#{key}'")
```

jwoertink:
## Purpose
Fixes #964

## Description
When you try to call a cookie by a key, and that cookie is not found, we currently just raise a simple string error. This means that in your code you'll need to just call `rescue` which could be dangerous and mask other potential errors. 

This PR adds in a new `CookieNotFoundError` class to allow you to rescue from that specifically. I didn't change the error message that was returned, but just wrapped it in a class.

## Checklist

### DIFF
--- a/src/lucky/cookies/cookie_jar.cr
+++ b/src/lucky/cookies/cookie_jar.cr
@@ -56,7 +56,7 @@ class Lucky::CookieJar
   end
 
   def get_raw(key : Key) : HTTP::Cookie
-    get_raw?(key) || raise CookieNotFoundError.new("No cookie with the key: #{key}")
+    get_raw?(key) || raise CookieNotFoundError.new(key)
   end
 
   def get_raw?(key : Key) : HTTP::Cookie?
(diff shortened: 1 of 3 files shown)


src/lucky/cookies/cookie_jar.cr
```diff
@@ -56,15 +56,15 @@ class Lucky::CookieJar
...
+    get_raw?(key) || raise CookieNotFoundError.new("No cookie with the key: #{key}")
   end
 
   def get_raw?(key : Key) : HTTP::Cookie?
     cookies[key.to_s]?
   end
 
   def get(key : Key) : String
-    get?(key) || raise "No cookie for '#{key}'"
+    get?(key) || raise CookieNotFoundError.new("No cookie for '#{key}'")
```
paulcsmith: To make the error the same no matter where you call it you can add an initializer that accepts the key and then raise the message in the error class. 

I think we’ve done that in a couple places and it’s worked well

Issue #1028, opened 2020-03-05
Disallow label inference. Instead make it explicit


paulcsmith:
### The problem

Right now `label_for` can infer/guess the label name `label_for op.name` will generate a label with `"Name"`

This often falls down for more complex cases: `label_for op.otp_code` -> `Otp Code`

It also is not useful when you need to use a different label: `label_for op.name, "Your name:"`

It also makes the form component complex because you'd have to use a conditional to to optionally pass a label name:

```crystal
if @label_text
  label_for @field, @label_text
else
  label_for @field # infer the label name
end
```

### The fix

Require a label name! Easy to understand, easy to code, always clear how to modify it. Guessing a name you could type in about 2 seconds isn't worth the annoyance when it doesn't work or messes up code

wontruefree:
I think this is a very helpful feature.  I think this makes a lot of basic redundant usage easier.  Also because this has an overwrite to add a a label this seems like a reduction in functionality.

paulcsmith:
I think we need to require label_text in label_for otherwise things get really hairy for anything beyond the simplest components, but what we could do is add inference in the field component. So you can still do `mount Shared::Field.new(op.name)` and leave out the text. And in the field we can do something like `label_text || Wordsmith::Inflector.humanize(field.name.to_s)`

I think that'd be a nice compromise so it is clear how the label is being set and is also easier to work with in components

paulcsmith:
Also I didn't write this but we could do something like inferring the label name if the text is `nil`:

```
label_text : String?

label_for field, label_text
```

But then we lose type-safety and have potential bugs when you may accidentally send nil to label_text

label_for field, oops_this_is_nil

So that is why that approach was not taken

paulcsmith:
@wontruefree I added inference in the Shared::Field component so functionality is restored, and the component now accepts label_text for when you need to override it, so we have a bit more functionality.

I may other tweaks to `label_for` like allowing a block so you can do stuff like this: `mount Shared::Field.new(op.name, ->{ strong("Name"); span("Your full name", class: "hint")})`

In fact I may just allow passing a proc like that to any tag method. Could make for some super interesting options in components

paulcsmith:
[...]

PR #1047, opened 2020-03-14
Revert removal of label text inference and allow `nil`

paulcsmith:
https://github.com/luckyframework/lucky/issues/1028#issuecomment-599078910

> Actually I may just be overthinking this. Lucky wants to avoid bugs,
but also be practical. Sometimes nils can sneak in when you don't expect
it...but maybe label_text is not one of those places. I have a hard time
seeing a use case where someone is conditionally generating a label and
it might accidentally be nil. I think of all the apps I've ever written
I've never accidentally set label text to nil.

> So I think what I'll do is make the label text String?. So that if you
pass nil to text it will infer the label name like it does now when you
leave the arg off completely. That way we can still keep the component
flexible by allowing String? but we also keep label_for flexible. I'll
go ahead and make that change now. This also means existing apps will
still work when you upgrade and this will no longer be a breaking change
👍

### DIFF
--- a/spec/lucky/label_helpers_spec.cr
+++ b/spec/lucky/label_helpers_spec.cr
@@ -1,5 +1,7 @@
 require "../spec_helper"
 
+include ContextHelper
+
 class TestUser
   def first_name
     "My Name"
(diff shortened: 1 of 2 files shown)


paulcsmith:
Thanks for your feedback on the issue @wontruefree. I think this should be more flexible than my previous solution and will make writing Field components a lot simpler

PR #1053, opened 2020-03-19
Lucky::Params and escape hatch methods (from_json/query/form/multipart)

paulcsmith:
This PR improves documentation in Lucky::Params as well as adds some methods for interacting with the raw data more easily if needed:

* from_json
* from_query
* from_form
* from_multipart

### DIFF
--- a/src/lucky/json_body_parser.cr
+++ b/src/lucky/json_body_parser.cr
@@ -1,28 +1,10 @@
-# Parse the body of an `HTTP::Request` to a `JSON::Any`
-#
-# > You can get the json body in `Lucky::Action` by calling `json_body`.
-# > You can also get JSON params with `Lucky::Params#get/nested`.
-# > You rarely need to use this class directly, but it is here if needed.
-#
-# Parsed the request body with `JSON.parse`. If the request body
-# is empty (`""`) it will parse an empty object (`{}`) rather than raising.
-#
-# If the body is not valid JSON it will raise a `Lucky::ParamParsingError`
+# :nodoc:
 class Lucky::JsonBodyParser
   getter request : HTTP::Request
 
   def initialize(@request)
   end
 
-  # Returns the parsed JSON or raises if JSON is not valid
-  #
-  # ```
-  # json_body = Lucky::JsonBodyParser.new(request).parsed_json
-  # # {"page": 1}
-  # json_body["page"].as_i # 1
-  # # {"users": [{"name": "Skyler"}]}
-  # json_body["users"].as_a.first.["name"].as_s # "Skyler"
-  # ```
   def parsed_json : JSON::Any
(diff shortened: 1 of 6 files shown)


src/lucky/json_body_parser.cr
```diff
@@ -0,0 +1,21 @@
+# :nodoc:
+class Lucky::JsonBodyParser
+  getter request : HTTP::Request
+
+  def initialize(@request)
+  end
+
+  def parsed_json : JSON::Any
+    if body.blank?
+      JSON.parse("{}")
```
wontruefree: instead of firing up the parser could this just be initialized.
```suggestion
      JSON::Any.new(nil)
```

paulcsmith: Fantastic idea!

jwoertink: I agree. Maybe instead of `nil` though, we do `JSON::Any.new({} of String => JSON::Any)`?

src/lucky/json_body_parser.cr
```diff
@@ -0,0 +1,21 @@
+# :nodoc:
+class Lucky::JsonBodyParser
+  getter request : HTTP::Request
+
+  def initialize(@request)
+  end
+
+  def parsed_json : JSON::Any
+    if body.blank?
+      JSON::Any.new({} of String => String)
```
jwoertink: I think this needs needs to be `{} of String => JSON::Any`

src/lucky/params.cr
```diff
@@ -396,21 +475,11 @@ class Lucky::Params
   @_parsed_json : JSON::Any?
 
   private def parsed_json : JSON::Any
```
jwoertink: This could just use `memoize`

src/lucky/json_body_parser.cr
paulcsmith: Yeah I got to trigger happy and forgot to run specs. Oops :)

src/lucky/params.cr
paulcsmith: Good idea

src/lucky/params.cr
```diff
@@ -393,24 +473,12 @@ class Lucky::Params
...
-    @_parsed_json ||= begin
-      if body.blank?
-        JSON.parse("{}")
-      else
-        JSON.parse(body)
-      end
-    end
-  rescue JSON::ParseException
-    raise Lucky::ParamParsingError.new(request)
+  private memoize def parsed_json : JSON::Any
```
jwoertink: Does this work? 😂  I didn't know that!

paulcsmith: Yup! It looks weird but macros before defs have to go between the private haha

PR #1059, opened 2020-03-21
Adding a new remote_ip method

### LINKED ISSUE #1000: Should we have a remote_ip helper method in our Actions?
In Rails, you have access to the `request.remote_ip` helper method which does this:

```ruby
# File actionpack/lib/action_controller/request.rb, line 221
    def remote_ip
      remote_addr_list = @env['REMOTE_ADDR'] && @env['REMOTE_ADDR'].scan(/[^,\s]+/)
      unless remote_addr_list.blank?
        not_trusted_addrs = remote_addr_list.reject {|addr| addr =~ TRUSTED_PROXIES}
        return not_trusted_addrs.first unless not_trusted_addrs.empty?
      end
      remote_ips = @env['HTTP_X_FORWARDED_FOR'] && @env['HTTP_X_FORWARDED_FOR'].split(',')
      if @env.include? 'HTTP_CLIENT_IP'
        if ActionController::Base.ip_spoofing_check && remote_ips && !remote_ips.include?(@env['HTTP_CLIENT_IP'])
          # We don't know which came from the proxy, and which from the user
          raise ActionControllerError.new("IP spoofing attack?!\nHTTP_CLIENT_IP=\#{@env['HTTP_CLIENT_IP'].inspect}\nHTTP_X_FORWARDED_FOR=\#{@env['HTTP_X_FORWARDED_FOR'].inspect}\n")
        end
        return @env['HTTP_CLIENT_IP']
      end
      if remote_ips
        while remote_ips.size > 1 && TRUSTED_PROXIES =~ remote_ips.last.strip
          remote_ips.pop
        end
        return remote_ips.last.strip
      end
      @env['REMOTE_ADDR']
    end
```
[with spec](https://github.com/rails/rails/blob/470e6bdac97249ca3406c635f611aa8f7df8b222/railties/test/application/middleware/remote_ip_test.rb)

I didn't see anything in crystal that does this, but I did fine [remote_address](https://github.com/crystal-lang/crystal/blob/41bd18fbea4aec50aad33aa3beb7a0bf30544186/src/http/request.cr#L30) which doesn't seem to help much. 

In one of my apps, I started by doing this:

```
  private def remote_ip : String
    request.headers["X-Real-IP"]? ||
      request.headers["HTTP_X_FORWARDED_FOR"]? ||
      request.headers["X_FORWARDED_FOR"]? ||
      request.headers["REMOTE_ADDR"]? ||
      request.headers["HTTP_CLIENT_IP"]? ||
      request.headers["HTTP_X_FORWARDED"]? ||
      request.headers["HTTP_X_CLUSTER_CLIENT_IP"]? ||
      "127.0.0.1"
  end
```

but realized that `X-Real-IP` gives me the wrong IP, and in my case, what I need is the `HTTP_X_FORWARDED_FOR` split by comma, and the first returned. It looks like the rails version of this would actually have returned the correct one for me. Also note that this method catches IP Spoofing attacks (which my rails apps throw all the time). 

I'd be curious to hear everyone else's thoughts on if this is the type of logic that would be a per app basis, or something that would be beneficial to have baked in.

jwoertink:
## Purpose
Fixes #1000

## Description
Crystal has a build in `request.remote_address` method which is nice, but if your app is behind a CDN, or reverse proxy, then the value you get back is of that CDN or proxy. The most common method I found for getting the client's IP is by checking the `HTTP_X_FORWARDED_FOR` http header. This value usually contains all of the IPs that are being forwarded along as a CSV. The first in the list is the client's IP.

There are several other headers that could technically contain a value (shown in the issue). We could always add those if we wanted, but I think this at least gives us a little boost. The only other thing I think I'd like to add later is the handling of the IP spoofing like rails has (also shown in the original issues).

## Checklist


src/lucky/remote_ip_address.cr
```diff
@@ -0,0 +1,15 @@
...
+  # Returns a `String` IP Address of the remote client by
+  # looking at a few different possible options.
+  #
+  # * HTTP_X_FORWARDED_FOR header
+  # * REMOTE_ADDRESS header
+  # * Fallback to localhost 127.0.0.1
+  def remote_ip : String
+    request = @context.request
+
+    request.headers["HTTP_X_FORWARDED_FOR"]?.try(&.split(',').first) ||
```
paulcsmith: There is a potential bug here. If HTTP_X_FORWARDED_FOR is an empty string it will try to split it and will return an empty string. See `"".try(&.split(',').first) # => ""`.

I think if you added a spec for an empty string HTTP_X_FORWARDED_FOR that'd be good

src/lucky/remote_ip_address.cr
```diff
@@ -0,0 +1,15 @@
...
+      request.remote_address ||
+      "127.0.0.1"
```
paulcsmith: I think rather than returning a potentially incorrect remote ip it'd be better to use nil and let people handle that however you want. For example if the ip is empty you may want to raise, or maybe log. Right now you couldn't do that because you don't know if the ip was actually 127.0.0.1 or if it just fell back to that

paulcsmith:
Also this little bit in the article linked above:

> Unless you’re doing IP-based authentication, it’s probably safe to ignore or disable this alarming exception

That's the only time I can think this really matters, so we can prob ship with this as long as we note that

bdtomlin:
@jwoertink I know you use caddy server. FYI, when using caddy as a proxy it uses X-Forwarded-For, not HTTP_X_FORWARDED_FOR. I don't really know where HTTP_X_FORWARDED_FOR is used commonly but it appears X-Forwarded-For is the standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

jwoertink:
@bdtomlin Good call. True, I do use Caddy, but only in front of my rails apps. My Lucky apps have Cloud front in front of them, and that thing passes `HTTP_X_FORWARDED_FOR`. I'm ok with adding an additional one. I noticed some PHP frameworks run through several different headers. I'll add that one in too.

bdtomlin:
@jwoertink, thanks!

jwoertink:
So I decided to drop the `HTTP_X_` in favor of what the standard is. After doing some digging, I found that the prefix of `HTTP_` was from back in the old CGI days. That's why PHP and [rails](https://github.com/rails/rails/blob/7b29bc2179702a04613068e6e6efd6cf7f343440/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L138) apps both use it. I'll need to dig in to our CF setup to see why we're getting that value. I'm wondering if we just mistakenly set that up wrong. 

In any case, best to use what the defacto is. Good call @bdtomlin!

bdtomlin:
@jwoertink it’s not just the http prefix X-Forwarded-For uses dashes and not underscores. I set it up with underscores in my app initially which led to empty data in production 😔

jwoertink:
I was under the impression that crystal handles that conversion internally [here](https://github.com/crystal-lang/crystal/blob/master/src/http/headers.cr#L40-L49). Is that not the case?

jwoertink:
Yeah, it looks like they're treated the same https://play.crystal-lang.org/#/r/8r5d 

Now interesting thing is, it looks like HTTP/2 will require all headers to be lowercase https://http2.github.io/http2-spec/#rfc.section.8.1.2. So really, it should be `x-forwarded-for` which crystal also seems to handle https://play.crystal-lang.org/#/r/8r5e

bdtomlin:
It sure does. I looked through the past commits on that to be sure. Clearly I was remembering that wrong :)

I was actually using 'HTTP_X_FORWARDED_FOR' and ended up with empty data. I think I must have used that originally based on this issue. I should have said something then. Sorry :[

jwoertink:
No worries! These sorts of things are hard to get right, so I want to make sure we're doing things the way we should. Always good to have a few extra set of eyes on it. 

I'm thinking maybe I leave this as is for now, and then we can go back and make some adjustments to all mentions of headers in a later PR. I'll open an issue to track it.

PR #1062, opened 2020-03-24
Only log halted pipes by default

### LINKED ISSUE #1054: Add option to only log 'before/after' pipes if the pipe halts
Sometimes you don't really need to show which pipe is running and you only care if it halts. For example the protect_from_forgery pipe and test_backdoor pipe.

Would love to do this:

```crystal
before require_admin # Log when it halts and when it continues
before protect_from_forgery, only_log_when_halted: true
before test_backdoor, only_log_on_halt: true
```

Really unsure of a good name for this option though. I want it to be clear since it won't be used that much, but also don't want it to be so long you fall asleep halfway through reading the line

paulcsmith:
But allow customizing it so that you can see all pipes being logged

Closes #1054


spec/lucky/action_callbacks_spec.cr
```diff
@@ -207,7 +207,7 @@ private def with_log
     log_formatter: RawLogFormatter
   )
 
-  Lucky.temp_config(logger: logger) do
+  Lucky.temp_config(logger: logger, action_pipe_log_level: Logger::Severity::INFO) do
```
paulcsmith: Actually, should we just call these `before/after` macros callbacks? We kind of have callbacks and pipes in the code and guides so I think we should choose one.

I partly named them pipes because they work a bit more similarly to pipes in Elixir where you have to return an explicit response/continue. And also because "callbacks" is a bad word to some people. But maybe we should just call them callbacks since that may make more sense. Thoughts? @edwardloveall @jwoertink

src/lucky.cr
```diff
@@ -22,6 +22,7 @@ require "./lucky/paginator/*"
 module Lucky
   Habitat.create do
     setting logger : Dexter::Logger
+    setting action_pipe_log_level : ::Logger::Severity? = nil, example: "Logger::Severity::INFO"
```
edwardloveall: What does `example` mean here? I think I could guess but I've never seen it. 😄

spec/lucky/action_callbacks_spec.cr
edwardloveall: I think pipes is fine, but I don't think I feel too strongly either way. There's also an ick factor about callbacks for me, but not so strong that I'd steer away from the word itself if it's the best word to discribe.

jwoertink: I like pipes because it sounds cool. Maybe that's because of callbacks having a bad taste, but we use the term "pipe" in our apps. I would like to see it a bit more standardized, and consistent though. 

Like we have:
```
module RequireSitePipe
  macro included
    before require_site
  end

  private def require_site
    if current_site?
      continue
    else
      raise Lucky::RouteNotFoundError.new(context)
    end
  end

  private def current_site : SiteHelpers::Site
    current_site?.not_nil!
  end
end
```

but just by looking at the name of the module, it doesn't immediately tell you why we added "Pipe" to it. We could have just called it `RequireSite`.

src/lucky.cr
jwoertink: `example` is what gets displayed in the error message when it's not set. But since this is nilable, does it throw an error?

src/lucky/action_callbacks.cr
```diff
@@ -127,10 +127,10 @@ module Lucky::ActionCallbacks
       #   end
 
       if callback_result.is_a?(Lucky::Response)
-        Lucky::ActionCallbacks.log_stopped_callback(context,"{{ callback_method.id }}")
+        Lucky::ActionCallbacks.log_stopped_callback("{{ callback_method.id }}")
```
jwoertink: Yeah, I see what you mean. This should be `Lucky::ActionPipes.log_stopped_pipe`, or we go full on callback everywhere. I do like "pipe", but maybe we don't fight it since typing `callback` just seems to be what you want to do first?

spec/lucky/action_callbacks_spec.cr
paulcsmith: As mentioned in our gitter room I'll stick with pipes. I was just worried it was confusing. I'll make some changes to the code to be pipe instead of callback!

src/lucky.cr
paulcsmith: Yeah it shows up if you forget to set it, but Jeremy is right that you'll never see it :P

I think I will leave it though since in the future I want to add a task to Habitat that will show the options available, examples, and also where the configuration is being set. It'll be pretty sick I think

paulcsmith:
It is passing but not reporting back to GitHub so I'll merge https://travis-ci.org/github/luckyframework/lucky

edwardloveall:
We had that problem on a different project also where circle wasn't reporting to GH. Maybe it's the same problem? https://github.com/gabebw/hotline-webring/pull/273

PR #1065, opened 2020-03-26
Remove changed file log

paulcsmith:
I've never seen this and thought "Oh yeah I did change that file!"

I think we can just remove it so there is less noise in the log


paulcsmith:
It still lets you know it’s compiling. This just removes the part that says what file changes in yellow. Is that what you meant?

jwoertink:
Yup, exactly.

paulcsmith:
Perfect!

PR #1067, opened 2020-03-26
Add compiling spinner to watch task

paulcsmith:
This does not work yet. We need to point the other shards in Lucky to
use LuckyCli master. Since they don't I just brought in the Spinner for
testing.

The spinner works fine when running `lucky watch` directly. But when run
inside of overmind it shows nothing...not sure why

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -5,6 +5,134 @@ require "yaml"
 require "../src/lucky/server_settings"
 require "option_parser"
 
+# Adding this in From LuckyCli for now
+class LuckyCli::Spinner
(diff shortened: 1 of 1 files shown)


skojin:
I not sure about spinner, but have "Done" in different color is nice to have, like webpack do

Screen Shot 2020-05-07 at 21 00 27

paulcsmith:
@skojin That's an awesome idea!

matthewmcgarvey:
Closing this just because I don't think it's going to be picked up any time soon. If anyone want's to pick it up feel free

Issue #1073, opened 2020-03-30
Add `current_page?` helper method to pages and components


paulcsmith:
This will be used by https://github.com/luckyframework/lucky/issues/969

Should work similarly to Rails https://apidock.com/rails/ActionView/Helpers/UrlHelper/current_page%3F

```crystal
current_page?(User::Index)
current_page?(Users::Show.with(user.id))
current_page?("/raw/string")
```

By default it should *not* care about query params but passing `check_query_params: true` will include the path and query params

cc @wout

wout:
@paulcsmith Do you want include functionality for full uri's, including `scheme`, `host` and `port`? Or only for `path` or `path` + `query`?

paulcsmith:
@wout Great question. I think host, scheme and port should be options. But I'd also be ok shipping those in a separate PR to get the basic use case of path done. Up to you though!

wout:
Ok, I'll have a look at host and port as well. However, I need a pointer here since Crystal's `HTTP::Request` does not carry scheme information. Is it available somewhere in Lucky?

I was wondering if it is relevant these days. I can't think of all the use cases, but isn't it either http or https? And if so, since https is becoming the default these days, should we even test on scheme?

paulcsmith:
@wout Great point. We can skip that I think. Everyone should be using ssl and Lucky makes it easy to force with the ForceSslHandler. Also scheme shouldn't change what page is shown as long as the rest of the host and path match 👍

wout:
@paulcsmith That's done then. :)

PR #1074, opened 2020-03-31
Add current_page? helper for pages.

wout:
## Purpose
Implementing #1073.

## Description
This adds the `current_page?` helper with similar functionality like the one in Rails. Although, it does not cover all Rails' cases. What it does not do:

1. handle cases with [double escaped params](https://github.com/rails/rails/blob/master/actionview/test/template/url_helper_test.rb#L594-L598)
2. handle cases with [different encoding](https://github.com/rails/rails/blob/master/actionview/test/template/url_helper_test.rb#L586-L592)
3. test against a [full uri](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/url_helper.rb#L567), including `protocol` and `port`

I'm not quite sure if they are relevant or even possible. Looking at the available data in a request, I wonder if it is even feasible to implement point **3**. There is no notion of the protocol, but even then, would we even need to test the protocol?

## Checklist

### DIFF
--- a/spec/lucky/url_helpers_spec.cr
+++ b/spec/lucky/url_helpers_spec.cr
@@ -67,6 +67,27 @@ describe Lucky::UrlHelpers do
           .current_page?("/action", check_query_params: true)
           .should be_false
       end
+
+      it "does not care about the order of query params" do
+        view_for("/action?order=desc&page=1")
+          .current_page?("/action?order=desc&page=1", check_query_params: true)
+          .should be_true
+        view_for("/action?order=desc&page=1")
+          .current_page?("/action?page=1&order=desc", check_query_params: true)
+          .should be_true
+      end
+
+      it "ignores anchors" do
+        view_for("/pages/123").current_page?("/pages/123#section")
+          .should be_true
+        view_for("/pages/123#section").current_page?("/pages/123")
+          .should be_true
+        view_for("/pages/123#section").current_page?("/pages/123#section")
+          .should be_true
+        view_for("/pages/123")
+          .current_page?("/pages/123#section", check_query_params: true)
+          .should be_true
+      end
     end
 
     context "given a browser action" do
@@ -76,6 +97,42 @@ describe Lucky::UrlHelpers do
         view_for("/pages/123").current_page?(Pages::Show.with(12))
           .should be_false
         view_for("/pages").current_page?(Pages::Index)
+          .should be_true
+        view_for("/pages")
+          .current_page?(Pages::Index.with(page: 2))
+          .should be_true
+        view_for("/pages?page=2")
+          .current_page?(Pages::Index)
+          .should be_true
+      end
+
+      it "checks query params if explicitly required" do
+        view_for("/pages")
+          .current_page?(Pages::Index, check_query_params: true)
+          .should be_true
+        view_for("/pages?page=2")
+          .current_page?(Pages::Index.with(page: 2), check_query_params: true)
+          .should be_true
+        view_for("/pages")
+          .current_page?(Pages::Index.with(page: 2), check_query_params: true)
+          .should be_false
+        view_for("/pages?page=2")
+          .current_page?(Pages::Index, check_query_params: true)
+          .should be_false
+      end
+
+      it "ignores anchors" do
+        view_for("/pages/123")
+          .current_page?(Pages::Show.with(123, anchor: "section"))
+          .should be_true
+        view_for("/pages/123#section")
+          .current_page?(Pages::Show.with(123))
+          .should be_true
+        view_for("/pages/123#section")
+          .current_page?(Pages::Show.with(123, anchor: "section"))
+          .should be_true
+        view_for("/pages/123")
+          .current_page?(Pages::Show.with(123, anchor: "section"), check_query_params: true)
           .should be_true
       end
     end
@@ -98,6 +155,8 @@ private class TestPage
 end
 
 class Pages::Index < TestAction
+  param page : Int32 = 1
+
   get "/pages" do
     plain_text "I'm just a list of pages"
   end
--- a/spec/lucky/url_helpers_spec.cr
+++ b/spec/lucky/url_helpers_spec.cr
@@ -1,0 +1,98 @@
+require "../spec_helper"
+
+include ContextHelper
+
+describe Lucky::UrlHelpers do
+  describe "#current_page?" do
+    context "given a string" do
+      it "tests if a path matches the request path or not" do
+        view_for("/").current_page?("/").should be_true
+        view_for("/page").current_page?("/gum").should be_false
+        view_for("/page").current_page?("/page").should be_true
+        view_for("/page").current_page?("/page/").should be_true
+        view_for("/page/").current_page?("/page").should be_true
+        view_for("/page/").current_page?("/page/").should be_true
+      end
+
+      it "tests if the path of a url matches request path or not" do
+        view_for("/").current_page?("https://example.com/")
+          .should be_true
+        view_for("/page").current_page?("https://example.com/page")
+          .should be_true
+        view_for("/page").current_page?("https://example.com/gum")
+          .should be_false
+        view_for("https://example.com/gum").current_page?("/gum")
+          .should be_true
+      end
+
+      it "only tests positive for get and head" do
+        view_for("/get", "GET").current_page?("/get").should be_true
+        view_for("/head", "HEAD").current_page?("/head").should be_true
+        view_for("/post", "POST").current_page?("/post").should be_false
+        view_for("/put", "PUT").current_page?("/put").should be_false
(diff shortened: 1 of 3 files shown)


src/lucky/page_helpers/url_helpers.cr
```diff
@@ -0,0 +1,65 @@
...
+  # current_page?(Pages::Show.with(456))
+  # # => false
+  # # Visiting https://example.com/pages
+  # current_page?(Pages::Index)
+  # # => true
+  # current_page?(Blog::Index)
+  # # => false
+  # ```
+  def current_page?(
+    action : Class | Lucky::RouteHelper,
```
wout: Is `Class` the correct type here for `Pages::Index`?

src/lucky/page_helpers/url_helpers.cr
```diff
@@ -0,0 +1,65 @@
...
+    check_query_params : Bool = false
+  )
+    current_page?(action.path)
```
wout: Does `.path` here need some arguments?

src/lucky/page_helpers/url_helpers.cr
```diff
@@ -0,0 +1,65 @@
...
+  # current_page?("https://example.com/shop/products")
+  # # => true
+  # current_page?("https://example.com/shop/products?order=desc&page=1")
+  # # => false
+  # ```
+  def current_page?(
+    path : String,
+    check_query_params : Bool = false
+  )
+    return false unless {"GET", "HEAD"}.includes?(@context.request.method)
```
wout: I ran some benchmarks and as it turns out, `Tuple#includes?` is by far the fastest option. Even faster than `==`:

```
includes? with % array  37.67M ( 26.55ns) (± 4.11%)  64.0B/op  12.39× slower
  includes? with array  37.81M ( 26.45ns) (± 3.71%)  64.0B/op  12.34× slower
  includes? with tuple 466.68M (  2.14ns) (± 3.95%)   0.0B/op        fastest
 includes? with string  49.91M ( 20.04ns) (± 1.65%)   0.0B/op   9.35× slower
               == & == 458.88M (  2.18ns) (± 5.26%)   0.0B/op   1.02× slower
                 match  20.62M ( 48.49ns) (± 3.14%)  16.0B/op  22.63× slower
```

src/lucky/page_helpers/url_helpers.cr
jwoertink: It would be `Lucky::Action.class`.

src/lucky/page_helpers/url_helpers.cr
jwoertink: `path` can take optional params, or an `anchor`. Not sure if that would be necessary here or not.

src/lucky/page_helpers/url_helpers.cr
```diff
@@ -0,0 +1,65 @@
...
+  # # => false
+  # # Visiting https://example.com/pages
+  # current_page?(Pages::Index)
+  # # => true
+  # current_page?(Blog::Index)
+  # # => false
+  # ```
+  def current_page?(
+    action : Class | Lucky::RouteHelper,
+    check_query_params : Bool = false
```
jwoertink: Is this used?

src/lucky/page_helpers/url_helpers.cr
wout: Thanks. Now I feel a bit stupid I haven't thought of that myself. :blush:

src/lucky/page_helpers/url_helpers.cr
wout: I was thinking about pagination for example. I'll have to dig a little further to see how it works.

AFAIK, anchors never reach the server, so I don't think we need to include those cases.

src/lucky/page_helpers/url_helpers.cr
wout: It depends on the question above about `.path`. If path does not take any arguments, this can co.

wout: @paulcsmith Can you shed some light here? It's not clear to me from reading the source if for example `Pages::Index.path` or `Pages::Show.with(123).path` would ever need query params as an argument. If not, then we do not need a `check_query_params` argument here. Or am I missing something?

wout:
Just pushed an updated version where host with port are tested as well if the given value is a full uri. For example:

```crystal
# Visiting https://example.com/action
view.current_page?("https://example.com/action")
# # => true
# Visiting https://example.io/action
view.current_page?("https://example.com/action")
# # => false
```

But:

```crystal
# Visiting https://example.com/action
view.current_page?("http://example.com/action")
# # => true
```

Might be confusing.

src/lucky/page_helpers/url_helpers.cr
jwoertink: What do you get with this?

```
# Visiting https://example.com/action?page=2&per_page=1
view.current_page?("https://example.com/action?page=1&per_page=100")
```

wout: That will return `true`, unless you set `check_query_params` to `true`:

```crystal
# Visiting https://example.com/action?page=2&per_page=1
view.current_page?(
  "https://example.com/action?page=1&per_page=100",
  check_query_params: true)
# => false
```

wout: But that reminds me of something. This would be `true`:

```crystal
# Visiting https://example.com/action?page=2&per_page=1
view.current_page?("https://example.com/action?page=2&per_page=1")
# => true
```

But this would be `false`:

```crystal
# Visiting https://example.com/action?page=2&per_page=1
view.current_page?("https://example.com/action?per_page=1&page=2")
# => false
```

I'm not sure if that's an issue or not. Will query params always have the same order?

jwoertink: No, because we don't know if they were built by hand or not. Like

`a href: "/action?per_page=1&page=#{current_page}"`

src/lucky/page_helpers/url_helpers.cr
paulcsmith: I believe that it does not need anything extra. If I remember correctly `#path` returns anchors and query params given to `Lucky::Action.with` :D

So if you do `current_page?(Users::Index.with(page: 2)` the page param should be included

spec/lucky/url_helpers_spec.cr
```diff
@@ -0,0 +1,110 @@
...
+
+private def view_for(
+  path : String,
+  method : String = "GET",
+  host_with_port : String = "example.com"
+)
+  request = HTTP::Request.new(method, path)
+  request.headers["Host"] = host_with_port
+  TestPage.new(build_context(path: path, request: request))
+end
```
paulcsmith, line 93: Nice helper method! 👍

spec/lucky/url_helpers_spec.cr
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    context "given a browser action" do
+      it "tests if the path matches or not" do
+        view_for("/pages/123").current_page?(Pages::Show.with(123))
+          .should be_true
+        view_for("/pages/123").current_page?(Pages::Show.with(12))
+          .should be_false
+        view_for("/pages").current_page?(Pages::Index)
+          .should be_true
```
paulcsmith, line 79: Thoughts on adding a spec where you include an anchor and another for an action with a query param?

spec/lucky/url_helpers_spec.cr
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      it "only tests positive for get and head requests" do
+        view_for("/get", "GET").current_page?("/get").should be_true
+        view_for("/head", "HEAD").current_page?("/head").should be_true
+        view_for("/post", "POST").current_page?("/post").should be_false
+        view_for("/put", "PUT").current_page?("/put").should be_false
+        view_for("/patch", "PATCH").current_page?("/patch").should be_false
+        view_for("/delete", "DELETE").current_page?("/delete").should be_false
+      end
```
paulcsmith, line 45: Nice catch! I hadn't thought of this

src/lucky/page_helpers/url_helpers.cr
wout: Ok, in that case the params need to be parsed, sorted and compared.

wout: That fixed.

spec/lucky/url_helpers_spec.cr
wout: They are now added.

wout:
I assume this was good to go, but just thought of something. The query params probably need to be escaped before testing.

wout:
Ok, this is the last bit I think. Now uri encoded query params are properly decoded before testing.

PR #1078, opened 2020-04-03
0.20.0 UPGRADE NOTES 

jwoertink:
## Purpose
Adding in the new UPGRADE notes for the next Lucky release.


UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+  - Crystal should be `0.34.0`
+  - Lucky should be `~> 0.20.0`
+  - Authentic should be `~> 0.5.2`
+  - LuckyFlow should be `~> 0.6.2`
+- Run `shards update`
+
+### General updates
+
+- Update: any `link()` where the `to:` option is a `String` to an action class. e.g. `link("Home", to: "/")` -> `link("Home", to: Home::Index)`. If no action class is available, use `a(href: "")` instead.
+- Remove: the `?` from any `needs` using a predicate method. e.g. `needs signed_in? : Bool` -> `needs signed_in : Bool`
```
paulcsmith, line 29: Maybe add 

> Lucky now automatically creates a method ending with `?` for needs that use a `Bool` type.

So people know they can still use `?` methods

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Update: your local `ENV["PORT"]` to be `ENV["DEV_PORT"]`.
```
paulcsmith, line 30: Where is this done? As far as I am aware this isn't used in default Lucky apps so is this something app specific? If so maybe something like: 

> If you want to customize the port that Luck runs on in development, use `ENV["DEV_PORT"]` instead of `ENV["PORT"]`

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Update: all `SaveOperation` classes where a raw hash is being passed in. e.g. `MyOperation.new({"name" => "Gary"})` -> `MyOperation.new(Avram::Params.new({"name" => "Gary"}))`
```
paulcsmith, line 31: I think the solution here should be to use named arguments instead:

```suggestion
- Update: all `SaveOperation` classes where a raw hash is being passed in. e.g. `MyOperation.new({"name" => "Gary"})` -> `MyOperation.new(name: "Gary")`, or if you must use a hash, wrap it in params first: `MyOperation.new(Avram::Params.new({"name" => "Gary"})`
```

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Remove: the `on:` option from `needs` inside every Operation class. e.g. `needs created_by : String, on: :create` -> `needs created_by : String`
```
paulcsmith, line 32: Maybe add that this means that you will need to pass in the needs when calling `new`, `create`, and `update`

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+
+
+### Optional updates
+
+- Update: all instance variables called from a `needs` on a page can now just use the method of that name. e.g. `@page` -> `page`
```
paulcsmith, line 37:
```suggestion
- Update: all instance variables called from a `needs` on a page or component can now just use the method of that name. e.g. `@current_user` -> `current_user`
```

Just changed a bit to use a variable that is most likely to be used in a Lucky app

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Add: `include Lucky::CatchUnpermittedAttribute` to the `class Shared::Field(T)` in `src/components/shared/field.cr`
```
paulcsmith, line 38:
```suggestion
- Add: `include Lucky::CatchUnpermittedAttribute` to the `class Shared::Field(T)` in `src/components/shared/field.cr`. This will raise a nicer error if you forget to permit a column in your SaveOperation
```

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Add: the new `Lucky::RemoteIpHandler.new` to your app handlers in `src/app_server.cr` just before `Lucky::RouteHandler.new`.
```
paulcsmith, line 39: 🎉

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Add: `robots.txt` to your `public/` directory.
+  ```
+  User-agent: *
+  Disallow:
+  ```
```
paulcsmith, line 44: 🎉 Thanks @watzon

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+- Update: `UserSerializer` to inherit from the `BaseSerializer` if it doesn't already.
+- Add: `cookie.http_only(true)` to your `config/cookies.cr` file. This goes inside your `settings.on_set` block.
+- Update: your node dependencies where needed
+- Add: this line `system_check: script/system_check && $SHELL` to your `Procfile.dev`
+- Add: the new `system_check` script in `script/system_check` and script helpers `function_helpers` and `text_helpers` in to `script/helpers/`
+<details>
```
paulcsmith, line 50: Maybe this should be a link to a gist. The summary helps, but a gist is easier to copy IMO because you can view the raw file. And it can be split into three separate files so it is a bit easier to read. Thoughts?

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,212 @@
...
+  }
+
+  # Prints out a check mark and Done.
+  # example:
+  #   print_done
+  print_done() {
+    printf "✔ Done\n" | indent
+  }
+  ```
+</details>
```
paulcsmith, line 207: Also I think an update script/setup may be helpful to include too

UPGRADE_NOTES.md
jwoertink: Can we make gists through the Lucky org? I can do them on mine directly, but I don't want to one day delete it and forget or whatever.

jwoertink: I guess we can also just link to the file in LuckyCli?

paulcsmith: Great idea! Yeah I think a permalink to the lucky CLI files is a great idea

PR #1084, opened 2020-04-04
Prints io directly instead of converting to string

### LINKED ISSUE #507: Render directly to IO
https://stackoverflow.com/questions/50774451/how-to-free-memory-allocated-for-some-structure-in-crystal-manually/50782379#50782379

For both JSON and HTML pages.

Right now I don't think this is a big deal as performance is already excellent, but this should be done eventually

paulcsmith:
Closes #507

This prints the io directly to the response, rather than making it a
string first. The bigger the response, the more helpful this will be.

Benchmarks of printing directly v. converting to String first:

```crystal
require "benchmark"

text = <<-TEXT
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua
TEXT

string_text = text * 40_000
io_text = IO::Memory.new(string_text)

Benchmark.ips do |x|
  x.report("string") do
    io = IO::Memory.new
    io.print(io_text.to_s)
  end
  x.report("io") do
    io = IO::Memory.new
    io.print(io_text)
  end
end
```

Results:

```
string 942.67  (  1.06ms) (± 1.33%)  12.7MB/op   2.78× slower
    io   2.62k (381.73µs) (± 3.52%)   8.0MB/op        fastest
```

So far medium/large HTML responses this will shave off about .6ms. Not massive, but a quick and easy win

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -131,7 +131,7 @@ describe Lucky::Action do
     it "render assigns" do
       response = Rendering::Index.new(build_context, params).call
 
-      response.body.should contain "Anything"
+      response.body.to_s.should contain "Anything"
       response.debug_message.to_s.should contain "Rendering::IndexPage"
     end
   end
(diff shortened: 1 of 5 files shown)


paulcsmith:
Also, we could further optimize this, but I wanted to get the easy win first. Maybe we can skip making a new IO to speed it up even more. But we'd have to modify TextResponse and do a few other things so I'm gonna put that off for now

PR #1086, opened 2020-04-05
Improve error message when returning non Lucky::Response

### LINKED ISSUE #460: (Re)Add helpful error message when failing to return a Lucky::Response in an action
See https://github.com/luckyframework/lucky/issues/437

> It seems this may be an issue in Crystal, but I need more time to track
it down. For now, I'm going to remove this special error-handling and
add it back later. That's better than giving people incorrect
instructions. Note that a compile-time error will still be raised, it
will just be a generic Crystal error.

### LINKED ISSUE #1037: Update mention of render -> html
https://github.com/luckyframework/lucky/blob/9aed433f9c63c28f9f5a4e6e7f514698cd4b6ef6/src/lucky/renderable.cr#L118

This line mentions `render`, but it should  say `html/redirect/json`

paulcsmith:
~Closes~ #460 https://github.com/luckyframework/lucky/pull/1086#discussion_r403729656
Closes #1037 and adds examples

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -103,19 +103,56 @@ module Lucky::Renderable
     response.print
   end
 
+  private def handle_response(_response : Nil)
+    {%
+      raise <<-ERROR
+
+
+      An action returned Nil
+
+      But it should return a Lucky::Response.
+
+      Try this...
+
+        ▸ Return a response with html, redirect, or json at the end of your action.
+        ▸ Ensure all conditionals (like if/else) return a response with html, redirect, json, etc.
+
+      For example...
+
+        get "/admin/users" do
+          # Make sure there is a response in all conditional branches
+          if current_user.admin?
+            html IndexPage, users: UserQuery.new
+          else
+            redirect Home::Index
+          end
+        end
+
+      ERROR
+    %}
+  end
+
   private def handle_response(_response : T) forall T
     {%
       raise <<-ERROR
 
 
-      Your action returned #{T}
+      An action returned #{T}
 
       But it should return a Lucky::Response
 
       Try this...
 
-        ▸ Use a method like render/redirect/json at the end of your action.
-        ▸ Ensure all conditionals (like if/else) return a response with render/redirect/json/etc.
+        ▸ Return a response with html, redirect, or json at the end of your action.
+        ▸ Ensure all conditionals (like if/else) return a response with html, redirect, json, etc.
+
+      For example...
+
+        get "/users" do
+          # Return a response with json, redirect, html, etc.
+          html IndexPage, users: UserQuery.new
+        end
+
       ERROR
     %}
   end


src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
...
+      For example...
+
+        get "/admin/users" do
+          # Make sure there is a response in all conditional branches
+          if current_user.admin?
+            html IndexPage, users: UserQuery.new
+          else
+            redirect Home::Index
+          end
+        end
```
paulcsmith, line 133: Added a specific example for when there is a nil returned

src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
...
+        ▸ Ensure all conditionals (like if/else) return a response with html/redirect/json/etc.
+
+      For example...
+
+        get "/users" do
+          # Render html, json, or redirect
+          html IndexPage, users: UserQuery.new
+          redirect to: Home::Index
+          json({name: "James"})
+        end
```
paulcsmith, line 164: Added an example here too

src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
     response.print
   end
 
+  private def handle_response(_response : Nil) forall T
+    # This is a hack to force Crystal to regenerate this method for
+    # each action. That way @type refers to the action with the problem,
+    # and not the first action to be compiled.
+    {% @type %}
```
paulcsmith, line 110: Thanks @asterite for the tip! And thanks @jwoertink for digging into this

asterite: This is not needed because type is mentioned a bit below.

I think type need to be mentioned in the method that calls this method.

asterite: in perform_action

paulcsmith: 🤦‍♂ I happened to check this with an action that worked with the broken version. Thanks @asterite. Will give that a shot

paulcsmith: Hmm this caused a compiler bug:

```crystal
BUG: trying to downcast SignUps::New <- BrowserAction (Exception)
web    |   from Crystal::CodeGenVisitor#downcast_distinct<LLVM::Value, Crystal::Type+, Crystal::Type+>:NoReturn
```

But I think I had an idea that might work...we'll see :P

src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
...
+      raise <<-ERROR
+
+
+      #{@type} returned Nil
+
+      But it should return a Lucky::Response.
+
+      Try this...
+
+        ▸ Ensure the action returns a response with html/redirect/json/etc.
```
edwardloveall, line 121: This felt like a path to me for a moment, what do you think about:
```suggestion
        ▸ Ensure the action returns a response with html, redirect, json, etc.
```

src/lucky/renderable.cr
edwardloveall: Love it

src/lucky/renderable.cr
edwardloveall: I liked your realistic example above instead of showing all possible options. I think showing a different example could be good, but I think it should still work if it's an example. 🙂

src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
...
+    # This is a hack to force Crystal to regenerate this method for
+    # each action. That way @type refers to the action with the problem,
+    # and not the first action to be compiled.
+    {% @type %}
     {%
       raise <<-ERROR
 
 
-      Your action returned #{T}
+      #{@type} returned #{T}
```
jwoertink, line 148: Isn't this what we had before and it didn't work? 🤔

src/lucky/renderable.cr
```diff
@@ -103,19 +103,66 @@ module Lucky::Renderable
...
 
-        ▸ Use a method like render/redirect/json at the end of your action.
-        ▸ Ensure all conditionals (like if/else) return a response with render/redirect/json/etc.
+        ▸ Use a method like html/redirect/json at the end of your action.
+        ▸ Ensure all conditionals (like if/else) return a response with html/redirect/json/etc.
+
+      For example...
+
+        get "/users" do
+          # Render html, json, or redirect
```
jwoertink, line 160:
```suggestion
          # Return one of html, json, or redirect
```

src/lucky/renderable.cr
jwoertink: Same ❤️

src/lucky/renderable.cr
paulcsmith: Yeah I was being dumb https://github.com/luckyframework/lucky/pull/1086/files#r403641754

And I couldn't figure out a way to get around it because there was a crystal bug in the compiler. Doing what Ary suggested did work when you didn't return a Lucky::Response, but when you the compiler blew up 🤷‍♂ So I guess this is still not possible. I think I'll keep the other changes and leave the other issue open for one day when we can do what we'd like or figure out a workaround :P

src/lucky/renderable.cr
paulcsmith: Good idea!

src/lucky/renderable.cr
paulcsmith: Noice

src/lucky/renderable.cr
paulcsmith: Yeah good point!

PR #1096, opened 2020-04-11
Put components in folder so they're easier to find

paulcsmith:
We need to link to these so people can easily copy/paste them so I moved them to their own folder. Maybe if/when we add more default components we move these all to `src/lucky/components` and use a `Lucky::Components` namespace.

For now the name remains the same though so it works with Lucky 0.20


paulcsmith:
Gonna merge so I can link to it from the blog and because this is a very low risk change

PR #1103, opened 2020-04-12
Serialize UUIDs by default, fixes #1069

### LINKED ISSUE #1069: Using UUID in serializers doesn't seem to work
I thought we had fixed this, but I'm working on a new Lucky 0.19 app, and it's not working:

```crystal
class TeamSerializer < BaseSerializer
  def initialize(@team : Team)
  end

  def render
    {
      id: @team.id, # this has to be turned in to a string
      name: @team.name,
      city: @team.city
    }
  end
end
```

```
Which expanded to:

 > 1 |
 > 2 |         json.field "id" do
 > 3 |           self[:id].to_json(json)
                           ^------
Error: no overload matches 'UUID#to_json' with type JSON::Builder

Overloads are:
 - Object#to_json(io : IO)
 - Object#to_json()
```

hanneskaeufler:
## Purpose

See #1069.

It's actually just a matter of using the right include,
but since I had to hunt that down in the docs as well
(https://crystal-lang.org/api/0.34.0/UUID.html#new(pull:JSON::PullParser)-class-method)
I'd say it is a win for lucky to include that for the serializer
by default.

## Checklist

### DIFF
--- a/src/lucky/serializer.cr
+++ b/src/lucky/serializer.cr
@@ -1,3 +1,5 @@
+require "uuid/json"
+
 abstract class Lucky::Serializer
   def to_json(io)
     render.to_json(io)
--- a/spec/lucky/serializer_spec.cr
+++ b/spec/lucky/serializer_spec.cr
@@ -18,6 +18,15 @@ private class TestNestedJSON < Lucky::Serializer
   end
 end
 
+private class TestUuidJSON < Lucky::Serializer
+  def initialize(@id : UUID)
+  end
+
+  def render
+    {id: @id}
+  end
+end
+
 describe Lucky::Serializer do
   describe "#to_json" do
     it "calls to_json on the render data" do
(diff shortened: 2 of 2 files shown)


src/lucky/serializer.cr
```diff
@@ -1,3 +1,5 @@
+require "uuid/json"
```
jwoertink, line 1: Nice! I didn't know this was a thing. I wonder if this should be included at a higher level so it's accessible in other areas of Lucky 🤔

hanneskaeufler:
My pleasure :) I first put it in uuid_extensions.cr but ultimately decided for the serializer because that‘s where the context of json serialization is. If you‘d want to use it somewhere else I‘d almost favor putting the include in your app code yourself. Decisons decisions 😝

src/lucky/serializer.cr
paulcsmith: YES! Thanks for tracking that down @hanneskaeufler. Definitely agree we should just do that for people

Maybe could be in another spot, but I think it's probably ok here and we can move it if needed.

PR #1104, opened 2020-04-12
Raise helpful message on invalid key, fixes #426

### LINKED ISSUE #426: Raise helpful error if key length is too short in MessageEncryptor

hanneskaeufler:
## Purpose

See #426 

When setting an invalid key, an error deep from the
cookie jar and then ssl cipher is not exactly
helpful to the user. We can rescue from the
error deep down and raise a more helpful one.

## Description

The output rendered to the console as well as the error page doesn't actually look that great. If you have any ideas on how to improve that, please let me know :)

Bildschirmfoto 2020-04-13 um 12 31 15

Bildschirmfoto 2020-04-13 um 12 32 01

## Checklist

### DIFF
--- a/src/lucky/support/message_encryptor.cr
+++ b/src/lucky/support/message_encryptor.cr
@@ -30,7 +30,7 @@ module Lucky
     def encrypt(value) : Bytes
       cipher = OpenSSL::Cipher.new(@cipher_algorithm)
       cipher.encrypt
-      cipher.key = @secret
+      set_cipher_key(cipher)
 
       # Rely on OpenSSL for the initialization vector
       iv = cipher.random_iv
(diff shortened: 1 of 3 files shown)


hanneskaeufler:
I keep thinking tho that we're better served tackling this at the point where `= "key"` is assigned in the config file. Probably by either assigning a custom type rather than a string or by adding some kind of validation options to habitat? I feel like this would then allow us to catch that mistake at either compile-time or at least at app startup rather than at runtime ...

src/lucky/support/message_encryptor.cr
```diff
@@ -49,13 +50,33 @@ module Lucky
...
+        raise InvalidSecretKeyBase.new(<<-MESSAGE
+          It appears the key you set for Lucky::Server.setting.secret_key_base is invalid:
+
+            #{error.message.colorize(:red)}
+
+          Please double check your #{"config/server.cr".colorize(:green)}. You can generate a secret key by entering the following command in your terminal:
+
+            #{"▸".colorize(:green)} lucky gen.secret_key
+        MESSAGE
+        )
```
paulcsmith, line 75: I think starting with a shorter summary helps, also colorizing doesn't work for errors as well since they are in red by default. Thoughts on this?

```suggestion
        raise InvalidSecretKeyBase.new(<<-MESSAGE
          The secret key is invalid:

            #{error.message}

          You can generate a new key using 'lucky gen.secret_key' in your terminal:

             ▸ lucky gen.secret_key
            
          Usually the key is set in 'config/server.cr'
              Lucky::Server.configure do |settings|
                settings.secret_key_base = Lucky::SecretKey.new("your-new-key") 
              end
              
              
          MESSAGE
        )
```

I also usually add a couple newlines at the end to give it some breathing room

hanneskaeufler:
@paulcsmith I took your suggestions to improve the message, thanks! I wouldn't hold my breath to getting to tackle the validation elsewhere, so I'd rather take the win as-is for now.

paulcsmith:
@hanneskaeufler Sounds good to me! Thank you so much!

PR #1108, opened 2020-04-16
Added support for array in paginator

### LINKED ISSUE #1106: Add support for arrays in paginator
I have arrays of data that i would like to paginate that is not coming from a database.

It would be nice, if it is not, to be able to use the paginate for normal arrays with any class (`T`)

my example would be this:
```crystal
  def paginate(
    items : Array(T),
    per_page : Int32 = paginator_per_page
  ) : Tuple(Paginator, Array(T)) forall T
    pages = Paginator.new \
      page: paginator_page,
      per_page: per_page,
      item_count: items.size,
      full_path: context.request.resource

    updated_query = items[pages.offset...pages.offset+pages.per_page]
    {pages, updated_query}
  end
```

confact:
## Purpose
Adding array support in paginator. Fixes #1106

## Checklist

### DIFF
--- a/spec/paginator/backend_helpers_spec.cr
+++ b/spec/paginator/backend_helpers_spec.cr
@@ -27,6 +27,10 @@ class Paginatable
     paginate(FakeUser::BaseQuery.new)
   end
 
+  def call_array
+    paginate_array(FakeUser::BaseQuery.new.to_a)
+  end
+
   def params
     Params.new(@page)
   end
@@ -76,6 +80,24 @@ describe Lucky::Paginator::BackendHelpers do
     pages.total.should eq(2)
     records.query.offset.should eq(25)
     records.query.limit.should eq(25)
+  end
+
+  it "accept array with default" do
+    pages, records = Paginatable.new.call_array
+
+    pages.page.should eq(1)
+    pages.per_page.should eq(25)
+    pages.total.should eq(2)
+    records.size.should eq(2)
+  end
+
+  it "uses array with the 'page' param if given" do
+    pages, records = Paginatable.new(page: "2").call_array
+
+    pages.page.should eq(2)
+    pages.per_page.should eq(25)
+    pages.total.should eq(2)
+    records.size.should eq(2)
   end
 
   it "allows overriding 'paginator_page' and 'paginator_per_page'" do
--- a/spec/paginator/backend_helpers_spec.cr
+++ b/spec/paginator/backend_helpers_spec.cr
@@ -100,7 +100,7 @@ describe Lucky::Paginator::BackendHelpers do
     records.size.should eq(25)
   end
 
-  it "return empty array if page is set above array size" do
+  it "return empty array if page is overflowed" do
     pages, records = Paginatable.new(page: "3").call_array
 
     pages.page.should eq(3)
--- a/spec/paginator/backend_helpers_spec.cr
+++ b/spec/paginator/backend_helpers_spec.cr
@@ -28,7 +28,7 @@ class Paginatable
   end
 
   def call_array
-    paginate_array(FakeUser::BaseQuery.new.to_a)
+    paginate_array([1]*50)
   end
 
   def params
@@ -88,7 +88,7 @@ describe Lucky::Paginator::BackendHelpers do
     pages.page.should eq(1)
     pages.per_page.should eq(25)
     pages.total.should eq(2)
-    records.size.should eq(2)
+    records.size.should eq(25)
   end
 
   it "uses array with the 'page' param if given" do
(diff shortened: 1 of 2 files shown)


confact:
@paulcsmith don't understand why i get that test error. That the db is not connected on only my two tests.

spec/paginator/backend_helpers_spec.cr
```diff
@@ -27,6 +27,10 @@ class Paginatable
     paginate(FakeUser::BaseQuery.new)
   end
 
+  def call_array
+    paginate_array(FakeUser::BaseQuery.new.to_a)
```
jwoertink, line 31: Looks like your spec fails because this actually makes the query to the DB, but this DB doesn't actually exist. Since this is just for an array, we should be able to make this a static array. If it works for static array, then it should work for a query turned into an array.

Alternatively, we would have to setup a real database for this project.

paulcsmith: Good catch and good suggestion! I think a static array would be a great fit here. I'd prefer not add a whole database setup for just these tests

```suggestion
    paginate_array([1,2,3])
```

paulcsmith: Or if you need a ton of items, `[1] * 50`

src/lucky/paginator/backend_helpers.cr
```diff
@@ -40,6 +40,47 @@ module Lucky::Paginator::BackendHelpers
...
+  # This method will return a `Lucky::Paginator` object and the requested page
+  # of items.
+  #
+  # ## Examples
+  #
+  # ```crystal
+  # class ListItems::Index < BrowserAction
+  #   get "/items" do
+  #     # The 'Array' will just show items for the requested page
+  #     # because 'paginate' will add a 'limit' and 'offset' to the query.
```
paulcsmith, line 54: This line can probably be removed

src/lucky/paginator/backend_helpers.cr
```diff
@@ -40,6 +40,47 @@ module Lucky::Paginator::BackendHelpers
...
+  #     pages, items = paginate_array([1, 2, 3])
+  #     render IndexPage, pages: pages, items: items
+  #   end
+  # end
+  #
+  # class Users::IndexPage < MainLayout
+  #   needs pages : Lucky::Paginator
+  #   needs items : Array(T)
```
paulcsmith, line 62: I think this was trying to show it can be any array, but I think it may confuse people since this wouldn't compile. Using the Array(Int32) may make it clearer

```suggestion
  #   needs items : Array(Int32)
```

src/lucky/paginator/backend_helpers.cr
```diff
@@ -40,6 +40,47 @@ module Lucky::Paginator::BackendHelpers
...
+  #
+  #   def content
+  #     # Render 'items' like normal
```
paulcsmith, line 65:
```suggestion
  #     # Render pagination links for the 'items' Array
```

src/lucky/paginator/backend_helpers.cr
confact: Agreed, I have changed this to Int32. Thanks!

spec/paginator/backend_helpers_spec.cr
confact: I chose `[1]*50` - Seems to be a nice way to make it work for all tests. Thanks!

src/lucky/paginator/backend_helpers.cr
```diff
@@ -40,6 +40,48 @@ module Lucky::Paginator::BackendHelpers
...
+    items : Array(T),
+    per_page : Int32 = paginator_per_page
+  ) : Tuple(Paginator, Array(T)) forall T
+    pages = Paginator.new \
+      page: paginator_page,
+      per_page: per_page,
+      item_count: items.size,
+      full_path: context.request.resource
+
+    return {pages, Array(T).new} if pages.offset > items.size
```
paulcsmith, line 79:
```suggestion
    return {pages, Array(T).new} if pages.overflowed?
```

spec/paginator/backend_helpers_spec.cr
```diff
@@ -78,6 +82,33 @@ describe Lucky::Paginator::BackendHelpers do
...
+  it "uses array with the 'page' param if given" do
+    pages, records = Paginatable.new(page: "2").call_array
+
+    pages.page.should eq(2)
+    pages.per_page.should eq(25)
+    pages.total.should eq(2)
+    records.size.should eq(25)
+  end
+
+  it "return empty array if page is set above array size" do
```
paulcsmith, line 103:
```suggestion
  it "return empty array if page is overflowed" do
```

PR #1109, opened 2020-04-18
Start on moving to new Dexter

paulcsmith:
Moves everything to Dexter 0.2.0 and Crystal `Log`.

Updates PrettyLogFormatter to work with Dexter and `Log::Context`

Also updates exception formatting to be easier to parse and properly indented. Also dims non-app lines

Screen Shot 2020-04-19 at 12 02 46 PM
Screen Shot 2020-04-19 at 12 03 46 PM

### DIFF
--- a/src/lucky.cr
+++ b/src/lucky.cr
@@ -21,11 +21,6 @@ require "./lucky/paginator/*"
 require "./lucky/paginator/components/*"
 
 module Lucky
-  Habitat.create do
-    setting logger : Dexter::Logger
-  end
-
-  def self.logger
-    settings.logger
-  end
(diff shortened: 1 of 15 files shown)


src/lucky.cr
```diff
@@ -21,11 +21,6 @@ require "./lucky/paginator/*"
...
 
 module Lucky
-  Habitat.create do
-    setting logger : Dexter::Logger
-  end
-
-  def self.logger
-    settings.logger
-  end
+  Log              = ::Log.for("lucky")
```
paulcsmith, line 24: Technically you can pass `self` but I like the idea of passing a string so we know exactly what the source name is

PR #1114, opened 2020-04-19
Simplify error styling, add human http status

paulcsmith:
Screen Shot 2020-04-19 at 3 14 13 PM
Screen Shot 2020-04-19 at 3 14 35 PM
Screen Shot 2020-04-19 at 2 39 17 PM

### DIFF
--- a/src/lucky/logger_helpers.cr
+++ b/src/lucky/logger_helpers.cr
@@ -1,15 +1,16 @@
 module Lucky::LoggerHelpers
-  def self.colored_status_code(status_code : Int32) : String
+  def self.colored_http_status(status_code : Int32) : String
+    status_name = Wordsmith::Inflector.humanize(HTTP::Status.from_value?(status_code) || "")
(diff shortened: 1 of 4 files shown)


src/lucky/logger_helpers.cr
```diff
@@ -1,15 +1,16 @@
 module Lucky::LoggerHelpers
-  def self.colored_status_code(status_code : Int32) : String
+  def self.colored_http_status(status_code : Int32) : String
+    status_name = Wordsmith::Inflector.humanize(HTTP::Status.from_value?(status_code) || "")
+    message = "#{status_code} #{status_name}".colorize.bold
+
     case status_code
-    when 200..399
```
paulcsmith, line 4: Most things *should* be green. So rather than distracting with a bold green status, just make it white. That way when a real errors pops up it stands out.

Note: We log halted pipes in yellow so redirect due to incorrect permissions still stands out

PR #1118, opened 2020-04-23
Create url safe password reset tokens.

### LINKED ISSUE #1092: Bug in password reset token
I've been testing our password reset flow and ran into a few 500 errors. At first, I thought it was on my side, but it turns out there is a bug caused by the generated tokens.

If a token contains a `+`, it is interpreted as whitespace. Here is an example:

```
GSSuuPIePHZIDu4fx/ugiIcVX3jQratcnG6V0MBgJ5M=--xm02SWDRKUUIzaf8P+fPub95NWo=

---------------------------------------------------------------^
```

At the error page, it shows the token without the `+`:

lucky-error

wout:
I'm sorry it has taken this long, I've been really busy the past two weeks. :blush: 

## Purpose
Fixes #1092.

## Checklist

### DIFF
--- a/src/lucky/support/message_verifier.cr
+++ b/src/lucky/support/message_verifier.cr
@@ -39,7 +39,7 @@ module Lucky
     end
 
     private def encode(data) : String
-      ::Base64.strict_encode(data)
+      ::Base64.urlsafe_encode(data)
     end
 
     private def decode(data) : Bytes


wout:
The trailing `=`'s are padding to make sure the length of the string is divisible by four. They're just there in case a string gets concatenated. They can be removed, but to what avail? They don't pose any danger unless you have a very unfortunate situation with query params and a `?` involved.

In what situation did you run into any issues?

jwoertink:
ah! Ok, that makes sense. I never knew what they were for. I'm cool with keeping them then. I don't remember the issue since it's been a while. I just remember stripping those off for some reason. It was probably just a poor hack 😅

PR #1121, opened 2020-04-26
Add route prefix macro and specs

### LINKED ISSUE #1087: Add `route_prefix` macro for prefixing/namespacing routes
https://github.com/luckyframework/lucky/discussions/834#discussioncomment-3323

```crystal
abstract class ApiAction < Lucky::Action
  route_prefix "/api/v1"
end
```

Would prefix routes that inherit from this class with `/api/v1` so you don't have to do it for every single route

rnice01:
Fixes #1087

## Purpose
This PR is for issue #1087 for adding a `route_prefix` macro for Lucky::Actions. This will prefix all the routes added by the HTTP method and match macros.

## Checklist


rnice01:
> This looks really good! Just to make sure we don't have any regressions could you add a couple extra specs?
> 
> * Layered inheritance. So maybe a `abstract class TestApiPrefixAction < TestAction` and then another action `ChildApiPrefixAction < TestApiPrefixAction`. It could have just one route and the spec would check that it still correctly prefixed the route
> * Modules. It may be helpful/common for people to make a route prefix module that is added to multiple actions. Could you create a `ApiPrefix` module and then an action that includes it and see if the action gets the prefix correctly?
> 
> Something like:
> 
> ```crystal
> module ApiPrefix
>   macro included
>     route_prefix "/api"
>   end
> end
> ```
> 
> Does that make sense?

@paulcsmith, for the Layered Inheritance spec, would the abstract class use the route_prefix macro and test the child correctly prefixes it's own routes from the parent's defined prefix?

`
abstract class ApiTestPrefixAction < TestAction
  route_prefix "/api"
end

class ChildApiTestPrefixAction < ApiTestPrefixAction
  get "/test" do
    #make sure this gets prefixed
  end
end
`

paulcsmith:
@rnice01 Yeah that's exactly what I had in mind! 👍

jwoertink:
oh man! Stoked for this ❤️

Issue #1070, opened 2020-03-30
Add params to lucky routes


jwoertink:
I'm not sure if we have access to this or not, but I was just thinking how cool it would be if you ran `lucky routes`, and there was another column that displayed which params you needed on that action. That might be kind of cool for a quick api documentation type deal. 

```text
╔══════╦════════════════════════════════╦════════════════════════════════════════╗
║ Verb | URI                            | Action               | Params          ║
╠──────┼────────────────────────────────┼──────────────────────┼─────────────────╣
║ POST | /v1/authenticate               | Authenticate::Create | token : String  ║
╠──────┼────────────────────────────────┼──────────────────────┼─────────────────╣
```

Though... now after making this example, I just realized on smaller (more standard) screens, this could become a mess real quick with multiple params 😞  But maybe some has an idea?

paulcsmith:
@jwoertink Cool idea! So specifically query params right? Since the path shows which named params are in it?

Another thought I had was maybe not using a table. Or making the table optional (useful when grepping)

```
POST /v1/authenticate
  Action > Authenticate::Create
  Params > token : String, page : Int32?, etc.
  Route helper > Authenticate::Create.with(token: "String") # Example route 
```

But with some colors and a nicer arrow. I think this is actually a bit easier to scan since you don't need to read a whole row. Would love to hear opinions on it though

jwoertink:
oooo I do actually kind of like that. But yeah, query params defined from `param` method.

PR #1122, opened 2020-04-27
Adding a new method to action classes to return permitted param declarations

jwoertink:
## Purpose
This is a precursor to #1070 

## Description
Before this PR there was no way to know what params have been defined on an Action class. Also interesting thing, calling something like `Users::Index::PARAM_DECLARATIONS` would throw an exception that `Crystal::Macros::TypeDeclaration` was undefined. I guess that constant only lives inside of macros...

Well, now we have a method called `permitted_query_params` on the class. This will be useful for adding which params are allowed in a route when running `lucky routes`. 

## Checklist

### DIFF
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -224,19 +224,19 @@ describe Lucky::Action do
     end
   end
 
-  describe ".permitted_query_params" do
+  describe ".query_param_declarations" do
     it "returns an empty array" do
-      PlainText::Index.permitted_query_params.size.should eq 0
+      PlainText::Index.query_param_declarations.size.should eq 0
     end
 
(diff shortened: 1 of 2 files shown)


spec/lucky/action_spec.cr
```diff
@@ -224,6 +224,22 @@ describe Lucky::Action do
     end
   end
 
+  describe ".permitted_query_params" do
```
paulcsmith, line 227: How about query_param_declarations? Permitted query params makes me think of operations and that it’ll return the params somehow. I think param declarations might be a bit clearer

Thoughts?

jwoertink: I struggled with this name lol. Yeah, that sounds good to me 👍

PR #1123, opened 2020-04-27
Cleanup/ameba

jwoertink:
Looks like we're moving to github actions for everything. One thing we can do is make sure all the code is formatted with ameba to keep things consistent. This gets that part rolling.

Note: I added 1 style to disable because ameba complains about `Log` not being `LOG`. We can't just change that, so better to just ignore it.


.ameba.yml
```diff
@@ -0,0 +1,10 @@
+# This configuration file was generated by `ameba --gen-config`
+# on 2020-04-27 21:05:58 UTC using Ameba version 0.12.0.
+# The point is for the user to remove these configuration records
+# one by one as the reported problems are removed from the code base.
+
+# Problems found: 2
+# Run `ameba --only Style/ConstantNames` for details
```
edwardloveall, line 7: Think we could fix this?

jwoertink: You mean so we don't have to have this disabled? This catches `Log` and says it should be `LOG`, but that would require everyone updating their apps to be `LOG.info {}`, etc... which doesn't really follow the crystal way. I think there was another one with one of the Pipes that it wanted PIPES...

jwoertink:
Hey @wontruefree I saw the 👎 . I was wondering if I could get some more context on why. Do you disagree with the move to github actions? Or is it using ameba to catch erroneous/irrelevant code? Or is it that the logger should actually be named LOG like ameba originally suggested?

.ameba.yml
edwardloveall: Ah yeah, that's a bummer. Thanks for the explanation. I agree with you keeping it the way it is despite the custom config.

jwoertink: I removed a comment on how many errors it found. It was irrelevant. I do like this comment that tells us how we can run to see those errors though if we ever need.

paulcsmith: I think @veelenga was thinking of changing this in the future so we don’t need the comment

.crystal-version
```diff
@@ -0,0 +1 @@
+0.34.0
```
paulcsmith, line 1: Could you put a new line here?

wontruefree:
@jwoertink ameba has caused some errors in other projects.  I personally like the `crystal tool format` defaults.  I think the changes here are good though.  It is mostly a preference.

.ameba.yml
veelenga: Hey, thanks for mentioning. This is already changed in ameba [`v0.12.1`](https://github.com/crystal-ameba/ameba/releases/tag/v0.12.1). You can safely upgrade.

veelenga:
@wontruefree would you mind sharing the references to these errors? Would love helping/discussing it in a [gitter channel](https://gitter.im/veelenga/ameba)

paulcsmith:
@wontruefree Thanks for the feedback. We won't be adding ameba or format checks to Lucky projects out of the box, this is just for Lucky lib itself so hopefully that helps as well.

I definitely don't want to enforce code style on every project. I think that should be opt-in for sure.

jwoertink:
@wontruefree ah. Yeah, I know what you mean. This will only be for our use internally. Individual Lucky projects would need to add this stuff in as everyone else sees fit. Thanks for the feedback! It's hard to know what a 👎 really means, or if it was just a mistake and hit by accident 😛

wontruefree:
I misunderstood github actions.  Thanks :)

jwoertink:
oh boo. @paulcsmith the format section doesn't have shards installed yet at that point, so it would need to run shards install first, or go in a later section. I'm not familiar with actions, if I install shards in one job, is it available in another? Or are those separate containers that each need their own?

paulcsmith:
@jwoertink they are per job I believe. So you'll need to copy the shards install step. If that doesn't work LMK and I can try to help out more!

PR #1127, opened 2020-04-29
Return error if you try to generate a model that already exists

### LINKED ISSUE #1008: Running gen.model multiple times, generates multiple migrations.
I would expect `gen.model` to check if the model, query, operation and migration already exist. Now, running the same command twice will generate the model, query and operation files the first time, but will not overwrite nor warn the second time. Migrations, on the other hand, are generated every time with a new timestamp.

jwoertink:
## Purpose
Fixes #1008

## Description
For those sleepy late programming moments!

## Checklist

### DIFF
--- a/tasks/gen/model.cr
+++ b/tasks/gen/model.cr
@@ -61,7 +61,7 @@ class Gen::Model < LuckyCli::Task
   end
 
   private def resource_name_not_taken
-    @error = "A model by the name '#{resource_name.camelcase}' already exists."
+    @error = "'#{resource_name.camelcase}' model already exists at #{"./src/models/#{template.underscored_name}.cr"}."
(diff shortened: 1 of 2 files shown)


tasks/gen/model.cr
```diff
@@ -59,6 +60,11 @@ class Gen::Model < LuckyCli::Task
     columns_are_valid?
   end
 
+  private def resource_name_not_taken
+    @error = "A model by the name '#{resource_name.camelcase}' already exists."
```
paulcsmith, line 64:
```suggestion
    @error = "'#{resource_name.camelcase}' model already exists at #{"./src/models/#{template.underscored_name}.cr"}."
```

I think it'd be good to show *why* it thinks that it exists since it is possible the file name is just wrong. This might give them an indication as to where they need to look

PR #1128, opened 2020-04-29
Implementing a new UI for lucky routes 

### LINKED ISSUE #1070: Add params to lucky routes

jwoertink:
## Purpose
Fixes #1070

## Description
Screen Shot 2020-04-29 at 2 10 20 PM

I wanted to keep it in the table, but on smaller screens, the table rows ended up wrapping when you had too many params (like a search form). This is a new UI that includes a bit more information. My thought is once the native task args are in place, we could add a compact flag to print these out in a smaller more concise manner if people wanted.

I open to suggestions on how to make this look really pretty. When you run `lucky routes`, your first thought should be "wow, this is amazing!". So any ideas anyone has would be awesome.

## Checklist

### DIFF
--- a/tasks/routes.cr
+++ b/tasks/routes.cr
@@ -1,21 +1,58 @@
 require "lucky_cli"
 require "colorize"
-require "shell-table"
 
 class Routes < LuckyCli::Task
   summary "Show all the routes for the app"
 
   def call
-    routes = Lucky::Router.routes.map do |route|
-      [route.method.to_s.upcase, route.path.colorize.green, route.action]
+    table = String.build do |output|
+      Lucky::Router.routes.map do |route|
+        output << "#{route.method.to_s.upcase} #{route.path.colorize.green}\n"
+        output << "  Action       ▸ #{route.action}\n"
+        if has_query_params?(route.action)
+          output << "  Query params ▸ #{query_param_display(route.action)}\n"
+        end
+        output << "  Route helper ▸ #{route_helper_display(route.action)}\n"
+        output << "\n\n"
+      end
     end
 
-    table = ShellTable.new(
-      labels: ["Verb", "URI", "Action"],
-      label_color: :blue,
-      rows: routes
-    )
-
     puts table
+  end
+
+  private def has_query_params?(action : Lucky::Action.class) : Bool
+    action.query_param_declarations.any?
+  end
+
+  private def param_declaration(declaration : String) : Array(String)
+    declaration.split(" : ")
+  end
+
+  private def query_param_display(action : Lucky::Action.class) : String
+    action.query_param_declarations.map { |declaration|
+      name, type = param_declaration(declaration)
+      "#{name} : #{type.colorize.yellow}"
+    }.join(", ")
+  end
+
+  private def route_helper_display(action : Lucky::Action.class) : String
+    if has_query_params?(action)
+      first_param = action.query_param_declarations.first
+      name, type = param_declaration(first_param)
+      %{#{action}.with(#{name}: "#{example_param_by_type(type)}")}
+    else
+      "#{action}.route"
+    end
+  end
+
+  private def example_param_by_type(type : String) : String
+    case type
+    when .includes?("Int")
+      "4"
+    when .includes?("Bool")
+      "false"
+    else
+      "example"
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


edwardloveall:
I dig it. I really like what @paulcsmith did with https://github.com/luckyframework/lucky/pull/1114 where the top-level sections got a contrasting background color. Maybe we could do that here with the verb and path? That way you can really easily scan down the list. Another idea that I feel less confident about is different colors for `GET` (green?), `POST`/`PUT`/`PATCH` (yellow or blue), and `DELETE` (red or gray).

jwoertink:
How's this, @edwardloveall ?

Screen Shot 2020-04-30 at 9 54 57 AM

Screen Shot 2020-04-30 at 9 56 11 AM

jwoertink:
hmm... I've been staring at these routes, and I'm noticing it's a little harder to see the path when scanning across a ton of routes. It's not horrible, but doesn't seem to stick out right away for me.

paulcsmith:
What if we use the same styling from the logs and go with non-style method name and a bold and underlined path name. So it even looks like a link?

I also worry about how this will wrap when there are a lot of args. I wonder if the params can be shown vertically and we combine route and query params into one:

```crystal
GET /users/:user_id.colorize.bold.underline  ▸.colorize.dim Users::Show.colorize.bold
  ▸.colorize.dim  user_id : String
  ▸.colorize.dim  page : Int32
```

I think this is still pretty easy to parse, handles up to a kazilion params. It doesn't show a route helper but maybe that is ok since there are so many ways to use them (`.with`, `.url`, `.path`, `.url_without_query_params`, etc.). So maybe just showing all the params with the type is good

Thoughts

paulcsmith:
Maybe we link to docs on how to use routes and params at the top?

jwoertink:
Ok, here's an updated version with @paulcsmith suggestions:

Screen Shot 2020-04-30 at 3 31 50 PM
Screen Shot 2020-04-30 at 3 31 17 PM

paulcsmith:
I like it! I’d maybe do the indent just 1 space and bold the action. But otherwise I like that it’s compact and easy to scan

jwoertink:
Ok, this might be a little overkill, but I felt it needed a little color somewhere 😅 

Screen Shot 2020-04-30 at 4 42 36 PM

paulcsmith:
That might be a tad bit too much 😂 what do you think about making the arrow in between the path and action green? Or maybe the action itself can be green and bold so it is easier to scan the list of routes

jwoertink:
Ok, I'm feeling good about this.

Screen Shot 2020-04-30 at 8 45 31 PM
Screen Shot 2020-04-30 at 8 45 16 PM

paulcsmith:
@edwardloveall Yeah I think it could work. My concern is with say, 10 or 15 routes on the scree it could get a bit overwhelming. Maybe we merge this and try it some stuff out locally with some more routes and see what it looks like?

> but I'm also aware that I'm bikeshedding at this point

It's not Lucky if it's not bikeshedding! But seriously I think bikeshedding is not so bad. Sometimes really nice stuff comes from it. As long as we are ok shipping and then continuing the discussion later I think it is fine!

tasks/routes.cr
```diff
@@ -1,21 +1,33 @@
...
+
+  private def dim_arrow
+    "▸".colorize.dim
+  end
+
+  private def print_banner_message
+    <<-TEXT.colorize.dim
+
+    Routing documentation:
+      https://luckyframework.org/guides/http-and-routing/routing-and-params
```
paulcsmith, line 30:
```suggestion
    Routing documentation:
    
      https://luckyframework.org/guides/http-and-routing/routing-and-params
```

PR #1135, opened 2020-05-07
Escape HTML in HTML helpers by default

paulcsmith:
## Purpose

Safer output

I will make a patch release with these changes. It potentially breaks apps if they relied on unescaped text, but my guess is few people are doing that and it is better safe than sorry

### Impact

Since Lucky disallows reading cookies by default: https://github.com/luckyframework/lucky_cli/blob/4ec6de7cfc00be018f328ca8b7fd509cf9118c3d/src/web_app_skeleton/config/cookies.cr.ecr#L12-L13

This issue should have minimal security impact unless users use legacy browsers that do not accept the `http_only` cookie setting.

Either way, escaping by default is a much safer way to go since it is a second layer of protection, and makes it so people can't mess up your layout with renegade HTML

### DIFF
--- a/spec/lucky/text_helpers/highlight_spec.cr
+++ b/spec/lucky/text_helpers/highlight_spec.cr
@@ -41,13 +41,17 @@ describe Lucky::TextHelpers do
       view(&.highlight("wow em", %w(wow em), highlighter: "<em>\\1</em>")).should eq %(<em>wow</em> <em>em</em>)
     end
 
-    it "highlights with html" do
-      view(&.highlight("<p>This is a beautiful morning, but also a beautiful day</p>", "beautiful")).should eq "<p>This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day</p>"
-      view(&.highlight("<p>This is a <em>beautiful</em> morning, but also a beautiful day</p>", "beautiful")).should eq "<p>This is a <em><mark>beautiful</mark></em> morning, but also a <mark>beautiful</mark> day</p>"
-      view(&.highlight("<p>This is a <em class=\"error\">beautiful</em> morning, but also a beautiful <span class=\"last\">day</span></p>", "beautiful")).should eq "<p>This is a <em class=\"error\"><mark>beautiful</mark></em> morning, but also a <mark>beautiful</mark> <span class=\"last\">day</span></p>"
-      view(&.highlight("<p class=\"beautiful\">This is a beautiful morning, but also a beautiful day</p>", "beautiful")).should eq "<p class=\"beautiful\">This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day</p>"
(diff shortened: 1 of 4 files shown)


paulcsmith:
Thanks for reporting this @skojin

spec/lucky/text_helpers/highlight_spec.cr
```diff
@@ -41,13 +41,17 @@ describe Lucky::TextHelpers do
...
-      view(&.highlight("<p class=\"beautiful\">This is a beautiful morning, but also a beautiful day</p>", "beautiful")).should eq "<p class=\"beautiful\">This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day</p>"
-      view(&.highlight("<p>This is a beautiful <a href=\"http://example.com/beautiful\#top?what=beautiful%20morning&when=now+then\">morning</a>, but also a beautiful day</p>", "beautiful")).should eq "<p>This is a <mark>beautiful</mark> <a href=\"http://example.com/beautiful\#top?what=beautiful%20morning&when=now+then\">morning</a>, but also a <mark>beautiful</mark> day</p>"
-      view(&.highlight("<div>abc div</div>", "div", highlighter: "<b>\\1</b>")).should eq "<div>abc <b>div</b></div>"
+    it "escapes HTML by default" do
+      view(&.highlight("<span>wow</span>", "wow")).should eq %(&lt;span&gt;<mark>wow</mark>&lt;/span&gt;)
+    end
+
+    it "allows unescaped HTML" do
+      view(&.highlight("<p>This is a beautiful morning, but also a beautiful day</p>", "beautiful", escape: false)).should eq "<p>This is a <mark>beautiful</mark> morning, but also a <mark>beautiful</mark> day</p>"
```
edwardloveall, line 49: Would it be good to warn the user? I.e. having the param be something like `unsafe_escape`?

skojin: I think it confuse, because escape is safe, but not escape is unsafe, ```escape: false``` is ok

if make warning parameter name, then ```raw_unsafe: true || unsafe_raw: true``` is less confuse, but I personally prefer ```escape: false```

paulcsmith: I think I am of the same opinion as @skojin.  I think if I saw `unsafe_escape: false` I'd be a bit confused because the alternative would I guess mean a safe_escape. So it almost seems like either way it is escaping, but one isn't safe.

I'll go with `escape` for now, but am open to other ideas. Maybe `raw: true`? Not sure 🤷‍♂️

paulcsmith:
Thanks for the reviews!

spec/lucky/text_helpers/highlight_spec.cr
skojin: I think better option is not allow this helper write to output directly, only via ```raw``` helper
```raw truncate(...)```

paulcsmith: Problem is that won’t work when passing a block to render HTML for a “read more” link because if you tried to do `text truncate(text, special_link_block` It will escape the link. Then if someone does `raw` It’ll work but the text will not be escaped unless they manually escape it before giving it to `truncate`

Does that make sense? Maybe there is a way around this but I couldn’t think of one :(

skojin: yes, you escape text before merge it with block

PR #1139, opened 2020-05-13
Improve the Memoizable module

### LINKED ISSUE #1119: Support `Nil` in memoize
https://github.com/luckyframework/lucky_cli/pull/508

It is sometimes useful to memoize things that can be nil. I think we can support this and it'd be great to have. We still shouldn't accept any other union, at least for now. But Nil union is pretty common and can definitely work

matthewmcgarvey:
Fixes #1119

## Purpose

This adds support for memoization of nil responses as well as using the macro with methods with arguments. I've not written any real macros before and this seems to be quite a monster so I have no idea if this is any good or not or if there are any missing edge cases.

It made sense to me to store the cached value in a tuple with all the arguments that are passed. If the method does not take any arguments but returns a string the tuple will just be `Tuple(String)` but if the method takes in a number it would be `Tuple(String, Int32)`. Then, those tuple values that aren't the return value are checked against all passed in args. I didn't expect iterating over the args with a for-loop to automatically handle named arguments but my test seems to show that it does.

### Example

```crystal
class Users::Show < BrowserAction
  get "/users/:user_id" do
    stuff = get_user_and_make_api_call(user_id)
    # other stuff...
  end

  memoized get_user_and_make_api_call(user_id : String) : User
    get_user(user_id) && make_api_call
  end
end
```

becomes...

```crystal
class Users::Show < BrowserAction
  get "/users/:user_id" do
    stuff = get_user_and_make_api_call(user_id)
    # other stuff...
  end

  @__memoized_get_user_and_make_api_call : Tuple(User, String)?

  def get_user_and_make_api_call__uncached(user_id : String) : User
    get_user(user_id) && make_api_call
  end

  def get_user_and_make_api_call__tuple_cached(user_id : String) : Tuple(User, String)
    @__memoized_get_user_and_make_api_call = nil if user_id != @__memoized_get_user_and_make_api_call.try &.at(1)
    @__memoized_get_user_and_make_api_call ||= -> do
      result = get_user_and_make_api_call__uncached(user_id)
      {result, user_id}
    end.call.not_nil!
  end

  def get_user_and_make_api_call(user_id : String) : User
    get_user_and_make_api_call__tuple_cached(user_id).first
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/memoize_spec.cr
+++ b/spec/lucky/memoize_spec.cr
@@ -6,15 +6,21 @@ private class ObjectWithMemoizedMethods
   include Lucky::Memoizable
   getter times_method_1_called = 0
   getter times_method_2_called = 0
+  getter times_method_3_called = 0
 
   memoize def method_1 : String
     @times_method_1_called += 1
     "method_1"
   end
 
-  memoize def method_2 : Int32
+  memoize def method_2 : Int32?
     @times_method_2_called += 1
-    5
+    nil
+  end
+
+  memoize def method_3(arg_a : String, arg_b : String = "default-arg-b") : String
+    @times_method_3_called += 1
+    arg_a + ", " + arg_b
   end
 end
 
@@ -23,11 +29,39 @@ describe "memoizations" do
     object = ObjectWithMemoizedMethods.new
 
     object.method_1.should eq "method_1"
-    2.times { object.method_1 }
+    9.times { object.method_1 }
     object.times_method_1_called.should eq 1
+  end
 
-    object.method_2.should eq 5
+  it "can cache a nil result" do
+    object = ObjectWithMemoizedMethods.new
+
+    object.method_2.should be_nil
     9.times { object.method_2 }
     object.times_method_2_called.should eq 1
+  end
+
+  it "caches based on argument equality" do
+    object = ObjectWithMemoizedMethods.new
+
+    object.method_3("arg-a", "arg-b").should eq("arg-a, arg-b")
+    9.times { object.method_3("arg-a", "arg-b") }
+    object.times_method_3_called.should eq 1
+
+    object.method_3("arg-a", "arg-c").should eq("arg-a, arg-c")
+    9.times { object.method_3("arg-a", "arg-c") }
+    object.times_method_3_called.should eq 2
+
+    object.method_3("arg-a").should eq("arg-a, default-arg-b")
+    9.times { object.method_3("arg-a") }
+    object.times_method_3_called.should eq 3
+
+    object.method_3("arg-a", arg_b: "different").should eq("arg-a, different")
+    9.times { object.method_3("arg-a", arg_b: "different") }
+    object.times_method_3_called.should eq 4
+
+    object.method_3(arg_b: "different", arg_a: "something").should eq("something, different")
+    9.times { object.method_3("something", "different") }
+    object.times_method_3_called.should eq 5
   end
 end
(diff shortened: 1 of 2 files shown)


src/lucky/memoizable.cr
```diff
@@ -15,25 +15,69 @@ module Lucky::Memoizable
...
+      {% end %}
+    ) : {{ method_def.return_type }}
       {{ method_def.body }}
     end
 
-    # Returns cached value
-    def {{ method_def.name }} : {{ method_def.return_type }}
-      @__{{ method_def.name }} ||= -> do
-        {{ method_def.name }}__uncached
+    def {{ method_def.name }}__tuple_cached(
```
jwoertink, line 41: Can we add a comment above this like the others? Maybe "Stores the cache in the tuple" or something?

spec/lucky/memoize_spec.cr
```diff
@@ -23,11 +29,39 @@ describe "memoizations" do
     object = ObjectWithMemoizedMethods.new
 
     object.method_1.should eq "method_1"
-    2.times { object.method_1 }
+    9.times { object.method_1 }
```
paulcsmith, line 32: Could you keep these and others as 2? I think using the bare minimum to find a failure is good because `9` looks like it is a "magic number"

spec/lucky/memoize_spec.cr
```diff
@@ -23,11 +29,39 @@ describe "memoizations" do
...
+    object.method_3("arg-a", "arg-c").should eq("arg-a, arg-c")
+    9.times { object.method_3("arg-a", "arg-c") }
+    object.times_method_3_called.should eq 2
+
+    object.method_3("arg-a").should eq("arg-a, default-arg-b")
+    9.times { object.method_3("arg-a") }
+    object.times_method_3_called.should eq 3
+
+    object.method_3("arg-a", arg_b: "different").should eq("arg-a, different")
+    9.times { object.method_3("arg-a", arg_b: "different") }
```
paulcsmith, line 60: Could you add an `eq` matcher to make sure the result is the correct result when using the memoized version?

src/lucky/memoizable.cr
```diff
@@ -15,25 +15,69 @@ module Lucky::Memoizable
   # then each subsequent call returns the user record.
   #
   # The `memoize` method will raise a compile time exception if you forget to include
-  # a return type for your method, or if your return type is a `Union`.
-  # Arguments are not allowed in memoized methods because these can change the return value.
+  # a return type for your method, or if any arguments are missing a type.
+  # Arguments are allowed but as soon as one changes, the previous value is no longer held on to.
```
paulcsmith, line 19: I'm not totally sure I understand this part. I think it does hold onto the value, but it creates a new value for the new arguments?

I wonder if this might just be expected and so this sentence could be left off. What do you think?

matthewmcgarvey: I can't tell if we're on the same page or not.  Calling the memoized method with new arguments and then with the old arguments will re-run the memoized method. Does the updated description make this more clear?

paulcsmith: Ohhhh, so if you do `my_method(1)` and then `my_method(2)`. `my_method(2)` will *overwrite* the cached value. So if you do `my_method(1)` again it will re-run the method.

Is that right? If so I think I got it and the doc changes look good!

matthewmcgarvey: Exactly 👍

paulcsmith:
Thank you :D 🚀 ❤️

PR #1159, opened 2020-05-29
Faster String#squish

### LINKED ISSUE #1157: Faster String#squish
Possible solution: https://github.com/luckyframework/lucky/compare/master...jamescook:james/faster-squish

```
squish original (5600 bytes)   4.48k (223.27µs) (± 1.02%)  60.4kB/op   9.52× slower
     squish new (5600 bytes)  42.63k ( 23.46µs) (± 1.40%)  16.3kB/op        fastest

squish original (44 bytes) 669.65k (  1.49µs) (± 1.60%)  528B/op   5.63× slower
     squish new (44 bytes)   3.77M (265.32ns) (± 5.51%)  272B/op        fastest

squish original (10 bytes)   1.48M (674.47ns) (± 2.30%)  288B/op   4.74× slower
     squish new (10 bytes)   7.02M (142.38ns) (± 5.15%)  240B/op        fastest
```
We can optimize ascii-only strings when squishing by avoiding the use of a regular expression. Also, I am not certain the regex version handles the extended set of "space" characters in the intended way( e.g. nonbreaking space ).

Consider the behavior in Ruby:
```ruby
irb(main):007:0> " foo \u00A0$\u00A0  bar    \n   \t   boo --- \u200A".gsub(/[[:space:]]+/, ' ')
=> " foo $ bar boo --- "
```
This looks correct to me in terms of collapsing inner spaces, including non-ascii space sequences.

Now in Crystal 0.34:
```crystal
" foo \u00A0$\u00A0  bar    \n   \t   boo --- \u200A".gsub(/[[:space:]]+/, " ")
=> "foo  $  bar boo ---  "
```
It may be hard to visually see the difference, but there are two spaces between the "foo" and "$". I know that Crystal uses libpcre and Ruby uses Oniguruma so it's not apples to apples anyway.

That said, if the regex doesn't collapse the extended set of space characters we could consider just going with the fast path regardless of ascii-only vs non-ascii.

jamescook:
## Purpose
Resolves #1157 

## Description
Optimizes String#squish for both ascii and unicode text. There is some unavoidable (I think) duplication to optimize the ascii-only case. The change could be simplified to only use `Char#whitespace?` (which works for both ascii and unicode) but it's slower for ascii-only strings. However, it's still significantly faster than using a regular expression.

I replaced bench.cr with my benchmark per @jwoertink 's comment in the Github issue. It shows the difference using the ascii-only optimization (Char#ascii_whitespace? vs Char#whitespace?). Here are the results from my machine (2014 macbook):

```
     squish regex ascii-only whitespace (600 bytes)  42.17k ( 23.71µs) (± 1.57%)  5.92kB/op   8.97× slower
 squish optimized ascii-only whitespace (600 bytes) 378.22k (  2.64µs) (± 0.54%)  1.19kB/op        fastest
squish simplified ascii-only whitespace (600 bytes) 277.88k (  3.60µs) (± 1.24%)  1.19kB/op   1.36× slower

     squish regex w/ unicode whitespace (630 bytes)  63.52k ( 15.74µs) (± 0.95%)  4.48kB/op   3.39× slower
 squish optimized w/ unicode whitespace (630 bytes) 214.08k (  4.67µs) (± 1.42%)  1.41kB/op   1.01× slower
squish simplified w/ unicode whitespace (630 bytes) 215.32k (  4.64µs) (± 0.69%)  1.41kB/op        fastest
```
Lastly, there appears to be a unintended bugfix for consecutive unicode whitespace. The original regex doesn't appear to match them (libpcre issue?). The benchmark will show the difference if you run it locally.

## Checklist

### DIFF
--- a/bench.cr
+++ b/bench.cr
@@ -1,21 +1,86 @@
 require "benchmark"
 
-text = <<-TEXT
-Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua
-TEXT
-
-string_text = text * 40_000
-io_text = IO::Memory.new(string_text)
-
-Benchmark.ips do |x|
-  x.report("string") do
-    io = IO::Memory.new
-
-    io.print(io_text.to_s)
+class String
+  # Original
+  def squish_regex : String
+    gsub(/[[:space:]]+/, " ").strip
   end
-  x.report("io") do
-    io = IO::Memory.new
 
-    io.print(io_text)
+  def squish_new : String
+    if ascii_only?
+      squish_ascii
+    else
+      squish_unicode
+    end
   end
+
+  def squish_simplified : String
+    squish_unicode
+  end
+
+  private def squish_ascii : String
+    String.build(size) do |str|
+      print_blank = false
+      each_char do |chr|
+        if chr.ascii_whitespace?
+          if print_blank
+            str << ' '
+            print_blank = false
+          end
+        else
+          print_blank = true
+          str << chr
+        end
+      end
+    end.strip
+  end
+
+  private def squish_unicode : String
+    String.build(size) do |str|
+      print_blank = false
+      each_char do |chr|
+        if chr.whitespace?
+          if print_blank
+            str << ' '
+            print_blank = false
+          end
+        else
+          print_blank = true
+          str << chr
+        end
+      end
+    end.strip
+  end
+end
+
+puts "Sanity check the return output is consistent:"
+example = " f f\u00A0\u00A0\u00A0f f \n \t \v\v \f\f 11111  a l0* あ\u00A0\u00A0\u00A0 "
+puts "String to squish " + example.inspect
+puts "regex: " + example.squish_regex.inspect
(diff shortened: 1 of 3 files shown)


spec/charms/string_spec.cr
```diff
@@ -2,11 +2,10 @@ require "../spec_helper"
 
 describe "String charm" do
   describe "squish" do
-    it "squishes the text by removing newlines and extra whitespace" do
-      og_string = " foo   bar    \n   \t   boo"
```
jwoertink, line 6: Can we also leave this in? I dig the new spec, but I'm thinking leave both of them in just for the extra safety. What are your thoughts?

bench.cr
```diff
@@ -1,21 +1,86 @@
...
+          print_blank = true
+          str << chr
+        end
+      end
+    end.strip
   end
 end
+
+puts "Sanity check the return output is consistent:"
+example = " f f\u00A0\u00A0\u00A0f f \n \t \v\v \f\f 11111  a l0* あ\u00A0\u00A0\u00A0 "
```
jwoertink, line 57: 🚀

PR #1160, opened 2020-05-29
Fix bug where session cookie was not always written at the right time.

### LINKED ISSUE #1064: Set-Cookie is not arrived to a browser
Whenever I restarted a browser, Lucky failed to send a Set-Cookie header for a session store. I can see context.response has "Set-Cookie" header but it doesn't not appear on Chrome's inspector.
If I refresh again, it sets a cookie but conflicts with existing csrf token value.

wout:
## Purpose
This fixes #1064.

## Description
- `Lucky::SessionHandler` has been removed
- `Lucky::FlashHandler` has been removed
- flash messages, session cookie and cookies are now written in `Lucky::TextResponse#print`

I tested this on our app as well and everything works as expected now. 😌️

## Checklist


src/lucky/text_response.cr
```diff
@@ -42,4 +43,19 @@ class Lucky::TextResponse < Lucky::Response
         Lucky::Server.settings.gzip_content_types.includes?(content_type)
     {% end %}
   end
+
+  private def write_cookies : Void
+    context.cookies.set(
```
jwoertink, line 48: Can we still break this one out in to a `write_session` method?

src/lucky/renderable.cr
```diff
@@ -196,7 +198,7 @@ module Lucky::Renderable
     {% raise "'text' in actions has been renamed to 'plain_text'" %}
   end
 
-  @[Deprecated("`render_text deprecated. Use `plain_text` instead")]
+  @[Deprecated("`render_text` deprecated. Use `plain_text` instead")]
```
jwoertink, line 201: 👍

spec/lucky/text_response_spec.cr
```diff
@@ -67,6 +159,11 @@ private def print_response(context : HTTP::Server::Context, status : Int32?)
   print_response_with_body(context, "", status)
 end
 
-private def print_response_with_body(context : HTTP::Server::Context, body = "", status = 200, content_type = "text/html")
+private def print_response_with_body(
+  context : HTTP::Server::Context,
```
jwoertink, line 163: Did the formatter do this? I ask because the one below has the first arg start on the same line as the signature. I'm thinking maybe we make them the same (one way or the other). I was curious if the formatter allowed both, or if it prefers one vs the other way. Does that make sense?

src/lucky/text_response.cr
wout: To be honest, I have been doubting about keeping them separate as well. My feeling was we're "writing cookies" here so I merged the two methods. But I can split them out again if you prefer.

jwoertink: They do sort of go hand in hand, but my thought is they're still treated and accessed differently. Plus it makes it easier to quickly look at the code and see what is doing what.

spec/lucky/text_response_spec.cr
wout: I tend to pit the first argument on a new line if everything doe not fit on the same line. But the formatter does respect that and will put all following arguments on the same line.

I like it because that will consistently put all multi-line arguments on the same indentation throughout the file. Otherwise, they are hanging at different places, sometimes with odd numbered indentation levels, depending on the length of the name of the method. 

Here again, if you prefer the first argument at the same line as the method, I'll change that. ☺️

src/lucky/text_response.cr
wout: You're right, it makes for an easier red. I'll revert that back.

spec/lucky/text_response_spec.cr
wout: BTW, I have been moving a lot of code around there, because that's where I initially added the changes. But then I realised it should be somewhere else. That's why it has been changed.

src/lucky/text_response.cr
```diff
@@ -20,6 +20,8 @@ class Lucky::TextResponse < Lucky::Response
   end
 
   def print : Nil
+    write_session
+    write_cookies
```
paulcsmith, line 24: Now that cookies/session are set here, I think the flash handler should also be removed and the flash set before `write_session`. That way it gets written to the session cookie when it is printed.

spec/lucky/text_response_spec.cr
paulcsmith: I think this look good 👍

src/lucky/text_response.cr
wout: Great, I'll do that as well.

wout:
> Do you know why it wasn't working before, but it worked this time? Was it that it changed to the `TextResponse` class? Just trying to make sure I understand why this works but it didn't work when done in `Renderable`

Well, yeah, that's what threw me off. I didn't realise `Lucky::TextResponse` is initialized in two places:

- https://github.com/wout/lucky/blob/1064-write-cookies-right-before-text-response/src/lucky/renderable.cr#L83
- https://github.com/wout/lucky/blob/1064-write-cookies-right-before-text-response/src/lucky/renderable.cr#L185

So I added the `write_session` and `write_cookies` methods to the latter, not the former. It makes sense to move those methods to a more central place, especially since it's also used in other places:

https://github.com/wout/lucky/blob/1064-write-cookies-right-before-text-response/src/lucky/redirectable.cr#L62-L81

src/lucky/text_response.cr
wout: Voila, that's done!

paulcsmith:
Ohhh that makes total sense. And I definitely agree on adding it to `TextResponse` so all the responses get it. Have you tested this new update with your flash messages and can confirm those are still working as expected? If so I think this LGTM!

spec/lucky/text_response_spec.cr
jwoertink: I'm ok with doing this. Should we do the same on https://github.com/luckyframework/lucky/pull/1160/files#diff-a949f27bf5c58356d03de361bcf6cba2R182 as well?

wout:
I'll have to look closer at it tomorrow. Specs run green, but not sure about my apps.

paulcsmith:
Sounds good. The flash/session stuff can be a bit gnarly with edge cases so better safe than sorry 😂

Thanks for double checking and for fixing this. Can't wait to get this out

wout:
Yeah, flash messages are a breed of their own. I'll have to install it into an app that uses them. More tomorrow!

wout:
@paulcsmith I've tested flash messages throughout one app and they all work as expected, except for the following two scenarios.

1. **a flash message set in the `SignIns::Delete` action is lost**
But I recall noticing this before. On one of my apps I even removed it.

2. **flash messages are not maintained during multiple redirects**
I 'm not sure if this was working before either.

spec/lucky/text_response_spec.cr
wout: That would be cleaner indeed, so I added those changes as well.

paulcsmith:
I think this is ok to merge. Could you open issues for those 2 redirect issues? I think they are likely bugs from before that we need to fix.

Thanks for the thorough investigation, and for the fix!

PR #1168, opened 2020-06-03
Add redirect_back functionality

### LINKED ISSUE #1158: Add redirect_back
Similar to Rail's `redirect_back`. Should require a `fallback` argument in case there is no previous page.

Example:

```crystal
redirect_back(fallback: Home::Index)
redirect_back(fallback: "/string/path")
```

matthewmcgarvey:
## Purpose

Fixes #1158 

`redirect_back` allows an action to send the user back to where they made the request from. This is really useful in situations like submitting a lead form where the app might have it in several locations but the app doesn't need to take them anywhere in particular after they submit it. Rather than sending the user to a specific place after submitting the form, we can now send them back to where they originally submitted it.

This was largely taken from Rails. [source](https://github.com/rails/rails/blob/97c3a5605bacd62e5a01872dfe5e5976cd982115/actionpack/lib/action_controller/metal/redirecting.rb#L90)

### Sidenote

If you look at the Rails code, there is functionality around limiting Referer redirects in some situations using the `allow_other_hosts` named argument. If it is set to `false` then it will not redirect to a Referer header that has a different host than the request host. So, if the app is running at `example.com`, it will not redirect to a Referer header of `differenturl.com` when set to false. While this might be useful, I didn't want to copy it over just to copy it.

## Description

Usage looks like:
```crystal
class NewsletterSignupsCreate < BrowserAction
  post "/newsletter/signup" do
    # do the signup code
    redirect_back fallback: Home::Index
  end
end
```

It's important that the fallback is required because the `Referer` header is not guaranteed so we need to send the user somewhere.

## Checklist

### DIFF
--- a/src/lucky/redirectable.cr
+++ b/src/lucky/redirectable.cr
@@ -24,6 +24,46 @@
 # method that takes a `String`. However, it's recommended you pass a
 # `Lucky::Action` class if possible because it guarantees runtime safety.
 module Lucky::Redirectable
+  # Redirect back with a `Lucky::Action` fallback
+  #
+  # ```crystal
+  # redirect_back fallback: Users::Index
+  # ```
+  def redirect_back(fallback : Lucky::Action.class, status = 302)
+    redirect_back fallback.route, status
+  end
+
+  # Redirect back with a `Lucky::RouteHelper` fallback
+  #
+  # ```crystal
+  # redirect_back fallback: Users::Show.with(user.id)
+  # ```
+  def redirect_back(fallback : Lucky::RouteHelper, status = 302)
+    redirect_back fallback.path, status
+  end
+
+  # Redirects the browser to the page that issued the request (the referrer)
+  # if possible, otherwise redirects to the provided default fallback
+  # location.
+  #
+  # The referrer information is pulled from the 'Referer' header on
+  # the request. This is an optional header, and if the request
+  # is missing this header the *fallback* will be used.
+  #
+  # ```crystal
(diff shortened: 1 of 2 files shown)


matthewmcgarvey:
@jwoertink Good catch, there are just sooo many overloads!

jwoertink:
😂  yup, but this is awesome! I saw what you mean with the `allow_other_hosts`. I think it's a little wonky. For now let's just roll without that, and we can think about a nice way to handle something like that later if people want it.

src/lucky/redirectable.cr
```diff
@@ -24,6 +24,55 @@
...
 # `Lucky::Action` class if possible because it guarantees runtime safety.
 module Lucky::Redirectable
+  # Redirect back with a `Lucky::Action` fallback
+  #
+  # ```crystal
+  # redirect_back fallback: Users::Index
+  # ```
+  def redirect_back(fallback : Lucky::Action.class, status = 302)
+    redirect_back fallback.route, status
+  end
```
paulcsmith, line 34:
```suggestion
  def redirect_back(*, fallback : Lucky::Action.class, status = 302)
    redirect_back fallback.route, status
  end
```

This forces the user to use a name arg. I think this is good since the args are not clear if the end-user leaves off the key.

This would need to be added in the other versions as well.

jwoertink: oh, that's sneaky.... I like it!

paulcsmith:
Looks great! Thank you 🎉

PR #1179, opened 2020-06-09
Upgrade to Crystal 0.35.0

bcardiff:
These are some of the changes I've notice to upgrade to Crystal 0.35.

Probably the local data of log entries can be redesigned to avoid intermediate structures. For now I was focused on make the spec pass.

The rest of the changes are self explanatory. Feel free to close the PR and use it as input for a cleaner migration.

dexter, avram, lucky_cli will need to be updated also.

The crystal property in shards.yml should be bumped to crystal: 0.35.0


paulcsmith:
Thank you very much! Closing this in favor of #1187 which is based on this PR 👍 🥳

PR #1188, opened 2020-06-16
[DO NOTE MERGE] v0.22.0

paulcsmith:
This is here for review, CI, and for future reference. But do not merge it. It will be tagged and then closed.


paulcsmith:
Not going to worry about ameba/formatting right now. Just need to get this out to fix Crystal 0.35.0. We're following up with a big release after this

Issue #1205, opened 2020-06-22
Make CONTRIBUTING.md, README.md, and PR Template consistent


stephendolan:
It was a bit confusing to figure out what's necessary for contributing to the progress (at least, for a first-time contributor).

First, I checked CONTRIBUTING.md, which has:
```md
## Submitting a Pull Request
1. [Fork][fork] the [official repository][repo].
2. [Create a topic branch.][branch]
3. Implement your feature or bug fix.
4. Add, commit, and push your changes.
5. [Submit a pull request.][pr]
```

But, upon creating a pull request, I noticed these tasks:
```md
## Checklist
* [ ] - An issue already exists detailing the issue/or feature request that this PR fixes
* [ ] - All specs are formatted with `crystal tool format spec src`
* [ ] - Inline documentation has been added and/or updated
* [ ] - Lucky builds on docker with `./script/setup`
* [ ] - All builds and specs pass on docker with `./script/test`
```

This was the first mention of Docker I'd encountered, so I went digging through the README and found:
```md
## Contributing

1. Fork it ( https://github.com/luckyframework/web/fork )
1. Create your feature branch (git checkout -b my-new-feature)
1. Install docker and docker-compose: https://docs.docker.com/compose/install/
1. Run `script/setup` to build the Docker containers with everything you need.
1. Make your changes
1. Make sure specs pass: `script/test`.
1. Add a note to the CHANGELOG
1. Commit your changes (git commit -am 'Add some feature')
1. Push to the branch (git push origin my-new-feature)
1. Create a new Pull Request

> Run specific tests with `script/test <path_to_spec>`

### Lucky to have you!

We love all of the community members that have put in hard work to make Lucky better.
If you're one of those people, we want to give you a t-shirt!

To get a shirt, we ask that you have made a significant contribution to Lucky.
This includes things like submitting PRs with bug fixes and feature implementations, helping other members
work through problems, and deploying real world applications using Lucky!

To claim your shirt, [fill in this form](https://forms.gle/w3PJ4pww8WDAuJov5).
```

I'd recommend:
- Moving the README "Contributing" content to CONTRIBUTING
- Linking to CONTRIBUTING for the README "Contributing" content
- Moving the PR Template checklist to CONTRIBUTING
- Simplifying the PR template checklist to one "I have adhered to all rules in CONTRIBUTING.md" or something

[...]

PR #1206, opened 2020-06-22
Add Lucky.root redirection to Dir.current through exception.

### LINKED ISSUE #1198: Create a Lucky.root method to allow intuitive access to project folder
It **looks** like `Dir.current` in a Lucky app will always return the root folder of the project, but I'm not confident that that's always true or sure about why.

What I'm after is something similar to `Rails.root` here:
https://github.com/rails/rails/blob/master/railties/lib/rails.rb#L63

It looks like we currently use `Dir.current` within the framework for the project path already:
https://github.com/luckyframework/lucky/blob/master/src/lucky/base_component.cr#L9

My naive suggestion would be to add this:
```cr
module Lucky
  def self.root
    Dir.current
  end
end
```

I'm new to Crystal from Ruby and would greatly appreciate input from someone with more experience to guide me through a pull request if it's agreed that this would be nice to have!

stephendolan:
## Purpose

Adds `Lucky.root` method which, when used in an application will raise a compile-time error redirecting users to leverage `Dir.current`. **Resolves #1198**.

## Description

Context can be found in #1198, including around the decision to make this a compile-time error.

No specs added since it's raised at compile-time, and not post-build.

_Not 100% sure about how this will look in code documentation with just the single comment, please let me know if I need to improve that._

_Didn't add a CHANGELOG entry since I didn't see an existing "UNRELEASED" section, but happy to help add that in._

## Checklist

### DIFF
--- a/src/lucky.cr
+++ b/src/lucky.cr
@@ -23,4 +23,9 @@ require "./lucky/paginator/components/*"
 module Lucky
   Log              = ::Log.for("lucky")
   ContinuedPipeLog = Log.for("continued_pipe_log")
+
+  # Use Dir.current to return the root folder of your Lucky application.
+  def self.root
+    {% raise "Please use Crystal's 'Dir.current' to return the root folder of your Lucky application." %}
+  end
 end
--- a/src/lucky.cr
+++ b/src/lucky.cr
@@ -25,6 +25,10 @@ module Lucky
   ContinuedPipeLog = Log.for("continued_pipe_log")
 
   # Use Dir.current to return the root folder of your Lucky application.
+  #
+  # In some frameworks there is a method called `root` that returns the root directory of the project.
+  # In Crystal there is a built-in method for this: `Dir.current`. This method exists purely to help new users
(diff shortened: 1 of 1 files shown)


src/lucky.cr
```diff
@@ -23,4 +23,9 @@ require "./lucky/paginator/components/*"
 module Lucky
   Log              = ::Log.for("lucky")
   ContinuedPipeLog = Log.for("continued_pipe_log")
+
+  # Use Dir.current to return the root folder of your Lucky application.
+  def self.root
```
paulcsmith, line 28:
```suggestion
  # Use Dir.current to return the root folder of your Lucky application.
  #
  # In some frameworks there is a method called `root` that returns the root directory of the project.
  # In Crystal there is a built-in method for this: `Dir.current`. This method exists purely to help new users
  # find `Dir.current`. If you call `Lucky.root` it will raise a compile-time error directing you to use `Dir.current`
  def self.root
```

stephendolan:
Cool, sounds good. @jwoertink / @paulcsmith I'll make a note in #1205 to discuss removing the CHANGELOG entry requirement.

paulcsmith:
This looks great. Thanks @stephendolan!

PR #1220, opened 2020-07-17
Added Renderable#data method to be able to respond with IO and String from an Action

### LINKED ISSUE #1219: Implement Lucky::Renderable#data
Currently the only way to respond with a file from a Lucky::Action is by using `Lucky::Renderable#file` https://github.com/luckyframework/lucky/blob/master/src/lucky/renderable.cr#L192.

However `#file` accepts a path to the file as a parameter. I think it would be great to have a '#data' method that will have the following signatures:

``` crystal
def data(
  data : String,
  content_type : String = "application/octet-stream",
  disposition : String = "attachment",
  filename : String? = nil,
  status : Int32? = nil
) : Lucky::DataResponse
```

``` crystal
def data(
  data : IO,
  content_type : String = "application/octet-stream",
  disposition : String = "attachment",
  filename : String? = nil,
  status : Int32? = nil
) : Lucky::DataResponse
```

Long story short, it should behave like `#send_data` (https://api.rubyonrails.org/v5.2.3/classes/ActionController/DataStreaming.html#method-i-send_data) in Rails (it is ok to mention Rails?:) ).

igor-alexandrov:
This PR closes #1219.

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -209,6 +209,26 @@ module Lucky::Renderable
     file(path, content_type, disposition, filename, status.value)
   end
 
+  def data(
+    data : String,
+    content_type : String = "application/octet-stream",
+    disposition : String = "attachment",
+    filename : String? = nil,
+    status : Int32? = nil
+  ) : Lucky::DataResponse
+    data(IO::Memory.new(data), content_type, disposition, filename, status)
+  end
+
+  def data(
+    io : IO,
+    content_type : String = "application/octet-stream",
+    disposition : String = "attachment",
+    filename : String? = nil,
+    status : Int32? = nil
+  ) : Lucky::DataResponse
+    Lucky::DataResponse.new(context, io.gets_to_end, content_type, disposition, filename, status)
+  end
+
   def send_text_response(
     body : String,
     content_type : String,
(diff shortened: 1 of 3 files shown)


src/lucky/renderable.cr
```diff
@@ -209,6 +209,26 @@ module Lucky::Renderable
...
+  def data(
+    io : IO,
+    content_type : String = "application/octet-stream",
+    disposition : String = "attachment",
+    filename : String? = nil,
+    status : Int32? = nil
+  ) : Lucky::DataResponse
+    Lucky::DataResponse.new(context, io.gets_to_end, content_type, disposition, filename, status)
+  end
+
```
jwoertink, line 231: Maybe we should also add an overload that takes `HTTP::Status` like the others? Just to keep those consistent.

stephendolan:
@igor-alexandrov If you want to throw a couple code snippets of using this functionality into this PR, I'm happy to take those over to the website and help document the new feature if you haven't done so already.

igor-alexandrov:
@stephendolan the most obvious example looks like:

``` crystal
class Reports::MyReport < ApiAction
  get "/reports/my_report" do
    result = CSV.build do |csv|
      csv.row "one", "two"
      csv.row "three"
    end

    data result, filename: "my_report.csv"
  end
end
``` 

Now I have doubts, do we really need `data` method that accepts `IO` or `data` with `String` first argument will be enough?
@jwoertink @paulcsmith what do you think?

igor-alexandrov:
@paulcsmith done, I also added specs for `"Content-Length"` header.

Issue #1221, opened 2020-07-21
Rename `m` to `mount`, but keep signature


paulcsmith:
https://github.com/luckyframework/lucky/issues/1049#issuecomment-653904514
https://github.com/luckyframework/lucky/issues/1215#issuecomment-661930262

stephendolan:
@paulcsmith / @jwoertink Were one of you already working on this in a branch, or could I take a swing at it if I find some time this weekend?

jwoertink:
I'm not. Feel free to have at it!

paulcsmith:
@stephendolan that'd be great! Thanks!

PR #1224, opened 2020-07-22
Replace references/generators for deprecated `mount` calls

stephendolan:
## Purpose
Update deprecated `mount` generator code to generate code that doesn't immediately get flagged with deprecation warnings!

## Description
We recently deprecated `mount`, but the `lucky gen.resource.browser` CLI command still generates resources using it. This replaces all references to `mount` in the framework code with the new `m` call, aside from in the actual docblocks of the deprecated methods.

## Checklist

### DIFF
--- a/src/lucky/paginator/backend_helpers.cr
+++ b/src/lucky/paginator/backend_helpers.cr
@@ -22,7 +22,7 @@ module Lucky::Paginator::BackendHelpers
   #
   #   def content
   #     # Render 'users' like normal
-  #     mount Lucky::Paginator::SimpleNav.new(@pages)
+  #     m Lucky::Paginator::SimpleNav, @pages
   #   end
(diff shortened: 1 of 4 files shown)


jwoertink:
Actually, this will go back https://github.com/luckyframework/lucky/issues/1221 The signature will stay the same, but we're gonna go back to calling it `mount`. 😅 I guess maybe update it with that in mind, and we can hold off merging until this is in? Or we can just close for now and reopen later? Either way.

paulcsmith:
Oops, just saw your comment @jwoertink. I think we can still merge this change and change it to `mount` later

stephendolan:
I'm actually really glad to see that we're reverting! I also favor the verbose nature of `mount`.

At least now everything is consistent before we revert!

Issue #1049, opened 2020-03-14
Shortcut for mounting


paulcsmith:
I mount a **lot** of components. I wonder if we shuold add some kind of shortcut method for `mount`?

```crystal
mount MyButton.new
m MyButton.new
```

Thoughts?

jwoertink:
same! Actually, this shortcut is what I'm using because I needed `context` and `current_site` in ALL of my components

wout:
Contrary to #1048, I think this can be very useful.

jwoertink:
I actually use this code in my app. Since we have a multitenant app, we need to know the `current_site` on just about every component. We also need context in some cases, so by having this, we can set defaults. 

```crystal
module Mountable
  # mounts `component` with `current_site` and `context`
  # m(Shared::Footer, current_installation: @current_installation)
  def m(component, **named_args)
    mount component.new(**{current_site: @current_site, context: @context}.merge(named_args))
  end
end
```

paulcsmith:
Just wanted to document an idea I had that I think will make this all much easier. You'd be able to use some kind of macro on your BaseComponent that sets up the "implicit defaults" that it should get without the user needing to pass them in. 

Here is a gist of the rough POC: https://play.crystal-lang.org/#/r/8v17
<details>
 <summary>Code sample</summary>
```crystal
class Component
  def initialize(@age : Int32, name : String)
  end
  
  def render
    p! @age
    p! @name
  end
  
  macro implicit_defaults
    {age: age}
  end
end
 
def age
  20
end
```
</details>
 

def mount(component : Component.class, **named_args)
  named_args = named_args.merge(Component.implicit_defaults)
  component.new(**named_args)
end
```

So what this would look like in Lucky:

```crystal
class BaseComponent < Lucky::Component
  needs current_user : User
  needs context : HTTP::Server::Context
  implicit_defaults :current_user, :context
end

class SettingsHeader < BaseComponent
  needs something_else : String

  def render
    h1 current_user.name
    para something_else
  end
end

# In your page or component
mount(SettingsHeader, something_else: "foo")
```

Still need to work out how to pass blocks to this and make sure the syntax all looks good, but this would be *super* cool I think because the component defines its implicit defaults and anything rendering them will try to pass those and will fail at compile time if it is not there. Also means testing is easy since you would pass the needs to the component through the initializer like always.

Thoughts?

jwoertink:
I like it!

paulcsmith:
[...]

PR #1226, opened 2020-07-26
Rename `m` to `mount`, but keep signature

### LINKED ISSUE #1221: Rename `m` to `mount`, but keep signature

stephendolan:
## Background and context
- Initial implementation was named `mount`, and had a signature of: `mount MyComponent.new(name: "Test")`
- Rename of `mount` -> `m` occurred, which also changed the method signature to: `m MyComponent, name: "Test"`
- #1221 opened after several issue discussions about which direction to take, deciding to revert the rename but maintain the new signature.

## Purpose
This PR renames `m` -> `mount`, retains the method signature as-is, and removes the old signature completely.

Closes #1221 

## Description
I believe I caught everything necessary. I used the regular expressions `\bm[\s(]+[A-Z]` to locate the items that needed to be changed. The only things I'd call out that need explicit review:
- I wasn't sure how to handle the old signature. Do we want deprecation of `m`, `mount` with the old signature, or do we just revert to `mount` with the new signature to get any confusion sorted during the next version bump?
- I wasn't sure whether updating the `UPGRADE.md` file was allowed, but it seemed prudent to let anyone migrating from 0.22 -> 0.23 know that they don't *really* need to rename `mount` -> `m`.

## Checklist

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -27,7 +27,7 @@ brew upgrade lucky
 ### General updates
 
 - Update: `params.get` now strips white space. If you need the raw value, use `params.get_raw`.
-- Rename: `mount` to `m` in all pages that use components.
+- Rename: `mount` to `m` in all pages that use components. **Note: This was reverted in the next version**
 - Update: all mounted components to use new signature `mount(MyComponent.new(x: 1, y: 2))` -> `m(MyComponent, x: 1, y:2)`.
 - Remove: `Lucky::SessionHandler` and `Lucky::FlashHandler` from `src/app_server.cr`
 
--- a/src/lucky/mount_component.cr
+++ b/src/lucky/mount_component.cr
@@ -6,46 +6,10 @@ module Lucky::MountComponent
   # starts and ends.
   #
   # ```
-  # mount MyComponent.new
-  # ```
-  @[Deprecated("Use `#m` instead. Example: m(MyComponent, arg1: 123)")]
-  def mount(component : Lucky::BaseComponent) : Nil
-    print_component_comment(component.class) do
-      component.view(view).render
-    end
-  end
-
-  # Appends the `component` to the view. Takes a block, and yields the
-  # args passed to the component.
-  #
-  # When `Lucky::HTMLPage.settings.render_component_comments` is
-  # set to `true`, it will render HTML comments showing where the component
-  # starts and ends.
-  #
-  # ```
-  # mount MyComponent.new("jane") do |name|
-  #   text name.upcase
-  # end
-  # ```
-  @[Deprecated("Use `#m` instead. Example: m(MyComponent, arg1: 123) do/end")]
-  def mount(component : Lucky::BaseComponent) : Nil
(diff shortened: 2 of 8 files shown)


UPGRADE_NOTES.md
```diff
@@ -27,8 +27,7 @@ brew upgrade lucky
 ### General updates
 
 - Update: `params.get` now strips white space. If you need the raw value, use `params.get_raw`.
-- Rename: `mount` to `m` in all pages that use components.
-- Update: all mounted components to use new signature `mount(MyComponent.new(x: 1, y: 2))` -> `m(MyComponent, x: 1, y:2)`.
```
paulcsmith, line 31: I think we should keep this all as-is since the signature still needs to change and so that people can upgrade one version at a time if needed

src/lucky/mount_component.cr
```diff
@@ -6,47 +6,10 @@ module Lucky::MountComponent
...
-  # set to `true`, it will render HTML comments showing where the component
-  # starts and ends.
-  #
-  # ```
-  # mount MyComponent.new("jane") do |name|
-  #   text name.upcase
-  # end
-  # ```
-  @[Deprecated("Use `#m` instead. Example: m(MyComponent, arg1: 123) do/end")]
-  def mount(component : Lucky::BaseComponent) : Nil
```
paulcsmith, line 31: Can you add a deprecated method for `m` that tells people to use `mount` instead?

wontruefree:
should we keep the `m` alias?

stephendolan:
@wontruefree ,

There have been a couple issue/PR comment threads that I've read through. I was going by the comment @paulcsmith left here: https://github.com/luckyframework/lucky/issues/1049#issuecomment-661930675

I think the idea is that we don't want multiple ways to mount components since `m` is hard to search for, and may not intuitively be linked to `mount`. The option I read for handling a shorter `mount` call would be to implement an alias in your application.

UPGRADE_NOTES.md
stephendolan: @paulcsmith Yeah, I was thinking that removing it would be okay since nothing actually _breaks_ and you just get deprecation warnings. Totally see the value in having each upgrade result in a clean, no-deprecation-warning app though. In this case if they go one by one it'll just mean a rename from `mount` -> `m` -> `mount`, but having done that myself I don't think it's too bad.

Are you good with me keeping the bolded note so that folks know it was later reverted?

src/lucky/mount_component.cr
stephendolan: @paulcsmith For sure! I'll submit that later this week.

stephendolan:
@paulcsmith Added back the `UPGRADE_NOTES` that I removed, but left the note that we reverted it. If you don't mind running through `src/lucky/mount_component.cr` one last time to make sure I added those deprecations correctly, I'd appreciate it!

Tests are all passing, though, so this should be good to merge!

PR #1227, opened 2020-07-26
Update Lucky gen.model to accept nilable types

### LINKED ISSUE #1217: Lucky gen.model doesn't accept nilable types `?`
`lucky gen.model name:String?` fails because of the `?`. It should let you use that though

stephendolan:
## Purpose
Closes #1217 
Allows users to generate models and browser resources with optional (nilable) columns like `lucky gen.model User name:String notes:String?`.

## Description
- Updated gen.model test to capture generating a model with a nilable column.
- Updated gen.resource.browser to capture generating a resource with a nilable column.
- Updated the migration with columns mixin to strip `?` from types before comparing to allowed list.

## Checklist

### DIFF
--- a/spec/tasks/gen/model_spec.cr
+++ b/spec/tasks/gen/model_spec.cr
@@ -31,7 +31,7 @@ describe Gen::Model do
       Gen::Migration.silence_output do
         io = IO::Memory.new
         model_name = "ContactInfo"
-        ARGV.push(model_name, "name:String", "contacted_at:Time")
+        ARGV.push(model_name, "name:String", "notes:String?", "contacted_at:Time")
 
         Gen::Model.new.call(io)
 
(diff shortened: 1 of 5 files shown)


spec/tasks/gen/model_spec.cr
```diff
@@ -50,14 +50,43 @@ describe Gen::Model do
     end
   end
 
+  it "generates a model with nilable columns" do
```
paulcsmith, line 53: Rather than adding a new block, what do you think about passing a nilable column in the spec above? That way there aren't so many shared setup and expectation lines

stephendolan: Can do, @paulcsmith ! I'll resubmit this week

stephendolan:
@paulcsmith This should be good once the pipelines complete!

PR #1229, opened 2020-07-29
Adding the optional routing in the to core

jwoertink:
## Purpose
Optional path params were added in to the router https://github.com/luckyframework/lucky_router/pull/18 but there was never a test done to see if Lucky core needed any updates. I added a spec to see if we needed to do anything, and it doesn't look like it.

## Checklist

### DIFF
--- a/spec/lucky/action_route_params_spec.cr
+++ b/spec/lucky/action_route_params_spec.cr
@@ -2,18 +2,33 @@ require "../spec_helper"
 
 include ContextHelper
 
-private class TestAction < TestAction
+private class TestParamAction < TestAction
   get "/test/:param_1/:param_2" do
+    plain_text "test"
+  end
+end
+
+private class TestOptionalParamAction < TestAction
+  get "/complex_posts/:required/?:optional_1/?:optional_2" do
     plain_text "test"
   end
 end
 
 describe "Automatically generated param helpers" do
   it "generates helpers for all route params" do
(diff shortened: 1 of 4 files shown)


jwoertink:
Ah, gotcha. Ok, I'll take a look at that!

paulcsmith:
Nice! I believe this is the spot: https://github.com/luckyframework/lucky/blob/72336b7d8bb96d90f1ca793d833435e9bd3a28e0/src/lucky/routable.cr#L198-L203. You'd want to check for `starts_with?("?:")` and use `params.get?()`. I think it'll also need to update the `with` and `route` helpers so that the params can be `nil` and they have a default value of `nil` which is somewhere in here I think :P https://github.com/luckyframework/lucky/blob/72336b7d8bb96d90f1ca793d833435e9bd3a28e0/src/lucky/routable.cr#L228-L243

jwoertink:
😂  ok, so now that I just merged in that Avram update, this is failing. Should I include those Avram updates in this? Or make a separate PR to update that, then get that PR merged in to here? 🤔

paulcsmith:
@jwoertink I think a separate PR with Avram fixes would be good since it’s small and we can get that in super fast

jwoertink:
@paulcsmith Alright, I sent in that other PR and merged it in. It was a pretty small change. Let me know if you see any cases missing on this one.

PR #1234, opened 2020-08-08
Clarify accepted_format doc blocks

stephendolan:
## Purpose
Revise the doc block for accepted_format to (hopefully) make it more clear.

## Description
This was a result of research that was done for https://github.com/luckyframework/website/pull/392

## Checklist

### DIFF
--- a/src/lucky/verify_accepts_format.cr
+++ b/src/lucky/verify_accepts_format.cr
@@ -10,24 +10,24 @@ module Lucky::VerifyAcceptsFormat
     end
   end
 
-  # Set what formats the Action accepts.
+  # Set the single format that the Action accepts.
   #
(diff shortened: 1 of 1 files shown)


src/lucky/verify_accepts_format.cr
```diff
@@ -10,24 +10,24 @@ module Lucky::VerifyAcceptsFormat
     end
   end
 
-  # Set what formats the Action accepts.
+  # Set the single format that the Action accepts.
```
paulcsmith, line 13: Actually I'm not sure this is right. it can set multiple formats so this part can probably be reverted

paulcsmith: Never mind. Facepalm

PR #1240, opened 2020-08-20
[DO NOT MERGE] BaseComponents do not take in arbitrary initialize arguments

### LINKED ISSUE #805: BaseComponent should fail at compile time if passed an unused argument
In pages this is nice because you might have a global `expose` but not all pages need the value. So the page will just ignore it.

But for components we should raise if we pass unused arguments. https://gitter.im/luckyframework/Lobby?at=5d03d50af5cf1f10bc709053

matthewmcgarvey:
## Purpose

Fixes #805 

## Description

Components will now fail to compile if passed arguments that it does not expect to have. This is different than pages which can still be passed anything and will ignore it.

## Checklist

### DIFF
--- a/src/lucky/html_builder.cr
+++ b/src/lucky/html_builder.cr
@@ -64,7 +64,9 @@ module Lucky::HTMLBuilder
           {% has_default = value || value == false || value == nil %}
           @{{ var.id }} : {{ type }}{% if has_default %} = {{ value }}{% end %},
         {% end %}
-        **unused_exposures
+        {% unless @type < Lucky::BaseComponent %}
+          **unused_exposures
+        {% end %}
         )
       end
     {% end %}


src/lucky/html_builder.cr
```diff
@@ -64,7 +64,9 @@ module Lucky::HTMLBuilder
           {% has_default = value || value == false || value == nil %}
           @{{ var.id }} : {{ type }}{% if has_default %} = {{ value }}{% end %},
         {% end %}
-        **unused_exposures
+        {% unless @type < Lucky::BaseComponent %}
```
matthewmcgarvey, line 67: Only accept arbitrary arguments if the class does not inherit from BaseComponent

matthewmcgarvey:
There's no reason to keep this PR open. It can be brought back if wanted

PR #1249, opened 2020-09-01
Update to latest ameba

jwoertink:
ignore all that other stuff....

Let's just update ameba 😬


jwoertink:
Ok, apparently I'm wrong in my assumption, but I'm having an issue where a lucky app isn't pulling the latest wordsmith, so something else must be wrong. 🤔

PR #1258, opened 2020-09-10
Lucky::Params refactor

### LINKED ISSUE #1075: Memoizing JsonBodyParser
In https://github.com/luckyframework/lucky/pull/1061 there was a change made where [memoize was added here](https://github.com/luckyframework/lucky/pull/1061/files#diff-d633a5eeaf04a8d42861bcc3614b84bfR480). I think we can also add it [here](https://github.com/luckyframework/lucky/blob/master/src/lucky/json_body_parser.cr#L18) to just match those up. Probably a super small performance boost since `body` would be called twice if there's json to be parsed.

### LINKED ISSUE #1257: JsonBodyParser should use body string from Params class
`Lucky::JsonBodyParser` current has its own [body method](https://github.com/luckyframework/lucky/blob/face632c8adcb700a4fc2eeff13931eb4547b380/src/lucky/json_body_parser.cr#L18-L20). It is not memoized so that method is run twice. That is twice more than is needed. If it took in the body it could utilize the memoized body method found in `Lucky::Params`.

There are also several methods in the Params class that are using the old form of memoization.

matthewmcgarvey:
## Purpose

Fixes #1257 
Fixes #1075 

There were also some easy refactorings made in the `Lucky::Params` class.

1. added the `memoize` macro where it was missing
2. some of the JSON::Any interactions were redundant
3. removed unused field from ParamParsingError that is used by JsonBodyParser
4. Switched gsub usage to lchop in places where it was removing the start of strings

## Description

## Checklist

### DIFF
--- a/src/lucky/errors.cr
+++ b/src/lucky/errors.cr
@@ -17,7 +17,9 @@ module Lucky
   class ParamParsingError < Error
     include Lucky::RenderableError
 
-    def initialize
+    getter request
+
+    def initialize(@request : HTTP::Request)
       super "Failed to parse the request parameters."
     end
 
--- a/src/lucky/errors.cr
+++ b/src/lucky/errors.cr
@@ -17,9 +17,7 @@ module Lucky
   class ParamParsingError < Error
     include Lucky::RenderableError
 
-    getter request
-
-    def initialize(@request : HTTP::Request)
+    def initialize
       super "Failed to parse the request parameters."
     end
 
--- a/spec/lucky/errors_spec.cr
+++ b/spec/lucky/errors_spec.cr
@@ -60,7 +60,7 @@ describe "Errors" do
 
   describe Lucky::ParamParsingError do
     it "is renderable" do
-      error = Lucky::ParamParsingError.new(request: build_request)
+      error = Lucky::ParamParsingError.new
       error.should be_a(Lucky::RenderableError)
       error.renderable_message.should contain("There was a problem parsing the JSON")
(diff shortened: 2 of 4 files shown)


wontruefree:
This looks good.  Do you have any benchmarks?  I am just curious if this increases performance at all.

matthewmcgarvey:
It's quite a pain to get both versions of the code at the same time so I just made a benchmark and ran it on this branch and master. Here's the output of both:
`old  30.03k ( 33.30µs) (± 7.85%)  26.4kB/op`
`new  35.00k ( 28.57µs) (± 1.51%)  24.8kB/op`

rounded, this branch 1.17 times faster than master

wontruefree:
@matthewmcgarvey Thanks for doing the work to check this!

src/lucky/errors.cr
```diff
@@ -17,9 +17,7 @@ module Lucky
   class ParamParsingError < Error
     include Lucky::RenderableError
 
-    getter request
-
-    def initialize(@request : HTTP::Request)
```
paulcsmith, line 22: Could this be added? The reason for it is that it can be used in `Errors::Show` or in future error pages to show the request body/url that could not be parsed

paulcsmith (commented):
Other than the one thought about keeping the request available in the `ParamParsingError` I think this looks good!

jwoertink (approved):
Looks good to me! Feel free to merge when you think it's ready @matthewmcgarvey

PR #1270, opened 2020-09-26
Moved the Memoizable module to Object. Fixes #1253

### LINKED ISSUE #1095: Move memoize  macro to Object
It’s helpful in operations, models, etc so it’d be good to move it to Object

### LINKED ISSUE #1253: include Lucky::Memoizable in Object so it usable everywhere
https://github.com/luckyframework/lucky/blob/master/src/lucky/memoizable.cr is only included in actions but we should add it o an `object_extensions` charm instead https://github.com/luckyframework/lucky/blob/master/src/charms/object.cr

jwoertink:
## Purpose
Fixes #1253 
Fixes #1095 

## Description
By including this module in to Object directly, we no longer need to include it in each class we want to use. This will allow us to use it in component, pages, actions, and more out of the box.

## Checklist


src/charms/object.cr
```diff
@@ -1,3 +1,4 @@
 class Object
   include ::Lucky::QuickDef
+  include Lucky::Memoizable
```
paulcsmith, line 3: Probably not necessary, but just to keep it consistent/safe this might be good

```suggestion
  include ::Lucky::Memoizable
```

paulcsmith (approved):
Looks like there is a formatting issue but overall this looks good!

PR #1287, opened 2020-10-23
Adding specs for the has_key_for? method

jwoertink:
## Purpose
This is to accompany https://github.com/luckyframework/avram/pull/500

## Description
This adds in specs for the new `params.has_key_for?` method. Though the method is defined in Avram, we can't accurately write specs for it in Avram since a user's `params` will actually be `Lucky::Params`. In this case, it's closer to "real world" tests.

## Checklist

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -40,7 +40,8 @@ dependencies:
     version: ~> 0.2.0
   avram:
     github: luckyframework/avram
-    version: ~> 0.17.0
+    branch: master
+    #version: ~> 0.17.0
   lucky_router:
     github: luckyframework/lucky_router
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (approved):

paulcsmith:
IMO I think we should remove this from Lucky and keep it in Avram. Right now we could change `has_key_for?` to anything we want in Avram and we wouldn't know if it works or is broken until we pull master into Lucky and run the specs here. I think since the logic is in Avram the spec should be there. I left an example here: https://github.com/luckyframework/avram/pull/500#discussion_r512787102 of how it could be done

PR #1303, opened 2020-11-09
Imporve needs error reporting

shortly-portly:
## Purpose
Returns a more helpful error message if you define a "needs" variable more than once.

## Description
Assignable.cr now checks against the list of ASSIGNS to ensure the same var isn't defined twice.

### Sample Error Message:

Error:

Duplicate needs definition: 'title : String' defined in "/data/spec/lucky/assignable_spec.cr":25:9
This needs is already defined as 'title : String' in "/data/spec/lucky/assignable_spec.cr":24:9

Please note:
* Could not define a test for this fix as the "test" for the fix would mean the code doesn't compile
* All existings tests have passed.
* This is my first ever submit so apologies if I've missed anything out.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Changes
+- Fixed 'Raise helpful error if the same needs is not unique[#531](https://github.com/luckyframework/lucky/issues/531)
+
 ### Changes in 0.24
 
 - Fixed: `send_text_response` default status to nil [#1214](https://github.com/luckyframework/lucky/pull/1214)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,3 @@
-### Changes
-- Fixed 'Raise helpful error if the same needs is not unique[#531](https://github.com/luckyframework/lucky/issues/531)
-
 ### Changes in 0.24
 
 - Fixed: `send_text_response` default status to nil [#1214](https://github.com/luckyframework/lucky/pull/1214)
--- a/src/lucky/assignable.cr
+++ b/src/lucky/assignable.cr
@@ -30,8 +30,17 @@ module Lucky::Assignable
         {% raise "Using '?' in a 'needs' var name is no longer supported. Now Lucky generates a method ending in '?' if the type is 'Bool'." %}
       {% end %}
 
-      {% if ASSIGNS.any? { |d| d.var == declaration.var } %}
-        {% raise "Found duplicate needs definition '#{declaration}'" %}
+      # Ensure that the needs variable name has not been previously defined.
+      {% previous_declaration = ASSIGNS.find { |d| d.var == declaration.var } %}
+      {% if previous_declaration %}
+        
+        {% raise <<-ERROR
+          \n
+          Duplicate needs definition: '#{declaration}' defined in #{declaration.filename}:#{declaration.line_number}:#{declaration.column_number}
+          This needs is already defined as '#{previous_declaration}' in #{previous_declaration.filename}:#{previous_declaration.line_number}:#{previous_declaration.column_number}
+          ERROR
+        %}
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+### Changes
+- Fixed 'Raise helpful error if the same needs is not unique[#531](https://github.com/luckyframework/lucky/issues/531)
```
jwoertink, line 2: Don't worry about adding in changelog entries. These get added in all at the same time.

src/lucky/assignable.cr
```diff
@@ -40,6 +53,7 @@ module Lucky::Assignable
         end
       {% end %}
 
+
```
jwoertink, line 56: Looks like the formatter failed. This extra line probably has to go.

src/lucky/assignable.cr
```diff
@@ -30,6 +30,19 @@ module Lucky::Assignable
         {% raise "Using '?' in a 'needs' var name is no longer supported. Now Lucky generates a method ending in '?' if the type is 'Bool'." %}
       {% end %}
 
+      # Ensure that the needs variable name has not been previously defined.
+      {% previous_declaration = ASSIGNS.find { |d| d.var == declaration.var } %}
+      {% if previous_declaration %}
+        
+        {% raise <<-ERROR
+          \n
+          Duplicate needs definition: '#{declaration}' defined in #{declaration.filename}:#{declaration.line_number}:#{declaration.column_number}
```
jwoertink, line 39: wait a minute.... `declaration.filename` is a thing?! 😱  That's amazing!

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
...
+
```
matthewmcgarvey, line 3: Could you remove this, we update this file ourselves before a release

matthewmcgarvey (changes requested):
Thanks for the PR 😄 Looks like you need to run `crystal tool format` and make sure running `bin/ameba` passes as well. Could you also include the error message as it is printed out to the terminal in the description?

I think this error will be helpful!

jwoertink (changes requested):
We always love a good error message around here! Thanks for submitting this.

Looks like you need to make sure ameba and the formatter are ran for those to go green. Also just a note on the CHANGELOG.

One thing to just be double clear here is that `needs` can be the same between two different unrelated classes. It's only if it's in the same class, or an inherited class (or maybe mixin) when it should raise.

shortly-portly:
Hi All. I thought I'd run the formatter but I'll do so again (and post the error message in the description). I'll also run  bin/ameba as well. I'll get this done in the next couple of hours. cheers.

shortly-portly:
HI guys - could do with a bit more help. I'm not quite sure what bin/ameba is - I assumed I'd just run it and all would be good with the world but I get an exec format error :-(.

shortly-portly:
Hi Gents - fixed the issues and added an example error message - let me know if there is anything else I need to do. cheers.

matthewmcgarvey:
@shortly-portly Just need to remove the changelog update 👍

CHANGELOG.md
shortly-portly: Hi Matthew - I've removed the Change Log comment.

CHANGELOG.md
shortly-portly: Done

src/lucky/assignable.cr
shortly-portly: Done

shortly-portly:
Sorry guys - left in the failing test (which proves the fix works but won't compile). Will fix now.

matthewmcgarvey (approved):
Nice addition 👍

PR #1305, opened 2020-11-11
WebSockets

### LINKED ISSUE #554: Feature Request: Add out-of-the-box support for websockets
For anyone that comes looking for websocket support, you can dive in to the [built in](https://crystal-lang.org/api/0.25.1/HTTP/WebSocket.html) websocket support for now. 

https://gitter.im/luckyframework/Lobby?at=5b5a51a61be9bb57bcc749bd

However, this is obviously something we should support out of the box with Lucky. I've never actually used websockets, so I don't know if they use HTTP methods or not, but looking at [Kemal docs](http://kemalcr.com/guide/#websockets) I would assume something like

```crystal
class SomeAction < WebsocketAction
  route "/" do
    # maybe context.on_message
    # or have access to socket.on_message here?
  end
end
```

jwoertink:
## Purpose
Fixes #554 

## Description
I wanted to get the discussion moving on how WebSockets might look in Lucky. The idea here would be that Lucky provides a very basic interface to using WebSockets from the backend, but it would be up to the developer to implement the client side portion on their own with whatever tools they know. 

This implementation tries to stick as close to current actions as we can while maintaining all of the type-safety features that Lucky gives us. It shouldn't feel like a ton of setup to get using. 

Imagine you just generated a fresh Lucky app with this in it (maybe the wizard asks if you need websockets?), here is code that would be in your generated app:

```crystal
# src/actions/web_socket_action.cr
class WebSocketAction < Lucky::WebSocketAction
  accepted_formats [:plain_text, :json], default: :plain_text

  # possibly some includes for authentic here?
end

# src/app_server.cr
def middleware : Array(HTTP::Handler)
    [
      Lucky::ForceSSLHandler.new,
      Lucky::HttpMethodOverrideHandler.new,
      Lucky::LogHandler.new,
      Lucky::ErrorHandler.new(action: Errors::Show),
      Lucky::RemoteIpHandler.new,
      Lucky::WebSocketHandler.new, # <-- this line added
      Lucky::RouteHandler.new,
      Lucky::StaticCompressionHandler.new("./public", file_ext: "gz", content_encoding: "gzip"),
      Lucky::StaticFileHandler.new("./public", fallthrough: false, directory_listing: false),
      Lucky::RouteNotFoundHandler.new,
    ] of HTTP::Handler
  end
```

Now for your code:

```crystal
class Chats::Index < WebSocketAction

  ws "/chat" do |socket|
    socket.send("Hi")
    # what should this return? maybe just `socket`?
  end

  def on_message(message)
    # you got a message
    # this may or may not work 🤔 
    websocket.send("I got your message")
  end

  def on_close
    # whatever you do when it closes
  end
end
```

## Checklist

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -58,6 +58,21 @@ module Lucky::Routable
     end
   {% end %}
 
+  # Define a route that responds to a WebSocket request
+  macro ws(path, &block)
+    {% unless path.starts_with?("/") %}
+      {% path.raise "Path must start with a slash. Example: '/#{path}'" %}
+    {% end %}
+
+    {% if block.args.size != 1 %}
+      {% raise "ws takes a block with 1 arg." %}
+    {% end %}
+
+    add_route({{method}}, {{ path }}, {{ @type.name.id }})
+
+    setup_ws_call_method(block)
+  end
+
   # Define a route with a custom HTTP method.
   #
   # Use this method if you need to match a route with a custom HTTP method (verb).
@@ -83,6 +98,35 @@ module Lucky::Routable
     add_route({{method}}, {{ path }}, {{ @type.name.id }})
 
     setup_call_method({{ yield }})
+  end
+
+  # :nodoc:
+  macro setup_ws_call_method(&block)
+
+    abstract def on_message(message)
+    abstract def on_close
+
+    def call(socket : Lucky::WebSocket)
+      # Ensure clients_desired_format is cached by calling it
+      clients_desired_format
+
+      %pipe_result = run_before_pipes
+
+      %response = if %pipe_result.is_a?(Lucky::Response)
+        %pipe_result
+      else
+        {{ block.args.first }} = socket
+        {{ block.body }}
+      end
+
+      %pipe_result = run_after_pipes
+
+      if %pipe_result.is_a?(Lucky::Response)
+        %pipe_result
+      else
+        %response
+      end
+    end
   end
 
   # :nodoc:
--- a/src/lucky/web_socket.cr
+++ b/src/lucky/web_socket.cr
@@ -1,0 +1,9 @@
+module Lucky
+  class WebSocket < HTTP::WebSocketHandler
+    getter proc
+
+    def initialize(@action : Lucky::Action.class, &@proc : HTTP::WebSocket, HTTP::Server::Context -> Void)
+    end
+
+  end
+end
--- a/src/lucky/web_socket.cr
+++ b/src/lucky/web_socket.cr
@@ -4,6 +4,5 @@ module Lucky
 
     def initialize(@action : Lucky::Action.class, &@proc : HTTP::WebSocket, HTTP::Server::Context -> Void)
     end
-
   end
 end
--- a/src/lucky/web_socket_action.cr
+++ b/src/lucky/web_socket_action.cr
@@ -1,0 +1,19 @@
+abstract class Lucky::WebSocketAction < Lucky::Action
+  getter websocket : Lucky::WebSocket
+
+  def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
+    @websocket = Lucky::WebSocket.new(self.class) do |ws|
+      ws.on_message {|message| on_message(message) }
+      ws.on_close { on_close }
+      call(ws)
+    end
+  end
+
+  abstract def call(socket : Lucky::WebSocket)
+
+  def call
+    raise <<-ERROR
+    WebSocketAction must define `call(socket)`
+    ERROR
+  end
+end
--- a/src/lucky/web_socket_action.cr
+++ b/src/lucky/web_socket_action.cr
@@ -3,7 +3,7 @@ abstract class Lucky::WebSocketAction < Lucky::Action
 
   def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
     @websocket = Lucky::WebSocket.new(self.class) do |ws|
-      ws.on_message {|message| on_message(message) }
+      ws.on_message { |message| on_message(message) }
       ws.on_close { on_close }
       call(ws)
     end
--- a/src/lucky/web_socket_handler.cr
+++ b/src/lucky/web_socket_handler.cr
@@ -1,0 +1,26 @@
+class WebSocketHandler
+  include HTTP::Handler
+
+  def call(context : HTTP::Server::Context)
+    if ws_route_found?(context) && websocket_upgrade_request?(context)
+      websocket_action.payload.new(context, websocket_action.params).websocket.call(context)
+    else
+      call_next(context)
+    end
+  end
+
+  private def ws_route_found?(context)
+    !!websocket_action
+  end
+
+  memoize def websocket_action : LuckyRouter::Match(Lucky::Action.class)?
+    Lucky::Router.find_action(:ws, context.request.path)
+  end
+
+  private def websocket_upgrade_request?(context)
+    return unless upgrade = context.request.headers["Upgrade"]?
+    return unless upgrade.compare("websocket", case_insensitive: true) == 0
+
+    context.request.headers.includes_word?("Connection", "Upgrade")
+  end
+end


src/lucky/routable.cr
```diff
@@ -85,6 +100,35 @@ module Lucky::Routable
     setup_call_method({{ yield }})
   end
 
+  # :nodoc:
+  macro setup_ws_call_method(&block)
+
+    abstract def on_message(message)
+    abstract def on_close
```
jwoertink, line 107: My thought was by having these abstract, the websocket can proxy to them, and forces you to set these up. Websockets have a few other methods though.... are any of them as necessary? Is there ever a time you'd use a websocket and not use these?

src/lucky/routable.cr
```diff
@@ -85,6 +100,35 @@ module Lucky::Routable
...
+
+    def call(socket : Lucky::WebSocket)
```
jwoertink, line 109: Actually, I think this signature is wrong and should be `HTTP::WebSocket`

src/lucky/web_socket.cr
```diff
@@ -0,0 +1,8 @@
+module Lucky
+  class WebSocket < HTTP::WebSocketHandler
+    getter proc
+
```
jwoertink, line 4: This class is probably not necessary, but I figure it makes it easier to reference a Lucky namespace, than digging around to see what the built-in crystal one is. Plus we could do some delegation stuff.

src/lucky/web_socket.cr
```diff
@@ -0,0 +1,8 @@
...
+    def initialize(@action : Lucky::Action.class, &@proc : HTTP::WebSocket, HTTP::Server::Context -> Void)
```
jwoertink, line 5: This `@action` isn't used at all, but if you have several different websocket routes, this at least lets us differentiate between them if we need.

src/lucky/web_socket_action.cr
```diff
@@ -0,0 +1,19 @@
+abstract class Lucky::WebSocketAction < Lucky::Action
+  getter websocket : Lucky::WebSocket
+
+  def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
+    @websocket = Lucky::WebSocket.new(self.class) do |ws|
+      ws.on_message { |message| on_message(message) }
+      ws.on_close { on_close }
```
jwoertink, line 7: If we don't make `on_message` and `on_close` abstract, then we can't guarantee they will exist which means we can't do this proxy. You'd just have to use them on your own. Not bad, but is this the better path?

src/lucky/web_socket_action.cr
```diff
@@ -0,0 +1,19 @@
...
+      call(ws)
+    end
+  end
+
+  abstract def call(socket : Lucky::WebSocket)
+
+  def call
+    raise <<-ERROR
+    WebSocketAction must define `call(socket)`
```
jwoertink, line 16: Since we inherit from `Lucky::Action`, you have to define `call`. This really should never get called, but just in case. Also needs a much better error message.

src/lucky/web_socket_handler.cr
```diff
@@ -0,0 +1,26 @@
+class WebSocketHandler
+  include HTTP::Handler
+
+  def call(context : HTTP::Server::Context)
+    if ws_route_found?(context) && websocket_upgrade_request?(context)
+      websocket_action.payload.new(context, websocket_action.params).websocket.call(context)
```
jwoertink, line 6: I should probably rename these, but `websocket_action` is actually a `LuckyRouter::Match` and `payload` is the actual `Lucky::WebSocketAction` class. Then you call the `websocket` getter to call `HTTP::WebSocketHandler#call`.

One thing Lucky does here in a normal action is calls `perform_action` which checks that your route returns a `Lucky::Response` object. What should the websocket routes return? Or should they even return a thing?

src/lucky/web_socket_handler.cr
```diff
@@ -0,0 +1,26 @@
...
+
+  private def ws_route_found?(context)
+    !!websocket_action
+  end
+
+  memoize def websocket_action : LuckyRouter::Match(Lucky::Action.class)?
+    Lucky::Router.find_action(:ws, context.request.path)
+  end
+
+  private def websocket_upgrade_request?(context)
```
jwoertink, line 20: I pulled this from Kemal... 🤷‍♂️

src/lucky/routable.cr
```diff
@@ -58,6 +58,21 @@ module Lucky::Routable
...
+  macro ws(path, &block)
+    {% unless path.starts_with?("/") %}
+      {% path.raise "Path must start with a slash. Example: '/#{path}'" %}
+    {% end %}
+
+    {% if block.args.size != 1 %}
+      {% raise "ws takes a block with 1 arg." %}
+    {% end %}
+
+    add_route({{method}}, {{ path }}, {{ @type.name.id }})
```
matthewmcgarvey, line 71: We will either need to have a separate router for websockets or do something to separate them from http routes. I don't think `{{ method }}` is a thing? We could make that "ws" or something I think?

fernandes:
`accepted_formats [:plain_text, :json], default: :plain_text` doesn't make sense for a WS action

why does `Chats::Index` inherits from `WebSocketAction`, the behavior is totally different

src/lucky/web_socket_handler.cr
```diff
@@ -0,0 +1,26 @@
+class WebSocketHandler
```
matthewmcgarvey, line 1: Namespace?

src/lucky/routable.cr
jwoertink: Yeah, that should be `:ws` there. 😂

jwoertink:
@fernandes With websockets, would you ever need to send an Accept header to say the messages being passed back and forth are of any specific type? Or does that not matter? I'm wondering if there should be a default `accepted_formats` set? 🤔 

Also, I inherited the action from a base class just because that's what you'd normally do in any other action. My hope here is that this can feel like the rest of the actions (i.e. BrowserAction, ApiAction). Then you could have all of your websocket actions inherit from the same parent class in case they all need to do something like ensure you're logged in before connecting or whatever.

confact:
This comes in nice timing. I was actually planning to use @fernandes cable repo for this. 

What I want in WebSocket support is:
* on message handling
* on connected handling
* on disconnect/close handling
* subscribe/channel handling

Most of WebSocket is JSON but can be plain_text. Maybe let it be JSON as default @jwoertink ?

fernandes:
@confact ohhh that's nice to hear! I'm working with @jwoertink in some stability/scalability issues, so I'd advice to not use in production atm, hope to fix this soon 🤞 

anything you can ping me on lucky discord server 😉

jwoertink:
I built a mini chat app using this setup over the weekend. Worked pretty well, but now I have a better idea of how I want to do this. It's going to require a TON more work than I originally wanted, but will be much better in the long run.

I'll try to get a PoC out to flesh out the details, but in the end, I think this may have to be a post 1.0 feature. For now, anyone coming to this looking for WebSocket support, use https://github.com/cable-cr/cable

grepsedawk:
As part of this PR, I think you'll also need to add `wss` to the possibilities for https://github.com/luckyframework/lucky/pull/1662/files#diff-88f9201f12cf1493e98c2358834dbae535ac1be4d45fe1f01429dd526b280ecaR51

jwoertink:
> I have a better idea of how I want to do this
> I'll try to get a PoC out to flesh out the details

I wish I would have done this :sob: I was going to come back to this, and I don't have any of that code I worked on now.... I guess I'll have to just re-imagine the whole thing :face_exhaling:

src/lucky/routable.cr
```diff
@@ -58,6 +58,17 @@ module Lucky::Routable
     end
   {% end %}
 
+  # Define a route that responds to a WebSocket request
+  macro ws(path, &block)
```
jwoertink, line 62: Would need `wss` too

src/lucky/web_socket_action.cr
```diff
@@ -0,0 +1,20 @@
+abstract class Lucky::WebSocketAction < Lucky::Action
+  @socket : HTTP::WebSocket?
+  @handler : HTTP::WebSocketHandler
+
+  def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
+    @handler = HTTP::WebSocketHandler.new do |ws|
+      @socket = ws
+      ws.on_ping { ws.pong("PONG") }
+      call
+    end
```
jwoertink, line 10: This would create a new handler for every websocket action you added. I'm not sure how many a person would add, but it seems like you should only have 1, and the handler would just route to each action based on where it connected. This looks like what [Kemal](https://github.com/kemalcr/kemal/blob/master/src/kemal/websocket_handler.cr) seems to be doing too. A single instance that just adds each route and the handler.

rmarronnier:
Thanks for having started work on this.

Right now, as you or others, I'm using https://github.com/cable-cr/cable in my lucky project.

It's not really hard to setup and the semantics (channel, connection) are familiar for those with a rails background.

The semantics are also really efficient (a single connection by user and several channels, each with its own callbacks).

Using the semantics of a lucky action (a route and a http verb to create or close a connection/channel) will lead to a lot of boilerplate (`create.cr `and `close.cr` files).

I don't think the DX of Lucky::WS::Action is going to beat Cable::Channel with all callbacks and logic in a single file.

But ! What I don't like about cable / action cable is that it relies on a serialized json communication to start / close channels.

The format of those json messages is specific to action cable and requires json serialization capability on the client side.

We could replace this complexity / overhead with a route based logic.

example (shamelessly stolen from cable) : 

```crystal
class ChatChannel < Lucky::WS::Channel
  path: "/ws/chatrooms/:id"

  def subscribed
    stream_from "chat_#{id}"
  end

  def receive(data : String)
    broadcast_message = {} of String => String
    broadcast_message["message"] = data
    broadcast_message["current_user_id"] = connection.current_user.id
    ChatChannel.broadcast_to("chat_#{id}", broadcast_message)
  end

  def unsubscribed
    ChatChannel.broadcast_to("chat_#{id}", "#{connection.current_user.name} left the chat")
  end

  # These lucky actions / routes are automatically generated
  # post "/ws/chatrooms/:id/subscribe"
  # delete "/ws/chatrooms/:id/unsubscribe"
 end
```

It's extremely similar to the Cable::Channel API (maybe `Lucky::WS::Channel < Cable::Channel` ?)

TLDR; We could just extend cable.cr to give it lucky specific semantics like route actions for example.

Or what didn't you like in cable.cr you wish you had in lucky ?

jwoertink:
@rmarronnier I definitely need to come back to this at some point. With my direction, you wouldn't do separate actions for `create` and `close`. The `create` name was just poor naming, but all of your connection would happen in that single action. This is the same on how it works with Kemal. 

So a better example would be more like 

```crystal
# src/actions/chat_messages/channel.cr
class ChatMessages::Channel < Lucky::WebSocketAction
  ws "/chat_messages/:id" do |socket|
    # handle things how you need
  end
end
```

The idea is more that it's lower level letting the developer decide if there should be a concept of channels, or how the subscription and message receiving should work. Then on the front-end it would be all up to you to handle the connections how you want. 

With that said, I totally agree about Cable. I love how the backend is setup, but I'm not a fan of how the messages are handled with the data being a stringified json... but I get why it's like that because if you wanted to map that data, you have no way to know the structure of the hash, but what you do know is that it's always `String => String`, it just happens to be that the value could be parsed in to JSON :grimacing:  

I'm open to other ideas though. Since this PR has fallen behind so much, I'm down to rethink it from the ground up. I still want to get it in, but I don't want to just bring in Cable and force everyone to have to use Cable on the client side.

rmarronnier:
Thanks @jwoertink for the feedback. 
A more low-level / barebone solution is actually quite nice. 
It'd make it easier to work with specific client scenarios (htmx websockets come to mind)
ANd, yeah, you're rigt having a dependency on `cable.cr` is overkill (might as well use `cable.cr` directly)

PR #1321, opened 2020-12-01
Add comment pointing to action html page is called from

matthewmcgarvey:
## Purpose

A `Lucky::HTMLPage` can fail to compile when it is missing assigns from the `Lucky::Action` it's rendered from. This can cause confusion for developers when they only remember one of the actions it was used in and they see all the correct assigns there. The most common problem I've seen is when user's report the error coming from a `NewPage` missing assigns and they forget that the page is also rendered in the failure scenario of the `Create` action.

Related to this discussion as well https://github.com/luckyframework/lucky/discussions/1319

## Description

This adds a little comment to indicate which action the compilation failure is coming from.

### Before

CleanShot 2020-12-01 at 15 34 03@2x

### After

CleanShot 2020-12-01 at 15 33 48@2x

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -71,6 +71,7 @@ module Lucky::Renderable
 
   # :nodoc:
   macro render_html_page(page_class, assigns)
+    # Found in {{ @type.name }}
     view = {{ page_class }}.new(
       context: context,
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
😂  That's pretty clever. I dig it 👍

PR #1322, opened 2020-12-04
adds gen.task task

robacarp:
fixes https://github.com/luckyframework/lucky_cli/issues/541

## Checklist


spec/tasks/gen/task_spec.cr
```diff
@@ -0,0 +1,49 @@
...
+  generator.output = IO::Memory.new
+  generator.print_help_or_call args
+  generator.output
+end
+
+describe Gen::Task do
+  it "generates a task" do
+    with_cleanup do
+      output = run_with_args ["search.reindex"]
+      output.to_s.should contain("Generated ./tasks/search/reindex.cr")
```
robacarp, line 17: I notice that we're adding `to_s` calls in a lot of places with this newer injected IO. It may be an easy win to cleanup the interface so that `should_create_files_with_contents` and friends expect a string rather than an IO.

spec/tasks/gen/task_spec.cr
```diff
@@ -0,0 +1,49 @@
...
+
+  it "displays an error if no name is provided" do
+    expect_raises Exception, /task_name is required/ do
+      output = run_with_args [] of String
+    end
+  end
+
+  it "displays an error if the name is formatted as a class" do
+    run_with_args(["GenericTask"]).to_s.should contain("needs to be formatted with dot notation")
+    run_with_args(["genericTask"]).to_s.should contain("needs to be formatted with dot notation")
```
robacarp, line 31: I opted here to enforce a specific pattern for the task input rather than trying to be smart enough to "just work" because the path forward seemed more obvious.

shard.yml
```diff
@@ -25,6 +25,8 @@ targets:
     main: src/precompiled_tasks/gen/page.cr
   lucky.gen.component:
     main: src/precompiled_tasks/gen/component.cr
+  lucky.gen.task:
```
robacarp, line 28: I have not been successful at actually testing this from a generated project. The sample project doesn't ever get the `bin` directory with the scripts populated. My testing for this is essentially:

```shell
random_dir $ local_lucky init.custom project_name --api --no-auth
random_dir $ cd project_name ; shards
project_name $ ls bin
ls: bin: No such file or directory
```

I think this is a bug in the `script/precompile_tasks` file, or a change in the way shard postinstall is handled. Either way, it seems out of scope for this PR.

Running it from within the lucky src directory produces this output, which leads me to believe that it will work when the problem is resolved.

```shell
lucky$ script/precompile_tasks 
Dependencies are satisfied
Building: lucky.exec
Building: lucky.watch
Building: lucky.gen.model
Building: lucky.gen.action
Building: lucky.gen.action.browser
Building: lucky.gen.action.api
Building: lucky.gen.resource.browser
Building: lucky.gen.page
Building: lucky.gen.component
Building: lucky.gen.task
Building: lucky.secret_key
```

spec/tasks/gen/task_spec.cr
```diff
@@ -0,0 +1,49 @@
...
+
+include CleanupHelper
+include GeneratorHelper
+
+private def run_with_args(args : Array(String))
+  generator = Gen::Task.new
+  generator.output = IO::Memory.new
+  generator.print_help_or_call args
+  generator.output
+end
```
robacarp, line 11: This method feels like it _could_ be helpful in other tests, but I don't want to make that kind of change without a discussion. For a more generic usage:

```crystal
def run_with_args(task : LuckyCli::Task.class, args : Array(String))
  runner = task.new
  runner.output = IO::Memory.new
  runner.print_help_or_call args
  runner.output # maybe with .to_s?
end
```

tasks/gen/task.cr
```diff
@@ -0,0 +1,81 @@
...
+    messages
+  end
+
+  def help_message
+    <<-TEXT
+
+    #{summary}
+
+    Example:
+      lucky gen.task email.monthhly_update
```
matthewmcgarvey, line 56:
```suggestion
      lucky gen.task email.monthly_update
```

matthewmcgarvey (commented):
Looks good to me! Just need to figure out why the specs are failing and it will be good to merge

robacarp:
@matthewmcgarvey I guess I don't know _why_ but taking your typo suggestion seems to have fixed the tests too 🦄

matthewmcgarvey (approved):

Issue #1351, opened 2020-12-05
Boolean action params don't work with false default value


matthewmcgarvey:
**Describe the bug**

We use an or to decide whether or not to return the default value or nil.

https://github.com/luckyframework/lucky/blob/f9f45e804a480a210e5b3a970d9d129121c019bc/src/lucky/routable.cr#L449

**To Reproduce**

Create an action with a param like `param test : Bool = false`. Any request to that action without passing the `test` param will fail.

**Expected behavior**

False default values should work.

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock): master branch
 - Crystal version (`crystal --version`): 0.35.1
 - OS: macOS

matthewmcgarvey:
Removing that conditional reveals another one right below it https://github.com/luckyframework/lucky/blob/f9f45e804a480a210e5b3a970d9d129121c019bc/src/lucky/routable.cr#L453-L454

PR #1352, opened 2020-12-05
Handle boolean params with false default values

matthewmcgarvey:
## Purpose

Fixes https://github.com/luckyframework/lucky/issues/1351

## Description

Three places were using the truthiness of param default values when they shouldn't have been. This causes problems when dealing with boolean params. The solution was to always check if the value is a [Nop](https://crystal-lang.org/api/0.35.1/Crystal/Macros/Nop.html) instead.

## Checklist


jwoertink (approved):
Interesting. I've never seen the `Nop` before.

PR #1365, opened 2020-12-15
Updated UPGRADE_NOTES for 0.25 update

jwoertink:
## Purpose
UPGRADE NOTES for the 0.25 update.

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -98,6 +98,28 @@ brew upgrade lucky
 
   UserQuery.new.to_sql #=> SELECT * FROM users WHERE admin = false
   ```
+- Update: any `has_many through` model association to include the new assocation chain.
+  ```crystal
+  # Before update
+  has_many posts : Post
+  has_many comments : Comment, through: :posts
+
+  # After update
+  # The first in the array is the association you're going through
+  # The second is that through's association.
+  has_many posts : Post
+  has_many comments : Comment, through: [:posts, :comments]
+  ```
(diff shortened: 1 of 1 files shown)


matthewmcgarvey (commented):
The has_many through update is worth mentioning probably

jwoertink:
Good catch. I'll hold off in case you or anyone else sees something I'm missing, but for sure that should go in.

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,128 @@
...
+
+- Run `shards update`
+
+### General updates
+
+- Update: all `Avram::Operation` to implement the new interface.
+  - Your main instance method should be called `run`
+  - The `run` method should return just the value you need. No more `yield self, thing` / `yield self, nil`.
+  - Call the operation with `MyOperation.run(params)` instead of `MyOperation.new(params).submit`
+  - The `MyOperation.run` method class method takes a block that yields the operation, and your return value. Similar to `SaveOperation`.
```
hanneskaeufler, line 32:
```suggestion
  - The `MyOperation.run` class method takes a block that yields the operation, and your return value. Similar to `SaveOperation`.
```

jwoertink:
We should also mention https://github.com/luckyframework/lucky/discussions/1359#discussioncomment-216431

matthewmcgarvey:
There will be places where certain query association methods won't work because of the `belongs_to` fixes. They just need to be singularized where they were plural. When I was updating my app, I was disappointed because crystal didn't offer a "did you mean xxx" in the error message so it could be confusing.

matthewmcgarvey (approved):

PR #1367, opened 2020-12-17
Fix HTTP::Request deprecations

bcardiff:
Ref: crystal-lang/crystal#10029

There is also a related change in a dependency at crystal-loot/exception_page#18

These changes keep compatibility with Crystal 0.35


matthewmcgarvey (approved):
Thanks 🎉

Issue #1271, opened 2020-09-26
Flash messages are always displayed for 2 requests


odinhb:
I'm having trouble with the flash messages.

I'm doing something similar to the generated browser sign_in endpoint:
```crystal
get "/some_endpoint" do
  flash.info = "here is a message"
  html SomePage
end
```
It doesn't matter if I use `flash.keep` or not. The flash message shows up in the response (as it should), but it doesn't go away if I refresh or trigger another request.

I would expect it to disappear in the response to the second request.

Lucky version: 0.24.0

I noticed that in `Shared::FlashMessages` we loop over flash keys using `#each`, but in the implementation of `Lucky::FlashStore` `#each` is delegated to `#all`, which means we are looping over both the `@next` messages and the `@now` messages.

I also noticed that in `Lucky::FlashStore#set` we modify `@next` instead of `@now`, meaning there is no difference between calling `flash.keep` or not, as `@next` gets put in a cookie (i assume thats why `#to_json` is there), sent back by the client and put back in `@now` for the next response, essentially always being "kept" and thus being displayed in both responses.

I appended a monkey-patch in `src/app.cr` which looks like this:
```crystal
class Lucky::FlashStore
  def set(key : Key, value : String) : String
    @now[key.to_s] = value
  end
end
```
And it seems to work fine now, but this is not the correct implementation, is it? I'm finding the implications of the implementation a bit hard to hold in my head.

jwoertink:
Thanks for reporting! Flash has seemed to be pretty tricky. The last major change was https://github.com/luckyframework/lucky/pull/1169 which may have caused some issues. I'll try to dig in soon and see if I can figure out what's up.

matthewmcgarvey:
After reading through Rails' flash implementation, specifically [here](https://github.com/rails/rails/blob/f2bd5197ee55e047f7cc86889b351d262a106e08/actionpack/lib/action_dispatch/middleware/flash.rb#L6-L7), it's becoming clear that we implemented flashes incorrectly.

The idea in rails is that normal flashes set in an action are shown in the _next_ action that handles a request for the user. The idea would be that the create endpoint sets a flash and redirects to the show which actually processes the flash and renders html. If you want to show a flash message in the _current_ action you call `flash.now` https://github.com/rails/rails/blob/f2bd5197ee55e047f7cc86889b351d262a106e08/actionpack/lib/action_dispatch/middleware/flash.rb#L229-L231

[...]

PR #1369, opened 2020-12-25
Lucky::FlashStore refactor

matthewmcgarvey:
## Purpose

Related to https://github.com/luckyframework/lucky/issues/1271

## Description

This does not change any behavior of the `Lucky::FlashStore`. All it does is change the internals to prepare for changes coming up. The main thing is that we remove the independent hashes of `@now` and `@next` in favor of a single `@flashes` hash with a list of `@discards` to know which items in the hash aren't passed along into the next request.

This prepares the code for two things:

- Refactoring for the `Lucky::FlashStore` to be an `Enumerable` that delegates to the `@flashes` instead of having to coordinate the two lists
- Updating the API to provide a way to set flashes that should be shown only in the current request and not the next one (for #1271)

This refactor was inspired by Rails' [implementation](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/flash.rb)

## Checklist

### DIFF
--- a/spec/lucky/cookies/flash_store_spec.cr
+++ b/spec/lucky/cookies/flash_store_spec.cr
@@ -1,19 +1,14 @@
 require "../../spec_helper"
 
-include ContextHelper
-
 describe Lucky::FlashStore do
   describe ".from_session" do
     it "creates a flash store from the json in a session" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Sweet! 👍  Looks good.

PR #1372, opened 2020-12-28
Better error when using route helpers incorrectly

### LINKED ISSUE #1005: Better error when passing incorrect number of params to routes
https://gitter.im/luckyframework/Lobby?at=5e3f48e6b612cc7bb15657d3

It should have a nicer error like:

```
Repositories::Issues::Show expected at least 2 params but got 1

Please include:
  * repository id
  * issue id

[Maybe link to route helpers section]
```

matthewmcgarvey:
## Purpose

Fixes #1005 

## Description

There's an implementation `Lucky::Routable.route` that handles correct usage but when the method is called incorrectly it provides a confusing error message. We're seeing people assume it was a problem in their action and not where the error actually is, where they try to create a link or something using the action class name.

### Before

CleanShot 2020-12-28 at 12 29 40@2x

### After

CleanShot 2020-12-28 at 16 47 25@2x

I'm still kind of disappointed with the error message. I was hoping that by adding the compilation error it would point towards the usage site but it still doesn't.

## Checklist

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -268,12 +268,6 @@ module Lucky::Routable
       Lucky::RouteHelper.new({{ method }}, path).url
     end
 
-    def self.route(
-    # required path variables
-    {% for param in path_params %}
-      {{ param.gsub(/:/, "").id }},
-    {% end %}
-
     {% params_with_defaults = PARAM_DECLARATIONS.select do |decl|
          !decl.value.is_a?(Nop) || decl.type.is_a?(Union) && decl.type.types.last.id == Nil.id
        end %}
@@ -281,10 +275,15 @@ module Lucky::Routable
          params_with_defaults.includes? decl
        end %}
 
-    # params without a default value, could be nilable
+    def self.route(
+    # required path variables
+    {% for param in path_params %}
+      {{ param.gsub(/:/, "").id }},
+    {% end %}
+
+    # required params
     {% for param in params_without_defaults %}
-      {% is_nilable_type = param.type.is_a?(Union) %}
-      {{ param }}{% if is_nilable_type %} = nil{% end %},
+      {{ param }},
     {% end %}
 
     # params with a default value set are always nilable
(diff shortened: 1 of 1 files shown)


src/lucky/routable.cr
```diff
@@ -323,6 +322,26 @@ module Lucky::Routable
       Lucky::RouteHelper.new {{ method }}, path
     end
 
+    def self.route(*_args, **_named_args) : Lucky::RouteHelper
```
jwoertink, line 325: Out of curiosity, if we just added `file = __FILE__, line = __LINE__` to this, would it pick up the correct file and line? It's a little hacky, but it would be nice to figure out how to get the location in here... There's also that weird `<#loc` thing in Crystal that I've always been curious about, but there's no docs on if it would even do what I'm thinking..

robacarp (commented):
This is a big area of confusion for me, and rendering an error which is more helpful than Crystal will help a ton. For me it goes like this:

- Edit the `need` of an action (ohh how easy it is to just add a `need`! just so easy! *spill the bucket of red paint on the floor*)
- Edit "the page that uses the action"
- See this opaque error
- Open up terminal and start grepping for other uses of the ActionName 
- Clean up the mess with a mop until the floor is clean enough to compile again

jwoertink (commented):
Just a few comments.

src/lucky/routable.cr
matthewmcgarvey: I'm not sure what you actually want me to try here. I tried copy/pasting that line but it fails to compile

src/lucky/routable.cr
```diff
@@ -323,6 +322,26 @@ module Lucky::Routable
...
+      \{% raise <<-ERROR
+        Invalid usage of route {{ @type }}
+
+        {% if !requireds.empty? %}
+        Required arguments:
+        {% for req in requireds %}\n- {{ req.id }}{% end %}
+        {% end %}{% if !optionals.empty? %}
+        Optional arguments:
+        {% for opts in optionals %}\n- {{ opts.id }}{% end %}
+        {% end %}
```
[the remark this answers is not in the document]
matthewmcgarvey: I looked into doing something like this but I'd be afraid that if I put that in the error message that it would be an indication of the correct way to call it but the `[ {{ optional }} ]` part is not right.

src/lucky/routable.cr
jwoertink: oh, sorry.

```crystal
def self.route(*_args, _file = __FILE__, _line = __LINE__, **_named_args) : Lucky::RouteHelper
  puts "route at #{_file} #{_line}"
end
```

Though, after writing it, I realize this still won't get us the info at compile-time, so it's probably irrelevant 🤷‍♂️

robacarp (approved):
this will save me buckets of time, thanks @matthewmcgarvey

jwoertink (approved):
👍

PR #1373, opened 2020-12-28
Inline render_html_page to change compilation error

matthewmcgarvey:
## Purpose

No connected issue

### Before

CleanShot 2020-12-28 at 13 20 01@2x

### After

CleanShot 2020-12-28 at 13 20 17@2x

Notice in the "Before" that the first code reference was a macro and not the usage site that caused the error. By removing the `render_html_page` we actually see the usage site in the error which will help people track down issues.

That macro was extracted [here](https://github.com/luckyframework/lucky/commit/3c1af65d46e6953c241c19232b74ed5cee424d58) but you'll notice that this PR achieves the same level of DRY-ness without the separate macro.

## Checklist

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -44,33 +44,9 @@ module Lucky::Renderable
   # end
   # ```
   macro html(page_class = nil, **assigns)
+    {% page_class = page_class || "#{@type.name}Page".id %}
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Ooo! I like this. I think this also might make it easier to pass down the `status` as well without having to change up the interface

PR #1374, opened 2020-12-28
Flash messages discarded unless kept

### LINKED ISSUE #1271: Flash messages are always displayed for 2 requests

matthewmcgarvey:
## Purpose

Fixes #1271 

## Description

Previously, flash messages written to the session by default and used in the next request. That led to seeing flashes twice if both requests rendered HTML. Now, we only pass flash messages onto the next request if `flash.keep` is called. Luckily, we call that when actions redirect https://github.com/luckyframework/lucky/blob/cff93b35f1cc8189b3c6f678dd53baaa089c191d/src/lucky/redirectable.cr#L122-L124

So, the two scenarios should be covered to avoid flash message duplicates. If flashes are set in an action that immediately renders HTML then they are discarded after, but if the action redirects the flash messages will carry over and continue carrying over until HTML is rendered.

I'm slightly wary of this because we are depending on code scattered across the library but I believe this is better than what we have right now.

## Checklist

### DIFF
--- a/spec/lucky/cookies/flash_store_spec.cr
+++ b/spec/lucky/cookies/flash_store_spec.cr
@@ -133,15 +133,14 @@ describe Lucky::FlashStore do
       flash_store.set(:name, "Pauline")
 
       flash_store.get(:name).should eq("Pauline")
-      next_flash(flash_store)["name"]?.should eq("Pauline")
     end
 
-    it "is persisted into the next request" do
-      flash_store = build_flash_store({"success" => "Message saved!"})
+    it "is not persisted into the next request" do
+      flash_store = build_flash_store
 
       flash_store.set(:success, "Message saved again!")
 
-      next_flash(flash_store)["success"]?.should eq("Message saved again!")
+      next_flash(flash_store).should be_empty
     end
   end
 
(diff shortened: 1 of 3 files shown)


jwoertink (approved):
Ah nice. This makes sense to me 👍 And good call on the redirect since we already have flash.keep in there.

PR #1376, opened 2020-12-31
Replace Lucky::Action#call custom error messages with compilation error

### LINKED ISSUE #460: (Re)Add helpful error message when failing to return a Lucky::Response in an action

matthewmcgarvey:
## Purpose

Fixes #460

## Description

If you look at the connected issue, one thing that was not possible to do was display the action that the error was coming from. This change might be controversial but I believe a simple compilation error might be as easy to understand as the custom one and with the comment we are pointing at the action where the error is coming from.

### Before

CleanShot 2020-12-31 at 10 07 39@2x

### After

CleanShot 2020-12-31 at 10 03 17@2x

## Checklist


jwoertink (commented):
hmm... I'm not sure about this one. I can see the benefit here, but it also may be just as confusing since you never define a `call` method yourself, and never actually interface directly with the `Lucky::Response`. 

I feel like this is more of something that's a blocker in the language itself. I'd like to hold off on this change. We might see a change come in Crystal that lets us do what we were originally trying.

PR #1377, opened 2020-12-31
Support html tags without content or block

### LINKED ISSUE #1263: Let HTML tags support hash as parameter without block
Hi

I have big issues with dynamic parameters when rendering HTML tags - I thought I was forced to use NamedTuple which doesn't work as It is very static and I can't set parameters for div tags based on params or parameters in components. It just won't let me.

I tried using a macro to generate NamedTuple, obviously in vain as it doesn't work:
https://gist.github.com/confact/0b07cd259a6fdfadd41008a5ffb72d02

So I wrote in Crystal gitter and we figured div and HTML tags support hash as parameter IF block is added.

so this won't work:
```
div a_hash_var
```
but this will work:
```
div a_hash_var do
end
```

I thought hash was not supported as I didn't try with a block.
This would be nice to be fixed and documented :)

matthewmcgarvey:
## Purpose

Fixes #1263 

## Description

Allows creating an empty tag without supplying a content string or block

```crystal
div class: "empty-div" #=> <div class="empty-div"></div>
```

## Checklist

### DIFF
--- a/spec/lucky/base_tags_spec.cr
+++ b/spec/lucky/base_tags_spec.cr
@@ -25,6 +25,7 @@ describe Lucky::BaseTags do
     view(&.para(42)).should contain "<p>42</p>"
     view(&.para(MySpecialClass.new)).should contain "<p>it works</p>"
     view(&.para(1_i64)).should contain "<p>1</p>"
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Nice!

PR #1381, opened 2021-01-05
Allow reading from UploadedFile#tempfile

### LINKED ISSUE #1301: Getting uploaded file content
When you have a `Lucky::UploadedFile`, there's a `tempfile` method which will return a closed tempfile because the tempfile is created using a block. To get the contents of the file after, you have to re-read the file like:

```crystal
IO::Memory.new(File.read(uploaded_file.path))
```

We should just have this as a built in method like `def body_io` or something.

matthewmcgarvey:
## Purpose

Fixes #1301

## Description

Before, the tempfile was closed once the content was written to it. Now it can be read from.

I saw this in Amber's file class https://github.com/amberframework/amber/blob/58eea3e147ba384fe4cb7c5f0a63da61b357e356/src/amber/router/file.cr#L15-L18

## Checklist

### DIFF
--- a/spec/lucky/uploaded_file_spec.cr
+++ b/spec/lucky/uploaded_file_spec.cr
@@ -14,6 +14,10 @@ describe Lucky::UploadedFile do
     it "returns the tempfile" do
       uploaded_file.tempfile.should be_a(File)
     end
+
+    it "can be read" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
That works for me 👍

Issue #1380, opened 2021-01-05
Lucky::Params does not support query param or form data with array of values


matthewmcgarvey:
**Describe the bug**

To submit a query param with an array of values, it would look like: `?name=sally&name=john`. Crystal's `HTTP::Params` can parse this into an array

```crystal
params = HTTP::Params.parse("name=sally&name=john")
params.fetch_all("name") #=> ["sally", "john"]
params.fetch("name") #=> "sally" (only the first value)
```

The problem is that a lot of the methods on `HTTP::Params` only pulls the first value. The method we call on it in almost every situation is `#to_h` which only returns a `Hash(String, String)` which means it's only the first value of any query param.

**Expected behavior**

We need to provide some way to retrieve these query param array values.

matthewmcgarvey:
Related to https://github.com/luckyframework/avram/issues/408

stephendolan:
I actually think there are more options than repeating the same parameter, if it's helpful!

For example, in Rails if I wanted to send an array of IDs to a route, I'd use `?ids[]=1025&ids[]=1027` rather than `?ids=1025&ids=1027`.

Again, not sure if this would actually help with resolving the issue, but it'd at least make them more easily identifiable.

matthewmcgarvey:
@stephendolan Those are actually the same. If you look at generated forms that will make an array, the name of the input is `ids[]` so if Crystal's `HTTP::Params` was given that, the values could be fetched with a key of `ids[]`.  You're right though that is is a Rails convention to name them in that way and we might consider doing the same.

matthewmcgarvey:
Same goes for form data https://github.com/luckyframework/lucky/blob/80fa91ee0d2abd03093dc4830323e30a8097bd49/src/lucky/params.cr#L469-L476

jwoertink:
One tricky part to this is that there's no actual RFC on how to handle arrays in params. It's basically up to each library to determine that. With that said, we do have `params.many_nested` already.

https://github.com/luckyframework/lucky/blob/80fa91ee0d2abd03093dc4830323e30a8097bd49/src/lucky/params.cr#L259-L268

robacarp:
Yeah, the RFC on http query params is a mess. Lucky currently relies on a small suite of rails frontend tech that is co-opted from rails-ujs. Does adopting the rails convention of array parameters help us play nice with their front end utilities?

matthewmcgarvey:
Here's some specifications I'm thinking about:

## Query Params and Form Data

[...]

PR #1384, opened 2021-01-09
Support multiple form data values for a key

matthewmcgarvey:
## Purpose

Related to https://github.com/luckyframework/lucky/issues/1380

## Description

This does not change the API of `Lucky::Params` at all. This only updates our form data storage to be more like `HTTP::Params` which acts like a hash with a single value but actually stores the values in a list and provides specific methods to access that list.

This does **NOT** provide the API to access the list of form data values since we don't need it yet. Future changes implementing support for multiple values in `Lucky::Params` will need to add it.

I did not add any specs as I consider this a refactor.

## Checklist

### DIFF
--- a/src/lucky/form_data.cr
+++ b/src/lucky/form_data.cr
@@ -1,0 +1,45 @@
+class Lucky::FormData
+  getter params = MultiValueStorage(String).new
+  getter files = MultiValueStorage(Lucky::UploadedFile).new
+
+  def add(part : HTTP::FormData::Part)
+    case part.headers
+    when .includes_word?("Content-Disposition", "filename")
+      files.add(part.name, Lucky::UploadedFile.new(part))
+    else
+      params.add(part.name, part.body.gets_to_end)
+    end
+  end
+
+  # Simpler, generic implementation of HTTP::Params
+  class MultiValueStorage(T)
+    include Enumerable({String, T})
+
+    private getter storage : Hash(String, Array(T))
+
+    def initialize
+      @storage = {} of String => Array(T)
+    end
+
+    def ==(other : self)
+      self.storage == other.storage
+    end
+
+    def []?(key : String) : T?
+      storage[key]?.try(&.first?)
+    end
+
+    def add(key : String, value : T)
+      storage[key] ||= [] of T
+      storage[key] << value
+    end
+
+    def each
+      storage.each do |name, values|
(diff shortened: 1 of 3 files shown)


wontruefree:
Are there no specs for this?

matthewmcgarvey:
@wontruefree since no behavior was changed I didn't add or change any specs. There will be specs added when we add support for retrieving multiple values though

src/lucky/params.cr
```diff
@@ -489,6 +469,10 @@ class Lucky::Params
     request.headers["Content-Type"]?
   end
 
+  private memoize def parse_form_data : Lucky::FormData
+    Lucky::FormDataParser.new(body, request).form_data
```
jwoertink, line 473: 👍

src/lucky/form_data_parser.cr
```diff
@@ -0,0 +1,19 @@
...
+
+  def initialize(@body, @request)
+  end
+
+  def form_data : Lucky::FormData
+    body_io = IO::Memory.new(body)
+    form_data = Lucky::FormData.new
+    boundary = MIME::Multipart.parse_boundary(request.headers["Content-Type"]).to_s
+
+    HTTP::FormData.parse(body_io, boundary.to_s) do |part|
```
jwoertink, line 14: Since you call `.to_s` on line 12, I think we can get rid of the `.to_s` here.

src/lucky/form_data.cr
```diff
@@ -0,0 +1,45 @@
...
+    include Enumerable({String, T})
+
+    private getter storage : Hash(String, Array(T))
+
+    def initialize
+      @storage = {} of String => Array(T)
+    end
+
+    def ==(other : self)
+      self.storage == other.storage
```
jwoertink, line 25: this will fail because `storage` is private. You could make it `protected`, or public is probably ok?

jwoertink (changes requested):
I dig the refactor. I only noticed the one issue, and then the one suggestion

src/lucky/form_data.cr
matthewmcgarvey: Let me just remove it. We don't use it anywhere, it was copied from the HTTP::Params class and I don't know why i felt like keeping it at the time.

jwoertink (approved):

PR #1389, opened 2021-01-14
Add Lucky::Params#get_all

### LINKED ISSUE #1380: Lucky::Params does not support query param or form data with array of values

matthewmcgarvey:
## Purpose

Fixes #1380
Related https://github.com/luckyframework/avram/issues/408

## Description

Allows fetching an array of values from `Lucky::Params`. This is needed to support things like multi-select inputs, and array columns in `Avram::Model`.

## Checklist

### DIFF
--- a/src/lucky/params.cr
+++ b/src/lucky/params.cr
@@ -145,6 +145,37 @@ class Lucky::Params
     route_params[key.to_s]? || body_param(key.to_s) || query_params[key.to_s]?
   end
 
+  # Retrieve values for a given key, returns an empty array if no values found
+  #
+  # Checks in places that could provide multiple values and returns first with values:
+  # - JSON body
+  # - multipart params
+  # - form encoded params
+  # - query params
+  #
+  # Does not check for path variables.
+  #
+  # For all params locations it appends square brackets
+  # so searching for "emails" in query params will look for values with a key of "emails[]"
+  #
+  # ```crystal
+  # params.get_all(:names)    # ["Paul", "Johnny"] : Array(String)
+  # params.get_all("missing") # [] of String : Array(String)
+  # ```
+  def get_all(key : String | Symbol) : Array(String)
+    key = key.to_s
+
+    body_values = if json?
+                    get_all_json(key)
+                  elsif multipart?
+                    get_all_params(multipart_params, key)
+                  else
+                    get_all_params(form_params, key)
+                  end
+
+    body_values || get_all_params(query_params, key) || [] of String
+  end
+
   # Retrieve a file from the params hash, raise if key is absent
   #
   # If no key is found a `Lucky::MissingParamError` will be raised:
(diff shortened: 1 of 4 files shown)


src/lucky/params.cr
```diff
@@ -492,4 +523,20 @@ class Lucky::Params
...
+
+  private def get_all_json(key : String)
+    val = parsed_json[key]?
+    return if val.nil?
+
+    val.as_a?.try(&.map(&.to_s)) || [val.to_s]
+  end
+
+  private def get_all_params(params, key : String)
+    vals = params.fetch_all(key + "[]")
```
matthewmcgarvey, line 535: You might ask, "What about if the user doesn't want to use the `key[]` pattern?"

I'm open to feedback but I think we're pretty clear in Lucky that we prefer convention with a backdoor and the backdoor in this scenario is to access the params directly: `params.from_query.fetch_all("keys")`

src/lucky/params.cr
```diff
@@ -492,4 +523,20 @@ class Lucky::Params
...
+    if vals.any?
+      vals
+    else
+      nil
+    end
```
wontruefree, line 540: What do you think about this?

```suggestion
    params.fetch_all(key + "[]").try &.any?
```

matthewmcgarvey: Doesn't `#any?` return a boolean?  `#fetch_all` always returns an array, the goal is to return nil if the array is empty

src/lucky/params.cr
```diff
@@ -145,6 +145,37 @@ class Lucky::Params
...
+  # Does not check for path variables.
+  #
+  # For all params locations it appends square brackets
+  # so searching for "emails" in query params will look for values with a key of "emails[]"
+  #
+  # ```crystal
+  # params.get_all(:names)    # ["Paul", "Johnny"] : Array(String)
+  # params.get_all("missing") # [] of String : Array(String)
+  # ```
+  def get_all(key : String | Symbol) : Array(String)
```
jwoertink, line 165: I keep wondering if maybe this should raise an error instead of return an empty array like how `get` does. Then have a `get_all?` that returns nil if missing. 🤔

jwoertink (commented):
For the most part, this looks good. I just have the one concern as far as the API for it goes. Also just to be clear that the intended use is different from the `params.many_nested` method, right? This is just for supporting raw arrays in params?

src/lucky/params.cr
matthewmcgarvey: I wasn't sure about that. I ended up copying the way `HTTP::Params#fetch_all` works but I'd be open to changing it. I think Avram would need to do an empty check with the current change or else it would always find a value.

What's the call?

jwoertink: I like that escape hatch with the `fetch_all` which could be the way out. Then I think with this, we should match up with what `get()` does now where if it's missing, `raise Lucky::MissingParamError.new(key.to_s)`, and then add a `get_all?()` that returns `Array(String) | Nil`.

jwoertink (approved):
Sweet! Yeah, this feels pretty good. Thanks for making the change 😃

PR #1394, opened 2021-01-19
Remove broken Lucky::CustomTags#tag overloads

### LINKED ISSUE #1392: No overload matches '...#tag' with type Hash(String, String)
In CustomTags there are two invocations to `tag(Hash)`. And they seem to be unused. In Crystal ~0.36.0 these invocations are causing the custom_tags_spec to fail to compile.

https://github.com/luckyframework/lucky/blob/ec55fc1ee7bb8e423608a31096c63878ab0caf9e/src/lucky/tags/custom_tags.cr#L19-L29

Actually, since the [first commit]( https://github.com/luckyframework/lucky/commit/1367320d91f0deafb92b9fb899d7dd34ab6513a3) of CustomTags is unclear to me if this was working. 

The only definitions of tag methods I can find are in CustomTags and all of them require a tag_name.

matthewmcgarvey:
## Purpose

Fixes #1392 

## Description

The overloads never seemed to work. Both of the overloads are covered by the other two defined overloads.

## Checklist

### DIFF
--- a/src/lucky/tags/custom_tags.cr
+++ b/src/lucky/tags/custom_tags.cr
@@ -16,18 +16,6 @@ module Lucky::CustomTags
     end
   end
 
-  def tag(tag_name : String, content : String | Lucky::AllowedInTags) : Nil
-    tag(EMPTY_HTML_ATTRS) do
-      text content
-    end
-  end
-
-  def tag(tag_name : String, &block) : Nil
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Yeah, if these are already covered by the other ones, then might as well get rid of them!

PR #1395, opened 2021-01-20
Fix typos in UPGRADE_NOTES

vjousse:
- Fix bad class name for Avram::Operation example
- Fix bad `seed/db/` path


jwoertink (approved):
Thanks for catching those!

PR #1398, opened 2021-01-25
Update for Crystal 0.36.0

bcardiff:
A small tweak to #1367 to be usable for 0.36.0

### DIFF
--- a/src/lucky/redirectable.cr
+++ b/src/lucky/redirectable.cr
@@ -155,7 +155,7 @@ module Lucky::Redirectable
   end
 
   private def allowed_host?(referer : String)
-    {% if compare_versions(Crystal::VERSION, "1.0.0-0") > 0 %}
+    {% if compare_versions(Crystal::VERSION, "0.36.0-0") > 0 %}
       if referer_host = URI.parse(referer).hostname
         referer_host == request.hostname
       else


jwoertink (approved):
Looks good to me 👍  Thanks!

bcardiff:
BTW the thing I notice for 0.36 is a spec that will start to fail due to a whitespace

https://github.com/bcardiff/test-ecosystem/runs/1764657336?check_suite_focus=true#step:13:65

```
# Failures:
# 
#   1) Lucky::HTMLPage empty tags can be called with various arguments
#      Failure/Error: view(&.br).should eq %(<br>)
# 
#        Expected: "<br>"
#             got: "<br> "
# 
#      # spec/lucky/html_page_spec.cr:105
# 
# Finished in 10.08 seconds
# 604 examples, 1 failures, 0 errors, 0 pending
```

Since it's html I figure is not an issue. I haven't tracked down where that whitespace is coming from.

jwoertink:
That's strange 🤔  It's coming from here https://github.com/luckyframework/lucky/blob/master/src/lucky/tags/base_tags.cr#L75 which actually does have that space after it. But I wonder why it would now just start failing....

jwoertink:
That spec is here https://github.com/luckyframework/lucky/blob/master/spec/lucky/html_page_spec.cr#L105 Maybe it's just that the spec `eq` didn't account for spaces at the end before and now it does? Either way, easy fix! I'll open an issue to track it.

PR #1405, opened 2021-02-03
Fix Avram::PermittedAttribute usages

### LINKED ISSUE #1404: options_for_select doesn't allow nil values anymore
**Describe the bug**

Since [this commit](https://github.com/luckyframework/avram/commit/03d1a7bcc981a75b103c2f54b81a430096ea0c84) in Avram, `Avram::PermittedAttribute` no longer has it's generic `T` nilable. It's nilable inside the attribute but the generic is never nilable. That breaks `options_for_select` since it restricts the values to be of type `T` and depended on the nilable type being there. https://github.com/luckyframework/lucky/blob/feead7fe1aa77dd28e965a1ebcb984d5644f5c91/src/lucky/tags/select_helpers.cr#L8

**Expected behavior**

`options_for_select` needs to change it's definition to allow for nilable `T` in the `select_options` type restriction

matthewmcgarvey:
## Purpose

Fixes #1404

## Description

Various places we were depending on the old way Avram worked where if the column was nilable, the attribute generated had a nilable generic `column name : String?` became `Avram::PermittedAttribute(String?)` but that is no longer the case. nilable and non-nilable columns both create a `Avram::PermittedAttribute(String)`

## Checklist

### DIFF
--- a/spec/lucky/select_helpers_spec.cr
+++ b/spec/lucky/select_helpers_spec.cr
@@ -1,5 +1,7 @@
 require "../spec_helper"
 
+include ContextHelper
+
 private class TestPage
   include Lucky::HTMLPage
 
@@ -29,7 +31,7 @@ end
 
 class SomeFormWithCompany
   def company_id
-    Avram::PermittedAttribute(Int32?).new(
+    Avram::PermittedAttribute(Int32).new(
       name: :company_id,
       param: "1",
       value: nil,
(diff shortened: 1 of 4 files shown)


spec/lucky/select_helpers_spec.cr
```diff
@@ -1,5 +1,7 @@
 require "../spec_helper"
 
+include ContextHelper
+
```
matthewmcgarvey, line 4: This was necessary because this file was dependent on it but the problem only occurred if you ran this file individually.

src/lucky/tags/input_helpers.cr
```diff
@@ -107,7 +107,7 @@ module Lucky::InputHelpers
   # radio(attribute, "checked_value", attrs: [:required])
   # # => <input type="radio" id="param_key_attribute_name_checked_value" name="param_key:attribute_name" value="checked_value" checked="true" required />
   # ```
-  def radio(field : Avram::PermittedAttribute(String?),
+  def radio(field : Avram::PermittedAttribute(String),
```
matthewmcgarvey, line 110: I'm not entirely sure what these changes entail. How did a non-nilable string attribute behave before?

jwoertink: None of my apps really utilize the form stuff much, so I'm not too sure 😕

jwoertink (approved):
Yeah, I think this should be fine... I hope 😂

Issue #1383, opened 2021-01-08
Lucky depends on Avrams type extensions for param parsing


matthewmcgarvey:
**Description**

https://github.com/luckyframework/lucky/blob/eb01295064a575bca6bebff8c2e678180ca7849c/src/lucky/routable.cr#L480-L484

Lucky does not define the nested `Lucky` namespaces of types with the `parse` function. It comes from [Avram's type extensions](https://github.com/luckyframework/avram/blob/b44793857899a06b4c3b9d45cdd60f0164ec9fd9/src/avram/charms/bool_extensions.cr#L14). This is one of a couple places that keeps Lucky tied to Avram for now.

**Expected behavior**

While it might seem weird, I believe we should duplicate whatever code we are depending on from Avram into Lucky.

PR #1419, opened 2021-02-23
Begin Avram extraction (first step module)

matthewmcgarvey:
## Purpose

Related to https://github.com/luckyframework/lucky/discussions/1402
Related to https://github.com/luckyframework/lucky/issues/1383

## Description

Right now Lucky depends on Avram and there's no way to remove the dependency. This is the first step in officially separating the two. It moves all the things that explicitly reference Avram into it's own module. The next step is to extract a library and add that as a dependency.

## Checklist


jwoertink (approved):
Let's do it! 👍

PR #1423, opened 2021-03-02
Pipe events

jwoertink:
## Purpose
This is an extension required for the upcoming [Breeze](https://github.com/luckyframework/breeze/pull/5) updates.

## Description
Similar to what we've added in Avram, this adds in a new PulsarEvent that publishes each time a pipe runs. When a pipe is executed, we publish if the pipe was continued or halted for both before and after pipes.

Right now, there's nothing subscribing to these events, so when they publish, nothing actually happens. In the future, Breeze will subscribe to these to track the pipes. 

## Checklist


src/lucky/events/pipe_event.cr
```diff
@@ -0,0 +1,15 @@
+class Lucky::Events::PipeEvent < Pulsar::Event
```
jwoertink, line 1: I originally debated on if this should be `BeforePipeEvent` and `AfterPipeEvent`, or just make them the same and pass through the position of the pipe... I'm open to other naming options if this sounds confusing.

src/lucky/action_pipes.cr
```diff
@@ -178,4 +182,20 @@ module Lucky::ActionPipes
...
+      name: pipe_name,
+      position: Lucky::Events::PipeEvent::Position::Before,
+      continued: continued
+    )
+  end
+
+  def publish_after_event(pipe_name : String, continued : Bool)
+    Lucky::Events::PipeEvent.publish(
+      name: pipe_name,
+      position: Lucky::Events::PipeEvent::Position::After,
```
jwoertink, line 197: I tested in a plain crystal class that I could pass `:after` here, but for some reason in this scenario, it throws a no method overload with `position : Symbol` 🤔  It's probably fine though since this isn't really user facing.

jwoertink:
@paulcsmith if you get a free min, would love to get your thoughts on this. No rush since I'll probably just update breeze and wait for the next Lucky release to add this stuff in

paulcsmith (commented):
I think a `position` with one event class makes sense 👍

src/lucky/events/pipe_event.cr
```diff
@@ -0,0 +1,15 @@
...
+  getter :name, :position, :continued
+
+  enum Position
+    Before
+    After
+  end
+
```
paulcsmith, line 8: I'd prob add `after?` and `before?` method just to make it easier to determine if the event was before or after.

paulcsmith (commented):
Looks good 👍

src/lucky/events/pipe_event.cr
jwoertink: ah, good point. I like that

jwoertink:
Awesome! Thanks for the review!

src/lucky/action_pipes.cr
```diff
@@ -178,4 +182,20 @@ module Lucky::ActionPipes
   def continue : Lucky::ActionPipes::Continue
     Lucky::ActionPipes::Continue.new
   end
+
+  def publish_before_event(pipe_name : String, continued : Bool)
```
matthewmcgarvey, line 186: These methods should probably be private right?

jwoertink: Good catch

matthewmcgarvey (approved):

PR #1428, opened 2021-03-10
Catch and error when href passed to link

### LINKED ISSUE #1415: link lets you pass href as well
This code gives you a weird side effect:

```crystal
# This code
link "users", to: Users::Index, href: "#"

# Generates
# <a href="#">users</a>
```

Normally you wouldn't ever write this, but I had an anchor as a placeholder that had 150k tailwind classes, so I didn't see the end of the method 😂  so it was like `link to: Thing, class: ".... forever ...", href: "#"` and I couldn't figure out why it wouldn't render the link properly. 

I think if you use `link`, you shouldn't be able to add the `href` property. If you want to, just use `a()`.

matthewmcgarvey:
## Purpose

Fixes #1415 

## Description

Link is a helper method that allows you to pass lucky routes instead of specifying the `href` string specifically. This adds a method overload to catch when `href` is accidentally passed.

Before I was able to add the overload, I had to refactor all the methods to have them funnel into one method so that an overload wouldn't have to be added for every different way you can call `link`

## Checklist

### DIFF
--- a/src/lucky/tags/link_helpers.cr
+++ b/src/lucky/tags/link_helpers.cr
@@ -1,12 +1,12 @@
 module Lucky::LinkHelpers
   def link(text, to : Lucky::RouteHelper, attrs : Array(Symbol) = [] of Symbol, **html_options) : Nil
-    link(to, attrs, **html_options) do
+    link(**html_options, to: to, attrs: attrs) do
       text text
     end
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Nice! 👍

PR #1317, opened 2020-11-28
Adds <blockquote> and <cite> to  Lucky::BaseTags

rmarronnier:
Fixes #1316 
Adds `<blockquote>` and `<cite>` html tags to `Lucky::BaseTags`.

## Checklist
* [X] - An issue already exists detailing the issue/or feature request that this PR fixes
* [X] - All specs are formatted with `crystal tool format spec src`
* [ ] - Inline documentation has been added and/or updated
* [X] - Lucky builds on docker with `./script/setup`
* [X] - All builds and specs pass on docker with `./script/test`

Issue #1414, opened 2021-02-16
Cryptic compiler error when using `title()` method as variable in templates


vjousse:
**Describe the bug**
When using the `title()` function in a template as a variable `title` by mistake, the compiler displays a strange error message. 

**To Reproduce**
Don't define the `title` variable in your template and try to use it.

```crystal
h1 title, class: "text-5xl font-serif font-bold pt-20 pb-16 text-center"
```

You'll get an error message like this one:
```crystal
web          | In lib/lucky/src/lucky/tags/base_tags.cr:137:15
web          |
web          |  137 | tag_attrs.each do |key, value|
web          |                  ^---
web          | Error: undefined method 'each' for Nil
```

**Expected behavior**
A better error message saying, at least, that the problem is with the `title` variable/function.

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock): 0.26
 - Crystal version (`crystal --version`): 0.36.1
 - OS: Archlinux

jwoertink:
Ok, so digging in more, this error comes up when you pass ANY type that isn't `String`. In the above error it says "each for Nil" because the return type of the `title()` method is `Nil`. You can get this same error by passing a `Time` object in.

```crystal
div user.created_at
```

```
web          | In lib/lucky/src/lucky/tags/base_tags.cr:137:15
web          |
web          |  137 | tag_attrs.each do |key, value|
web          |                  ^---
web          | Error: undefined method 'each' for Time
```

Where in this one, it says `Time`. 

So really what we need is an error message that says "HTML tags must pass a String" or something. I think if we do an override that catches a few of the common types (Nil, Time, Ints, Floats...) that may catch the bulk of these.

PR #1433, opened 2021-03-14
Adds remaining missing html tags

rmarronnier:
## Purpose

Follow up to #1317

All added tags are all present and not deprecated in the current HTML Living Standard

https://developer.mozilla.org/en-US/docs/Web/HTML/Element was my source of information

## Checklist
* [x ] - Lucky builds on docker with `./script/setup`
* [x ] - All builds and specs pass on docker with `./script/test`


stephendolan (approved):
Looks pretty good! Thanks for submitting this, @rmarronnier!

@jwoertink / @matthewmcgarvey ,

Thoughts on breaking out this array to separate lines? It'll make the file super long, but would make reviewing diffs like this way easier 😄

rmarronnier:
Good idea. Made it into multilines ;-) 

It'll allow commenting specific tags if needed.

jwoertink:
I wonder how some of these tags will affect some apps 🤔  Like, I use the variable `data` all over the place on my pages. Also, how common are some of these tags? You can always do `tag("data")`, so I'm wondering if some of these tags that are less common would be better as just using `tag` as needed. It's nice to have every tag to make our set complete, but if that comes at a cost of existing apps getting errors similar to https://github.com/luckyframework/lucky/issues/1414 then that might not be good. @stephendolan do you think there's any issue here?

But as for the multi-line, I think that's fine 👍

stephendolan:
@jwoertink My gut says we should work through those issues as they come up and support the full HTML spec by default for our Lucky HTML views.

Otherwise, we'll probably end up in a position where we have to continually justify why some tags are included and others aren't, perpetually bike-shedding with folks that have clear use cases in their apps where specific tag helpers would be beneficial.

jwoertink:
@stephendolan Sounds good to me!

jwoertink (approved):

PR #1445, opened 2021-03-28
Crystal 1.0 support

jwoertink:
## Purpose
Update for Crystal 1.0 support

## Description
The only main issue to support Crystal 1.0 is that cookie values are now more strict according to rfc6265. Crystal will raise `IO::Error` when setting a cookie value that contains invalid characters, so I added a nicer error that tells you which cookie is getting the error, and which characters to be on the look for.

NOTE:
This PR points all shards to master branch, but it's sort of 🐔 🥚  while everything gets updated.


spec/lucky/cookies/cookie_jar_spec.cr
```diff
@@ -117,7 +126,7 @@ describe Lucky::CookieJar do
     it "raises an error if the cookie is > 4096 bytes" do
       expect_raises(Lucky::CookieOverflowError) do
         jar = Lucky::CookieJar.empty_jar
-        jar.set_raw(:overflow, "x" * (4097 - 27)) # "overflow=x...x; path=/; HttpOnly",
+        jar.set_raw(:overflow, "x" * 4097) # "overflow=x...x; HttpOnly",
```
jwoertink, line 129: I'm not sure why there was maths here, but it wasn't throwing an exception, so I just made it the raw number because `"x" * 4097` is `4097 bytes`

src/lucky/cookies/cookie_jar.cr
```diff
@@ -1,6 +1,6 @@
 class Lucky::CookieJar
   MAX_COOKIE_SIZE         = 4096
-  LUCKY_ENCRYPTION_PREFIX = Base64.encode("lucky") + "--"
+  LUCKY_ENCRYPTION_PREFIX = Base64.strict_encode("lucky") + "--"
```
jwoertink, line 3: `Base64.encode` adds a newline at the end, but newline characters are invalid cookie values.

PR #1447, opened 2021-03-31
Use the latest ameba

jwoertink:
This allows us to use the latest ameba. Also hoping to see all green from the CI

### DIFF
--- a/spec/lucky/cookies/flash_store_spec.cr
+++ b/spec/lucky/cookies/flash_store_spec.cr
@@ -92,12 +92,14 @@ describe Lucky::FlashStore do
     it "returns true if there are key/value pairs" do
       flash_store = build_flash_store({"some_key" => "some_value"})
 
+      # ameba:disable Performance/AnyInsteadOfEmpty
       flash_store.any?.should be_true
     end
 
     it "returns false if there are no key/value pairs" do
       flash_store = build_flash_store
 
+      # ameba:disable Performance/AnyInsteadOfEmpty
       flash_store.any?.should be_false
     end
   end
--- a/src/lucky/memoizable.cr
+++ b/src/lucky/memoizable.cr
@@ -21,7 +21,10 @@ module Lucky::Memoizable
   # Equality (==) is used for checking on argument updates.
   macro memoize(method_def)
     {% raise "You must define a return type for memoized methods" if method_def.return_type.is_a?(Nop) %}
-    {% raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop) %}
+    {%
+    # ameba:disable Style/IsAFilter
+    raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop)
+    %}
 
     @__memoized_{{method_def.name}} : Tuple(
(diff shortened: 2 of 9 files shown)


jwoertink:
I swear I ran that locally and it all passed 🤔

spec/lucky/cookies/flash_store_spec.cr
```diff
@@ -92,12 +92,14 @@ describe Lucky::FlashStore do
     it "returns true if there are key/value pairs" do
       flash_store = build_flash_store({"some_key" => "some_value"})
 
+      # ameba:disable Performance/AnyInsteadOfEmpty
       flash_store.any?.should be_true
```
jwoertink, line 96: This suggested doing `!(flash_store.empty?).should`, but the spec is actually testing the `any?` method 😂

src/lucky/memoizable.cr
```diff
@@ -21,7 +21,10 @@ module Lucky::Memoizable
   # Equality (==) is used for checking on argument updates.
   macro memoize(method_def)
     {% raise "You must define a return type for memoized methods" if method_def.return_type.is_a?(Nop) %}
-    {% raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop) %}
+    {%
+    # ameba:disable Style/IsAFilter
+    raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop)
```
jwoertink, line 26: This actually can't be updated. Filed issue on Ameba: https://github.com/crystal-ameba/ameba/issues/219

src/lucky/mime_type.cr
```diff
@@ -26,7 +26,7 @@ class Lucky::MimeType
   end
 
   def self.known_formats : Array(Symbol)
-    accept_header_formats.values.uniq
+    accept_header_formats.values.uniq!
```
jwoertink, line 29: Ameba suggested this for performance, but I'm not too sold on it. I'll disable if it seems to cause any issues.

tasks/watch.cr
```diff
@@ -3,7 +3,6 @@ require "option_parser"
 require "colorize"
 require "yaml"
 require "../src/lucky/server_settings"
-require "option_parser"
```
jwoertink, line 6: Love this! 😂

matthewmcgarvey (approved):

jwoertink:
Turns out this one might be a bit trickier than I thought. Turns out that due to how the cookies changed, we either have to drop support for 0.36.1, or do a few things like:

```crystal
cookies = {% if compare_versions(Crystal::VERSION, "0.36.1") > 0 %}
        HTTP::Cookies.from_client_headers(headers)
      {% else %}
        HTTP::Cookies.from_headers(headers)
      {% end %}
```

I'll think about this one a bit more to see how we can continue to support both before I go merging this in.

matthewmcgarvey:
@jwoertink I can put up a PR for the cookie thing. This pr can go through

jwoertink:
Ok. I was gonna wait on some updates for Lucky Cli, but I think I'll just merge this as is so you can have at it.

PR #1448, opened 2021-04-01
Fix spec to work on Crystal 0.36.1 and 1.0.0

matthewmcgarvey:


jwoertink (approved):
Ahh.. It was all the specs! 😂  wow... I went a little crazy in my branch locally

PR #1452, opened 2021-04-02
Support VueJS style attributes on tag_defaults

### LINKED ISSUE #1409: Unable to use specific VueJS attribute in Shared::Field
**Describe the bug**
When adding some VueJS attributes like `v-on:change` or `@change` to a `select_input()` inside a `Shared:Field` it produce an error.

**To Reproduce**
Install VueJS on a lucky application (of course :) )
Create a form with a line like this:
`mount Shared::Field, operation.some_field, &.select_input({"@change" => "onchoose($event)"}) do`

**Expected behavior**
We should have the attribute added to the select HTML element like:
`<select @change="onchoose($event)">`
We also need to have the ability to use `v-on:change`.

 - Lucky version: 0.25.0
 - Crystal version: 0.35.1
 - OS: MacOS BigSur

matthewmcgarvey:
## Purpose

Fixes #1409 

## Description

Allows using VueJS style attributes (like `@click`) when using `tag_defaults`. This only works on Crystal 1.0.0 because of https://github.com/crystal-lang/crystal/pull/10388

## Checklist

### DIFF
--- a/spec/lucky/tag_defaults_spec.cr
+++ b/spec/lucky/tag_defaults_spec.cr
@@ -44,9 +44,19 @@ private class TestTagDefaultsPage
       end
     end
 
+    {% if compare_versions(Crystal::VERSION, "0.36.1") > 0 %}
+      tag_defaults do |tag_builder|
+        tag_builder.div "@click": "onclick($event)" do
+          text "block content"
+        end
+      end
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Boom

PR #1453, opened 2021-04-02
Allow creating custom tag with name and hash only

### LINKED ISSUE #1418: Custom tags don't support Hash
**Describe the bug**
custom tags don't support Hash as params, but divs and others do. This makes it hard to make custom tags work with flexible data.

**To Reproduce**
make a normal `Hash(String, String)` called `params` with variables.
make a code like: 
```
tag "turbo-frame", params do
  text "test"
end
``` 

If you make it with div like:
```
div params do
  text "test"
end
``` 
It will work.

**Expected behavior**
Have the same consistent behavior everywhere. Make custom tags work as standard tags.

 - Lucky version (check in shard.lock): 0.26
 - Crystal version (`crystal --version`): 0.36.1
 - OS: Linux

matthewmcgarvey:
## Purpose

Fixes #1418 

## Checklist


jwoertink (approved):
🎉

PR #1454, opened 2021-04-02
Update the build.release task

### LINKED ISSUE #1387: The lucky build task should emit to bin
We added in https://github.com/luckyframework/lucky/pull/1358 which builds `start_server` in to `bin`, but missed this task which still emits the binary to the root directory.

https://github.com/luckyframework/lucky/blob/5f8a9d1836a3fbeb40b4570ec2f44c0f47d4522c/tasks/build/release.cr#L13

jwoertink:
## Purpose
Fixes #1387

## Description
We should now be building all binaries to the `bin/` dir. When we updated to do this, the `build.release` task was missed. This also uses the `output` property from LuckyCli.

One strange thing I couldn't figure out is that `log "Building binary with '#{command}"` prints out to the screen now 🤔  Even though it's a `IO::Memory`, it still seems to print just that one message. It doesn't print out the other one... 

## Checklist


tasks/build/release.cr
```diff
@@ -6,20 +6,19 @@ class Build::Release < LuckyCli::Task
 
   summary "Compile app for production"
 
-  def initialize(@io : IO = STDOUT, @error_io : IO = STDERR)
-  end
+  property error_io : IO = STDERR
```
jwoertink, line 9: This should be added in to LuckyCli::Task directly to match up with `output`.

tasks/build/release.cr
```diff
@@ -6,20 +6,19 @@ class Build::Release < LuckyCli::Task
...
 
   def call
-    command = "crystal build --release src/start_server.cr"
+    command = "crystal build --release src/start_server.cr -o bin/start_server"
 
     log "Building binary with '#{command}"
```
jwoertink, line 14: Why does this print out, but the one below doesn't?

Screen Shot 2021-04-02 at 2 46 41 PM

matthewmcgarvey: Maybe try adding an else and check if the process isn't completing successfully?

jwoertink: ah. Good call. It is failing.. 🤔  but I guess why.

jwoertink: oh, but that still doesn't explain why `IO::Memory` is printing to STDOUT 🤨

matthewmcgarvey: Ohh, it's actually the second spec that is printing to stdout, it doesn't have it's output changed, only error_io

jwoertink: ooohhhh 😂  yup. That fixes it. Thanks!

matthewmcgarvey (approved):

PR #1456, opened 2021-04-03
Better errors when passing invalid values to html tags

### LINKED ISSUE #1414: Cryptic compiler error when using `title()` method as variable in templates

jwoertink:
## Purpose
Fixes #1414

## Description
HTML tags require either a String content, or an object that includes `Lucky::AllowedInTags`. These are any Ints, UUID, and Bool by default. They don't allow `Time` because when calling `to_s`, we don't know what format you want it in, and they don't allow `nil` because that could lead to invalid displays. So for `Nil` and `Time`, I've added 2 new compile-time messages to make it easier to know what's going on. 

The two most common cases are passing `nil` which could also be from using an html tag you thought was another method (i.e. `title`).

```crystal
# You think you're using your title(), but you might actually be calling <title></title> which returns nil
h1(title)
```

and passing in a time object like `created_at`.

```crystal
# What format did you want here?
div(user.created_at)
```

Here's what they look like:

Screen Shot 2021-04-03 at 8 23 39 AM
Screen Shot 2021-04-03 at 8 24 29 AM

## Checklist

### DIFF
--- a/src/lucky/tags/base_tags.cr
+++ b/src/lucky/tags/base_tags.cr
@@ -134,17 +134,11 @@ module Lucky::BaseTags
       \{%
         raise <<-ERROR
           HTML tags content must be a String or Lucky::AllowedInTags object, not nil.
-          Be sure the content isn't a nil value. For nested HTML tags, use the block
-          method overload
 
-          Example...
+          Try this...
 
-            {{method_name.id}}(title, class: "header")
-
-            to
-
-            {{method_name.id}}(class: "header") do
-              title "My Site Title"
+            if value = some_nilable_value
+              {{method_name.id}}(value, class: "header")
             end
 
           ERROR
--- a/src/lucky/tags/base_tags.cr
+++ b/src/lucky/tags/base_tags.cr
@@ -160,7 +160,7 @@ module Lucky::BaseTags
       \{%
         raise <<-ERROR
           HTML tags content must be a String or Lucky::AllowedInTags object.
-          {{method_name.id}} recieved a Time object which has an ambiguous display format.
+          {{method_name.id}} received a Time object which has an ambiguous display format.
 
           Try this...
 
--- a/src/lucky/tags/base_tags.cr
+++ b/src/lucky/tags/base_tags.cr
@@ -125,6 +125,51 @@ module Lucky::BaseTags
       end
     end
 
+    def {{method_name.id}}(
+        content : Nil,
+        options = EMPTY_HTML_ATTRS,
+        attrs : Array(Symbol) = [] of Symbol,
+        **other_options
+      ) : Nil
+      \{%
+        raise <<-ERROR
+          HTML tags content must be a String or Lucky::AllowedInTags object, not nil.
+          Be sure the content isn't a nil value. For nested HTML tags, use the block
+          method overload
+
+          Example...
+
+            {{method_name.id}}(title, class: "header")
+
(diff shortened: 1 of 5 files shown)


src/lucky/tags/base_tags.cr
```diff
@@ -125,6 +125,51 @@ module Lucky::BaseTags
...
+    def {{method_name.id}}(
+        content : Time,
+        options = EMPTY_HTML_ATTRS,
+        attrs : Array(Symbol) = [] of Symbol,
+        **other_options
+      ) : Nil
+      \{%
+        raise <<-ERROR
+          HTML tags content must be a String or Lucky::AllowedInTags object.
+          {{method_name.id}} recieved a Time object which has an ambiguous display format.
```
matthewmcgarvey, line 163:
```suggestion
          {{method_name.id}} received a Time object which has an ambiguous display format.
```

src/lucky/tags/base_tags.cr
```diff
@@ -125,6 +125,51 @@ module Lucky::BaseTags
...
+
+          Example...
+
+            {{method_name.id}}(title, class: "header")
+
+            to
+
+            {{method_name.id}}(class: "header") do
+              title "My Site Title"
+            end
```
matthewmcgarvey, line 148: Idk how I feel about this. I've accidentally passed in a nil a couple times but it was always a nilable value. Could that be the example instead and we just add a sentence to the above that's like `Maybe you accidentally called a method that renders html?` but more eloquently of course

jwoertink: Yeah, this felt a little weird to me too, but I wanted to showcase the original error since it felt like a fairly common use case. Maybe I do something like this?

```crystal
<<-ERROR
HTML tags content must be a String or Lucky::AllowedInTags object, not nil.

Maybe you called a method that also renders html? (e.g. title, address, summary, etc...)

Try this...

  if value = some_nilable_value
    {{method_name.id}}(value, class: "header")
  end

ERROR
```

matthewmcgarvey: Still feels like that sentence and the example are connected when they're not. I think the error without that sentence wuold be enough for the user to have something to go on to figure out where they went wrong.

jwoertink: So remove the example code completely and just leave the error? Or you're saying remove the "Maybe you called a method" line?

matthewmcgarvey: The last thing about accidentally calling a method that renders html is what I think can go

matthewmcgarvey (approved):
❤️

PR #1459, opened 2021-04-04
Replace LuckyCli with LuckyTask

matthewmcgarvey:
Removing heavy lucky_cli dependency with just the task library

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -61,6 +61,9 @@ dependencies:
   pulsar:
     github: luckyframework/pulsar
     version: ~> 0.2.2
+  teeplate:
+    github: luckyframework/teeplate
+    version: ~> 0.8.3
 
 development_dependencies:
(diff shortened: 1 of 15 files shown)


shard.yml
```diff
@@ -61,6 +61,9 @@ dependencies:
   pulsar:
     github: luckyframework/pulsar
     version: ~> 0.2.2
+  teeplate:
+    github: luckyframework/teeplate
```
jwoertink, line 65: 😆  oh well.. this should still get rid of the circular dependency

jwoertink (approved):
🚀

shard.yml
matthewmcgarvey: It's only ever required by tasks and the tasks aren't required by the main `lucky.cr` file so it won't affect compile time of the app

PR #1462, opened 2021-04-09
UPGRADE_NOTES for 0.27

jwoertink:
## Purpose
Updating UPGRADE_NOTES to include updates for 0.27

### DIFF
--- a/UPGRADE_NOTES.md
+++ b/UPGRADE_NOTES.md
@@ -50,7 +50,17 @@ brew upgrade lucky
   LuckyTask::Runner.run
   ```
 - Update: all tasks in your `tasks/` directory to inherit from `LuckyTask::Task` instead of `LuckyCli::Task`. (e.g. `Db::Seed::RequiredData < LuckyCli::Task` -> `Db::Seed::RequiredData < LuckyTask::Task`)
+- Update: your `config/cookies.cr` with a default cookie path of `"/"`.
+  ```crystal
+  Lucky::CookieJar.configure do |settings|
+    settings.on_set = ->(cookie : HTTP::Cookie) {
+      # ... other defaults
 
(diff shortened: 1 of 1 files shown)


UPGRADE_NOTES.md
```diff
@@ -1,3 +1,63 @@
...
+
+- Update versions in `shard.yml`
+  - Crystal should be `">= 0.36.1, < 2.0.0"`
+  - Lucky should be `~> 0.27.0`
+  - Authentic should be `~> 0.7.3`
+  - Carbon should be `~> 0.1.4`
+  - Dotenv should be `~> 1.0.0` or replace with [LuckyEnv 0.1.0](https://github.com/luckyframework/lucky_env)
+  - LuckyFlow should be `~> 0.7.3`
+  - JWT (if you use Auth) should be `~> 1.5.1`
+  - LuckyTask needs to be added as a dependency
```
matthewmcgarvey, line 26: This isn't necessary as it will be a transitive dependency from lucky but it doesn't hurt I guess

jwoertink: oh right. So maybe this doesn't need to be here? https://github.com/luckyframework/lucky_cli/blob/master/src/generators/web.cr#L166-L168

UPGRADE_NOTES.md
```diff
@@ -1,3 +1,63 @@
...
+    ```
+    lucky_task:
+      github: luckyframework/lucky_task
+      version: ~> 0.1.0
+    ```
+
+- Run `shards update`
+
+### General updates
```
matthewmcgarvey, line 35: Need to update the cookie block with the path thingy, you know what I mean!

matthewmcgarvey (changes requested):

matthewmcgarvey (approved):

PR #1467, opened 2021-04-09
Support previous versions of cookie encryption

### LINKED ISSUE #1466: Sessions invalid after upgrading from 0.26 to 0.27
**Describe the bug**

https://github.com/luckyframework/lucky/commit/cf5b125a63d726776ed81de682d89127610374bd#diff-c743e0f7ad2dbadc7113e95dc52c0327f37ec275cc703e37a7f32266fdb20f16R3

This change swapped from using `Base64.encode` to `Base64.strict_encode`. The value is exactly the same except `Base64.encode` appends a `\n`. That means that apps that have user's with the old session token will not be recognized as valid and it produces a 500 error.

**Expected behavior**

If we actually want to use `Base64.strict_encode` going forward. I think we need to allow the legacy prefix for this release cycle and make sure that the session token uses the new value so everyone migrates gradually and we can remove the legacy value in the future.

I believe **NO ONE** should update and deploy existing apps using the `0.27.0` release. New apps will not be affected

matthewmcgarvey:
## Purpose

Fixes #1466 

## Description

In previous releases, cookies were www form encoded/decoded and our prefix was slightly different. This change supports the old way to allow apps to gradually migrate their users. We will read the old way but always write with the new way so it should be safe to remove for the next release.

## Checklist

### DIFF
--- a/src/lucky/cookies/cookie_jar.cr
+++ b/src/lucky/cookies/cookie_jar.cr
@@ -1,6 +1,7 @@
 class Lucky::CookieJar
-  MAX_COOKIE_SIZE         = 4096
-  LUCKY_ENCRYPTION_PREFIX = Base64.strict_encode("lucky") + "--"
+  MAX_COOKIE_SIZE                = 4096
+  LUCKY_ENCRYPTION_PREFIX        = Base64.strict_encode("lucky") + "--"
+  LEGACY_LUCKY_ENCRYPTION_PREFIX = Base64.encode("lucky") + "--"
   alias Key = String | Symbol
   private property cookies
   private property set_cookies
@@ -147,6 +148,8 @@ class Lucky::CookieJar
       base_64_encrypted_part = cookie_value.lchop(LUCKY_ENCRYPTION_PREFIX)
       decoded = Base64.decode(base_64_encrypted_part)
       String.new(encryptor.decrypt(decoded))
+    elsif encrypted_with_legacy?(cookie_value)
+      decrypt(update_from_legacy_value(cookie_value), cookie_name)
     else
       raise <<-ERROR
       It looks like this cookie's value is not encrypted by Lucky. This likely means the cookie was set by something other than Lucky, like a client side script.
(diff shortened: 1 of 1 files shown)


src/lucky/cookies/cookie_jar.cr
```diff
@@ -159,10 +162,22 @@ class Lucky::CookieJar
...
   private def encrypted_with_lucky?(value : String) : Bool
     value.starts_with?(LUCKY_ENCRYPTION_PREFIX)
   end
 
+  # legacy encrypted values had a \n between the encoded lucky and -- and were also www form encoded
+  # this allows apps made before 0.27.0 to not have to log all users out
+  private def encrypted_with_legacy?(value : String) : Bool
+    decoded_value = URI.decode_www_form(value)
+    decoded_value.starts_with?(LEGACY_LUCKY_ENCRYPTION_PREFIX)
+  end
```
jwoertink, line 179: I think this is fine, but what's a good upgrade path we can take so we can drop this by the next release or something? Maybe we can just add a deprecation notice, as well as some info logging? It'll be noisy, but only happens while people are on legacy cookies. Then for the next release, we yank this stuff out.

matthewmcgarvey: I don't think there is a deprecation notice that would be useful for devs. It's not their app that needs to update, it's their users. For instance, if a user doesn't sign in while the app's deployed as 0.27 and we remove this code in 0.28 they'll run into this issue as well if their session cookie still exists. It would also be a problem if devs skip the release with the legacy code

jwoertink: Ok, maybe we keep it until 1.0, but for 1.0 we rip it out. It'll be a breaking change, but it'll be up to the developers to have figured out on their end how to get their users to clear the cookies or whatever. I think we should definitely at least log a warning letting people know.

matthewmcgarvey: To be clear, this change will fix users going forward. This reads them in the old way but writes cookies the new way so all users will be fixed going forward. The only worry is apps skipping this legacy handling or users somehow keeping their cookies and not logging on while the legacy version is deployed.

Neither of those seem like warnings we can give the devs, do they?

robacarp: What's the failure mode when this code doesn't exist?

matthewmcgarvey: Tokens encrypted the old way cause this error https://github.com/luckyframework/lucky/blob/9474f6d1d996dc73dab813a1e1af12edd97d5afb/src/lucky/cookies/cookie_jar.cr#L151-L158

jwoertink (approved):
Ok, we can at least get this patched in. If we need more changes, those can be added later. 👍

PR #1477, opened 2021-04-20
Implement `mount_with_defaults`

### LINKED ISSUE #1152: Implement `mount_defaults`
As described here: https://github.com/luckyframework/lucky/issues/1049#issuecomment-611689438

The idea is that common `needs` can be passed in automatically. Stuff like `context`, `current_user`

stephendolan:
## Purpose
Fixes #1152 by adding a new `mount_with_defaults` method that can inject some helpful defaults into components.

## Description
I'd like to get some feedback from others before I go too far down any given path. That said, I've got the thing that was blocking me working. Say we have a component with a `form`, like this:

```crystal
class Filters::Form < BaseComponent
  needs current_user : User
  needs operation : SaveFilter
  needs route : Lucky::RouteHelper
  needs button_text : String

  def render
    form_for route, class: "divide-y divide-gray-200" do
      div class: "px-4 py-5 sm:p-6 space-y-4" do
        mount Filters::FormFields, operation
      end

      div class: "px-4 py-4 sm:px-6" do
        mount UI::Button, &.submit button_text
      end
    end
  end
end
```

This is a really nice abstraction, and greatly cleans up the duplication between `edit_page.cr` and`new_page.cr` for the `Filter` model, in this case.

**I'm not super thrilled about having to manually add `needs context` and `needs current_user` to the component, but also not sure if we can inject that after the component is initialized so that it's available but invisible.**

The use of this component looks something like this:

```crystal
class Filters::NewPage < MainLayout
  needs operation : SaveFilter
  quick_def page_title, "New Filter"

  def content
    div class: "bg-white overflow-hidden shadow rounded-lg divide-y divide-gray-200" do
      div class: "px-4 py-5 sm:px-6" do
        h3 class: "text-lg leading-6 font-medium text-gray-900" do
          text "New filter"
        end
      end

      mount_with_defaults Filters::Form, operation: operation, route: Filters::Create.route, button_text: "Create filter"
    end
  end
end
```

## To-do
* [ ] - Add specs for the new `mount_with_defaults`

## Checklist

### DIFF
--- a/src/lucky/memoizable.cr
+++ b/src/lucky/memoizable.cr
@@ -22,7 +22,7 @@ module Lucky::Memoizable
   macro memoize(method_def)
     {% raise "You must define a return type for memoized methods" if method_def.return_type.is_a?(Nop) %}
     {%
-      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop) # ameba:disable Style/IsAFilter
+      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop)
     %}
 
     @__memoized_{{method_def.name}} : Tuple(
--- a/src/lucky/mount_component.cr
+++ b/src/lucky/mount_component.cr
@@ -82,7 +82,7 @@ module Lucky::MountComponent
 
            ▸ mount MyComponent
            ▸ mount_instance MyComponent.new
-           ▸ mount_with_defaults MyComponent.new
+           ▸ mount_with_defaults MyComponent
         ERROR
     %}
   end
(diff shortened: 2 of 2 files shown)


src/lucky/memoizable.cr
```diff
@@ -22,7 +22,7 @@ module Lucky::Memoizable
   macro memoize(method_def)
     {% raise "You must define a return type for memoized methods" if method_def.return_type.is_a?(Nop) %}
     {%
-      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop) # ameba:disable Style/IsAFilter
+      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop)
```
stephendolan, line 25: I can break this out into a separate PR after we work through the rest of the stuff, but I wanted to get the CheckFormat specs passing while we're tinkering.

stephendolan:
I think I misunderstood what Paul/Jeremy were thinking about in the issue linked here, but I'm still happy to take this in the right direction with some guidance!

stephendolan:
This was instructive for me to poke around with, but I don't think it's a viable solution. Closing for now.

PR #1478, opened 2021-04-20
Nightly Crystal and edge shard testing in CI

stephendolan:
## Purpose
Test against the `master` branch of all dependencies, as well as the `nightly` Crystal version.

## Description
~~No idea if this is actually going to work like I think it will.~~

It worked!

This adds one additional matrix variable, `shard_file`, which is used to optionally install from alternate sources. The only one I've added to source control as a part of this PR is `shard.edge.yml`. The idea is that we should always keep every shard in that file on the latest version (`branch: master/main`).

I've added two additional `experimental`-flagged builds:

- Test the `shard.edge.yml` content with Crystal 1.0
- Test the `nightly` Crystal build against our standard `shard.yml` content

Because we leverage `continue-on-error` in our builds based on the `experimental` flag, it means that we could still push PRs through even if one of the above scenarios fails. That said, it's probably best to tackle these issues as they come up.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Lucky CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
     branches: "*"
 
@@ -11,18 +11,29 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        shard_file:
+          - shard.yml
         crystal_version:
           - 0.36.1
           - 1.0.0
         experimental:
           - false
+        include:
+          - shard_file: shard.edge.yml
+            crystal_version: 1.0.0
+            experimental: true
+          - shard_file: shard.yml
+            crystal_version: nightly
+            experimental: true
     runs-on: ubuntu-latest
     continue-on-error: ${{ matrix.experimental }}
     container: crystallang/crystal:${{ matrix.crystal_version }}-alpine
+    env:
+      SKIP_LUCKY_TASK_PRECOMPILATION: "1"
     steps:
       - uses: actions/checkout@v1
       - name: Install shards
-        run: shards install
+        run: SHARDS_OVERRIDE=${{ matrix.shard_file }} shards install
       - name: Format
         run: crystal tool format --check
       - name: Lint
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -11,18 +11,29 @@ jobs:
...
+            crystal_version: 1.0.0
+            experimental: true
+          - shard_file: shard.yml
+            crystal_version: nightly
+            experimental: true
     runs-on: ubuntu-latest
     continue-on-error: ${{ matrix.experimental }}
     container: crystallang/crystal:${{ matrix.crystal_version }}-alpine
+    env:
+      SKIP_LUCKY_TASK_PRECOMPILATION: "1"
```
stephendolan, line 32: I'm honestly not sure if this is a Crystal bug or not. When any `postinstall` tasks would run (like for Avram), `SHARDS_OVERRIDE` would **also** impact that shard's environment, and Crystal would complain about `Avram` not having a `shard.edge.yml`.

I don't think we really need to precompile the tasks, though, so this seems like a nice way to speed up the test suite anyways.

src/lucky/memoizable.cr
```diff
@@ -22,7 +22,7 @@ module Lucky::Memoizable
   macro memoize(method_def)
     {% raise "You must define a return type for memoized methods" if method_def.return_type.is_a?(Nop) %}
     {%
-      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop) # ameba:disable Style/IsAFilter
+      raise "All arguments must have an explicit type for memoized methods" if method_def.args.any? &.is_a?(Nop)
```
stephendolan, line 25: Had to add this to get the specs green. Apparently Ameba updated since our last PR was merged.

.github/workflows/ci.yml
```diff
@@ -11,18 +11,29 @@ jobs:
...
       matrix:
+        shard_file:
+          - shard.yml
         crystal_version:
           - 0.36.1
           - 1.0.0
         experimental:
           - false
+        include:
+          - shard_file: shard.edge.yml
```
jwoertink, line 22: Is this a new file Crystal uses?

.github/workflows/ci.yml
```diff
@@ -31,22 +42,33 @@ jobs:
...
+            crystal_version: 1.0.0
+            experimental: true
```
jwoertink, line 55: why is 1.0.0 experimental? Wouldn't this be tested from line 49 above?

src/lucky/memoizable.cr
jwoertink: Sweet! Yeah, I had filed a bug. Looks like it was updated 👍

.github/workflows/ci.yml
stephendolan: No, I just didn't want to create a `shard.override.yml` since that would *always* be taken by default. I named it according to how Brian C. recommended here: https://crystal-lang.org/2020/08/20/preparing-our-shards-for-crystal-1.0.html

jwoertink (approved):
I dig it. Feel free to merge when ready

.github/workflows/ci.yml
stephendolan: It's experimental because it's using the `shard.edge.yml`! Every `includes` just has to have every matrix value, and I figured we'd want to test the all-`master` shard file against the latest Crystal version we support.

stephendolan:
I'm going to :shipit:, and we'll see how it works on a few PRs!

PR #1483, opened 2021-04-24
Extract Assignable code to single module and refactor

matthewmcgarvey:
## Purpose

Related to https://github.com/luckyframework/lucky/pull/1480

Wanting to make sure I fully understood how our `needs` setup works, I cleaned up some of the macro code and made it completely detached from the html stuff so anyone can include the module in their code if they want to use `needs` in other areas.

## Description

- Moved everything related to `needs` and the generated initialize methods into the `Lucky::Assignable` module
- Remove old compilation error about needs ending with question marks (year old deprecation seemed reasonable to remove)
- changed getters to actually use the `getter/`getter?` macros
- Updated `macro included` to delegate to two sup methods to setup needs and initialize
- removed `SETTINGS` from `inherit_page_settings` (found no usage)
- `Lucky::HTMLPage` no longer includes `Lucky::HTMLBuilder` in a `macro included` since `Assignable` handles when it's run in a module (there was also a problem with the amount of times `Assignable` could be included that was fixed, it could only be included twice I think and then there would be a bug but I didn't actually verify)
- Removing the `macro included` from `HTMLPage` made the habitat stuff fail in `FormHelpers` which was fixed

## Checklist

### DIFF
--- a/spec/lucky/assignable_spec.cr
+++ b/spec/lucky/assignable_spec.cr
@@ -81,6 +81,16 @@ class OverrideGetterPage
   end
 end
 
+class NonPageClass
+  include Lucky::Assignable
+
+  needs param : String
+end
+
+class InheritedNonPageClass < NonPageClass
+  needs other_param : String
+end
+
 describe "Assigns within multiple pages with the same name" do
   it "should only appear once in the initializer" do
     PageOne.new build_context, title: "foo", name: "Paul", second: "second"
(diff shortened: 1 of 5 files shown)


matthewmcgarvey:
If I want to provide an "escape hatch" to opt out of the `needs` stuff I would have to remove including `Assignable` in `HTMLBuilder` and move it into the pages and components directly. It would be in lucky_cli and added into the base classes.

jwoertink (approved):
I like the refactor. Looks like we gain some additional functionality too! 👍

PR #1488, opened 2021-05-06
Pass context to components through mount

matthewmcgarvey:
## Purpose

Related to https://github.com/luckyframework/lucky/issues/1152

## Description

Instead of implementing `mount_defaults` the main code that we want passed down into components without being asked for is the context. This adds it to be passed in just like view is.

### DIFF
--- a/src/lucky/base_component.cr
+++ b/src/lucky/base_component.cr
@@ -21,6 +21,15 @@ abstract class Lucky::BaseComponent
     self
   end
 
+  private def context : HTTP::Server::Context
+    @context || raise "No context was set in #{self.class.name}. Use 'mount' or set it with 'context(@context)' before rendering."
+  end
+
+  def context(@context : HTTP::Server::Context?)
+    # This is used by Lucky::MountComponent to set the context.
+    self
+  end
+
   def render_to_string : String
     String.build do |io|
       view(io)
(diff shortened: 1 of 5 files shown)


stephendolan:
@matthewmcgarvey This is awesome, and if we can extend `@context` to hold more than just `HTTP::Server::Context` (like `current_site` or `current_user`), then I think the need for `mount_defaults` goes away entirely!

I'm having trouble visualizing when I'd hit the scenario you describe here:
> or set it with 'context(@context)' before rendering.

What would that look like in an app to render a component without `mount`?

jwoertink (approved):
That's funny, I had the same thought just the other day! Really, this is the big one. Anything else like `current_user` or `current_site` can still be as needed. This was really the only annoying one. I dig it!

stephendolan (approved):

matthewmcgarvey:
On my phone so I can't link to the specific lines but at the bottom of this file where components can be provided also needs to pass in the context https://www.github.com/luckyframework/lucky/tree/master/src%2Flucky%2Frenderable.cr

stephendolan:
Ah, right!

jwoertink (approved):
Yup. Looks good to me

stephendolan:
Likewise! I don't see any downsides to getting this merged.

PR #1489, opened 2021-05-06
Remove deprecated m for mounting components

matthewmcgarvey:
## Purpose

The deprecation notice was added way back in [0.24.0](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md#v0240-2020-09-05) so it seems like time to remove it.


jwoertink (approved):
That was a fun experiment 😅

PR #1492, opened 2021-05-12
Add additional JSON render overload for string bodies

mjeffrey18:
## Purpose
Some JSON serializer shards output raw JSON-compatible strings, which means we need to use `JSON.parse(*)` before passing the payload into the `json` render method.

This PR adds support for JSON strings being directly passed.

### Before

```crystal
get "/foo" do
  data = Serializer.new.serialize(scope)
  json(JSON.parse(data))
end
```

### After

```crystal
get "/foo" do
  data = Serializer.new.serialize(scope)
  json(data)
end
```

## Checklist

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -257,7 +257,7 @@ module Lucky::Renderable
   end
 
   # :nodoc:
-  def json(raw_string : String, status : Int32? = nil) : Lucky::TextResponse
+  def json(body : String, status : Int32? = nil) : Lucky::TextResponse
     {%
       raise <<-ERROR
 
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -248,8 +248,27 @@ module Lucky::Renderable
   end
 
   # allows json-compatible string to be returned directly
-  def json(body : String, status : Int32? = nil) : Lucky::TextResponse
+  def raw_json(body : String, status : Int32? = nil) : Lucky::TextResponse
     send_text_response(body, "application/json", status)
+  end
+
+  def raw_json(body : String, status : HTTP::Status) : Lucky::TextResponse
+    raw_json(body, status: status.value)
+  end
+
+  # :nodoc:
+  def json(raw_string : String, status : Int32? = nil) : Lucky::TextResponse
+    {%
+      raise <<-ERROR
+
+      Looks like your trying to pass a string to json response.
+
+      Use `raw_json(body, ...)` instead.
+
+      NOTE: `raw_json` doesn't validate JSON string validity/integrity, use at your own risk.
+
+      ERROR
+    %}
   end
 
   def json(body, status : Int32? = nil) : Lucky::TextResponse
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -247,6 +247,11 @@ module Lucky::Renderable
     head(status.value)
   end
 
+  # allows json-compatible string to be returned directly
+  def json(body : String, status : Int32? = nil) : Lucky::TextResponse
+    send_text_response(body, "application/json", status)
+  end
+
   def json(body, status : Int32? = nil) : Lucky::TextResponse
     send_text_response(body.to_json, "application/json", status)
   end
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -31,6 +31,12 @@ class Rendering::JSON::Index < TestAction
   end
 end
 
+class Rendering::JSON::WithStringBody < TestAction
+  get "/foo1" do
+    json("{\"name\":\"Paul\"}")
+  end
+end
+
 class Rendering::JSON::WithStatus < TestAction
   get "/foo" do
     json({name: "Paul"}, status: 201)
@@ -200,6 +206,10 @@ describe Lucky::Action do
 
   it "renders JSON" do
     response = Rendering::JSON::Index.new(build_context, params).call
+    response.body.to_s.should eq %({"name":"Paul"})
+    response.status.should eq 200
+
+    response = Rendering::JSON::WithStringBody.new(build_context, params).call
     response.body.to_s.should eq %({"name":"Paul"})
     response.status.should eq 200
 
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -31,14 +31,20 @@ class Rendering::JSON::Index < TestAction
   end
 end
 
-class Rendering::JSON::WithStringBody < TestAction
-  get "/foo1" do
-    json("{\"name\":\"Paul\"}")
+class Rendering::JSON::WithRawStringBody < TestAction
+  get "/foo" do
+    raw_json("{\"name\":\"Paul\"}")
+  end
+end
+
+class Rendering::JSON::WithRawStringBodyWithStatus < TestAction
+  get "/bar" do
+    raw_json("{\"name\":\"Paul\"}", status: 201)
   end
 end
 
 class Rendering::JSON::WithStatus < TestAction
-  get "/foo" do
+  get "/foo1" do
     json({name: "Paul"}, status: 201)
   end
 end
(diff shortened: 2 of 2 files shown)


src/lucky/renderable.cr
```diff
@@ -247,6 +247,11 @@ module Lucky::Renderable
     head(status.value)
   end
 
+  # allows json-compatible string to be returned directly
+  def json(body : String, status : Int32? = nil) : Lucky::TextResponse
+    send_text_response(body, "application/json", status)
```
stephendolan, line 252: Should we still `JSON.parse` the body, here?

My only concern would be that this allows me to do `json "HAHA THIS ISN'T JSON"`, and now my app is silently returning invalid content for the content type `application/json`.

jwoertink:
Hey @mjeffrey18 thanks for the PR! 

This is a tough one. I know it's been a pain point (even in my apps). It totally makes sense that you may already have a json formatted string, so rendering it becomes a bit wonky.

My only concern, along with what @stephendolan mentioned, is that Lucky's top goal is to be type-safe which goes just a bit beyond "is it the same type?". Doing this would allow people to pass in a malformed JSON string, so there's no safety to ensure what you're passing in is valid JSON.

Now, calling `JSON.parse(body).to_json` would be the safety we want, but that comes at a pretty big performance hit ([see discussion](https://github.com/luckyframework/lucky/discussions/1442)).

Alright, just thinking out loud here, and this might sound wild, but what if we made a habitat config setting that's enabled in development by default. This setting would then say if it's enabled, then parse the json strings. This would then catch malformed json in development, but in production it would just pass in the raw strings with the assumption that you've already done your checks?

What are your thoughts?

jwoertink:
Actually... looking in to this, I now see that XML has the same issue https://github.com/luckyframework/lucky/blob/master/src/lucky/renderable.cr#L258-L260 😝  bleh... 

@stephendolan maybe we just merge this in for now to get "something", and work out a new config for better safety on both of these in a separate PR?

stephendolan:
Yeah, I don't love the idea that a Lucky app could silently send invalid JSON/XML as a response, but if we already have a similar API in place for XML then I think that bringing the JSON implementation to parity makes sense.

I almost wonder if we should rename the arguments to "json_content" and "string_content" or something to make it extra clear... haha I can already envision the Discord messages saying "My Vue app is crashing because the API is sending invalid JSON". Maybe that's something we open up as a separate follow-up issue to get Matthew's thoughts on as well, though.

jwoertink:
Well, I guess the xml probably makes sense because we don't call `{a: 1}.to_xml`. Your only option at that point is to generate the XML first. For the `json`, we don't know if you're passing an array, hash, named tuple, or JSON::Any object.

A better case would probably be making an empty module that is included in the BaseSerializer, then saying that the argument needs to either be JSON::Any or of that included module... Then `json(some_string)` would be a compile-time error to say use `send_text_response` as the escape hatch when you want to control passing a raw json string...

mjeffrey18:
Hey @stephendolan and @jwoertink, thanks for the quick feedback and warm welcome! 

As you mentioned, I really want to avoid `JSON.parse`due to the performance hit.
On the other hand, I completely understand the motive to keep Lucky type-safe.

I like the idea of having safety in development vs production environment.
Having the method named `json_conent`/`json_string`/`json_raw`etc also works. Just curious how you guys normally approach this with an existing XML method already supporting raw strings? Deprecation warnings or dual legacy support? I guess whatever naming we use for JSON can be a convention for XML versions also.

I can probably work on one of these options early week?

jwoertink:
I think the xml was basically like "meh... who uses xml?" 😂  I haven't needed to return xml in Lucky yet, and last time I did in rails was probably 10 years ago lol. We should for sure find a solution to that, but I think figuring out json is higher priority.

My original idea was something like:

```crystal
# config/some_setup_thing.cr
WhateverWeCallThis.configure do |settings|
  settings.ensure_valid_json_response = Lucky::Env.development?
end

def json(raw_string : String, ...)
  ensure_valid_json_response!(raw_string)
  send_text_response(raw_string, ...)
end

private def ensure_valid_response(raw_string : String)
  if WhateverWeCallThis.settings.ensure_valid_json_response
    JSON.parse(raw_string)
  end
end
```

We wouldn't need to convert back to json here because if the parser doesn't raise an exception, then `raw_string` is valid, so just pass that in. Then in production, it would skip that check. 

I'd be cool with going this route for now, then working out a plan that could be used for the XML stuff and a bit more safety with json later. I'm horrible with naming, but if you have any ideas, or see major drawbacks to this approach, I'm open to suggestions.

matthewmcgarvey:
We already have a method that allows passing json as a raw string, the only caveat is that you have to specify the content type. https://github.com/luckyframework/lucky/blob/c778254b9394ad93183c1e9d9c5b551809afd5b4/src/lucky/renderable.cr#L210

I don't think that's been brought up yet and I have used it before.

stephendolan:
Oh, yeah that seems ideal because it makes the intention explicit.

I'm not sure how to make that more "discoverable", but it does seem to satisfy the need.

jwoertink:
Using that `send_text_response` is what I've been using. This is basically the escape hatch. Maybe it's really just as simple as saying "Use the Lucky Serializers", but then if you can't / don't want to, then use the escape hatch. In this case, what we do is something like this:

```crystal
def json(raw_string : String, ...)
  {% clippy "Hey, I see you're trying to pass a raw json string.. Use `send_text_response` instead" %}
end
```

Then in that case it at least gives you an example of how to use the escape hatch.

jwoertink:
Ok, been thinking about this, and there's a few things I think we should do.

1. [Update body](https://github.com/luckyframework/lucky/blob/master/src/lucky/renderable.cr#L250) to be `body : Lucky::Serializer`.
2. Create `json` overload that takes `body : String`, and raises a compile time error explaining how to use the escape hatch
3. Create a new escape hatch method called `raw_json` that takes a `String`. This way it's easier to remember, and a bit more clear what you're passing in.
4. Optionally we create a `RawSerializer` that takes a NamedTuple or Enumerable and passes it in to `render`. Then instead of doing `json {some: "data"}`, you'd do `json RawSerializer.new({some: "data"})`. It's a little more verbose, but maybe will help focus where you structure your data at?

@mjeffrey18 What are your thoughts on going this route? 
/cc @matthewmcgarvey @stephendolan

matthewmcgarvey:
@jwoertink

I think we can leave body untyped, but add an overload for String specifically with that compile-time error.
How about `Lucky::RawSerializer` takes in a string so that we don't need a separate method?

mjeffrey18:
Hey everyone, thanks for the feedback and suggestions.

I've made the updates as per a few of the suggestions.
I figured the `Lucky::Serializer` with `json(body, ...)` route is what most folk will take, it's easy and safe plus well documented.
However, if like me you decided to move forward with your own serializers, you'll get that nice warning to use `raw_json(body: String, ...)` at your own risk.

What's everyone thoughts? 

@jwoertink @stephendolan @matthewmcgarvey

stephendolan:
I'm a fan of the approach we landed with!

jwoertink (approved):
Yup 👍  this is much nicer indeed. Thanks for putting this together @mjeffrey18

src/lucky/renderable.cr
```diff
@@ -247,6 +247,30 @@ module Lucky::Renderable
...
+  def raw_json(body : String, status : Int32? = nil) : Lucky::TextResponse
+    send_text_response(body, "application/json", status)
+  end
+
+  def raw_json(body : String, status : HTTP::Status) : Lucky::TextResponse
+    raw_json(body, status: status.value)
+  end
+
+  # :nodoc:
+  def json(raw_string : String, status : Int32? = nil) : Lucky::TextResponse
```
matthewmcgarvey, line 260:
```suggestion
  def json(body : String, status : Int32? = nil) : Lucky::TextResponse
```
Just to make sure using named arguments doesn't get past this overload

matthewmcgarvey (changes requested):

mjeffrey18:
Thanks, guys!  
Just created my first shard if anyone is interested in a JSON-API serializer - https://github.com/mjeffrey18/fast-jsonapi-serializer
Looking forward to more contributions to Lucky in the future.

matthewmcgarvey (approved):

PR #1495, opened 2021-05-14
Added 'multipart?' request helper

### LINKED ISSUE #1494: Add new multipart? request helper
We have a list of these helper methods that can be used in your Actions like `html?` or `ajax?`. While working on Breeze I came up with a use case for `multipart?` which should return true if the `"Content-Type"` header is `multipart/form-data;`. 

This could be used in before action pipes to do specific things if you're doing a multipart POST.

https://github.com/luckyframework/lucky/blob/master/src/lucky/request_type_helpers.cr

nodeMD:
## Purpose
Fixes #1494 

## Description
Adds 'multipart?' request helper

## Checklist

### DIFF
--- a/spec/lucky/request_type_helper_spec.cr
+++ b/spec/lucky/request_type_helper_spec.cr
@@ -66,6 +66,14 @@ describe Lucky::RequestTypeHelpers do
     action.context.request.headers["X-Requested-With"] = "XMLHttpRequest"
     action.ajax?.should(be_true)
   end
+
+  it "checks if request is multipart" do
+    action = FakeAction.new
+    action.multipart?.should(be_false)
+
+    action.context.request.headers["Content-Type"] = "multipart/form-data"
+    action.multipart?.should(be_true)
+  end
 end
 
 private def override_format(format : Symbol?)
--- a/spec/lucky/request_type_helper_spec.cr
+++ b/spec/lucky/request_type_helper_spec.cr
@@ -71,7 +71,7 @@ describe Lucky::RequestTypeHelpers do
     action = FakeAction.new
     action.multipart?.should(be_false)
 
-    action.context.request.headers["Content-Type"] = "multipart/form-data"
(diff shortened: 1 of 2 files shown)


src/lucky/request_type_helpers.cr
```diff
@@ -91,6 +91,13 @@ module Lucky::RequestTypeHelpers
     request.headers["X-Requested-With"]?.try(&.downcase) == "xmlhttprequest"
   end
 
+  # Check if the request is multipart
+  #
+  # This tests if the `Content-Type` header is `multipart/form-data`
+  def multipart? : Bool
+    request.headers["Content-Type"]?.try(&.downcase) == "multipart/form-data"
```
jwoertink, line 98: I believe that most content-types won't be just `multipart/form-data`. Generally these will look more like

```
multipart/form-data; boundary=------boundary123
```

What I had to do in a recent commit was use `starts_with?` https://github.com/luckyframework/breeze/blob/master/src/breeze/actions/mixins/action_helpers.cr#L76-L78

jwoertink (changes requested):
This looks like another great contribution! Just one small change to make sure we can catch a wider range.

src/lucky/request_type_helpers.cr
nodeMD: Thanks 👍 I read a little more about it and I see now how multipart is working 🎉

spec/lucky/request_type_helper_spec.cr
```diff
@@ -71,7 +71,7 @@ describe Lucky::RequestTypeHelpers do
     action = FakeAction.new
     action.multipart?.should(be_false)
 
-    action.context.request.headers["Content-Type"] = "multipart/form-data"
+    action.context.request.headers["Content-Type"] = "multipart/form-data; boundary="
```
jwoertink, line 74: ah nice! 👍

jwoertink (approved):
Great work! Thanks for taking care of this.

PR #1496, opened 2021-05-17
Refactor/speed up BaseTags methods

matthewmcgarvey:
Considering this [TechEmpower Framework Benchmark](https://www.techempower.com/benchmarks/#section=test&runid=e81c66ad-1e37-4f2c-9e25-bcc79796fab8&hw=ph&test=fortune&l=zik0yn-e7&a=2) I began looking into our html dsl since the only big difference between Lucky and the other frameworks is that we use a Crystal DSL while they use ECR.
CleanShot 2021-05-17 at 00 14 57@2x

I created a benchmark with this code in it:
```crystal
require "benchmark"
require "./src/lucky"

class MyPage
  include Lucky::HTMLPage

  needs name : String
  needs list : Array(String)

  def render : String
    h1 name
    h2 "WOAH"
    ul do
      list.each do |item|
        li item
      end
    end
    view.to_s
  end
end

request = HTTP::Request.new("GET", "/")
response = HTTP::Server::Response.new(IO::Memory.new)
context = HTTP::Server::Context.new(request, response)
Benchmark.ips do |x|
  x.report("html rendering") { MyPage.new(context, name: "Johnny", list: ["dog", "cat", "turtle", "hamster"]).render }
end
```

With that benchmark in hand, I made tweaks and ended up with these changes.

## Description of Changes

The main two changes made were
1. Don't call a block override if content is passed
2. avoid parsing attrs/options if possible

## Performance Outcomes (using the above benchmark)

Before: 410.54k (average)
After: 741.85k (average)

**That's an 80% improvement**

There is a major caveat though that this speed up is only in certain conditions so users will most likely see a lesser improvement.

### DIFF
--- a/src/lucky/tags/base_tags.cr
+++ b/src/lucky/tags/base_tags.cr
@@ -119,10 +119,14 @@ module Lucky::BaseTags
         attrs : Array(Symbol) = [] of Symbol,
         **other_options
       ) : Nil
+      boolean_attrs = build_boolean_attrs(attrs)
       merged_options = merge_options(other_options, options)
-      {{method_name.id}}(attrs, merged_options) do
-        text content
-      end
+      tag_attrs = build_tag_attrs(merged_options)
+      view << "<{{tag.id}}" << tag_attrs << boolean_attrs << ">" << HTML.escape(content.to_s) << "</{{tag.id}}>"
+    end
+
+    def {{method_name.id}}(content : Lucky::AllowedInTags | String) : Nil
+      view << "<{{tag.id}}>" << HTML.escape(content.to_s) << "</{{tag.id}}>"
     end
 
     def {{method_name.id}}(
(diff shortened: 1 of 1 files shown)


src/lucky/tags/base_tags.cr
```diff
@@ -119,10 +119,14 @@ module Lucky::BaseTags
         attrs : Array(Symbol) = [] of Symbol,
         **other_options
       ) : Nil
+      boolean_attrs = build_boolean_attrs(attrs)
       merged_options = merge_options(other_options, options)
-      {{method_name.id}}(attrs, merged_options) do
-        text content
-      end
+      tag_attrs = build_tag_attrs(merged_options)
+      view << "<{{tag.id}}" << tag_attrs << boolean_attrs << ">" << HTML.escape(content.to_s) << "</{{tag.id}}>"
```
robacarp, line 125: This doesn't seem like it'd be a regression, but are the tag_attrs being escaped properly? For example, an `&` in the middle of an `a href`?

jwoertink (approved):
Boom! Nice find

PR #1497, opened 2021-05-17
added DeleteOperation to delete action

### LINKED ISSUE #1487: The generated delete action should use a DeleteOperation
When you generate a new browser resource, you'll get this action

https://github.com/luckyframework/lucky/blob/master/tasks/gen/templates/resource/actions/%7B%7Bfolder_name%7D%7D/delete.cr.ecr

It should use the `DeleteOperation` by default.

nodeMD:
## Purpose
Fixes #1487

## Description
added DeleteOperation to delete action
@jwoertink I am pretty sure that it is not as easy as that change that I've added but I would like to work on it so if it is not enough then could you please give me more guidance and direct me where I can find information to fulfil this issue? 🙂 

## Checklist

### DIFF
--- a/spec/tasks/gen/resource_spec.cr
+++ b/spec/tasks/gen/resource_spec.cr
@@ -16,7 +16,7 @@ describe Gen::Resource::Browser do
           "./src/actions/users/create.cr": "class Users::Create < BrowserAction",
           "./src/actions/users/edit.cr": "class Users::Edit < BrowserAction",
           "./src/actions/users/update.cr": "class Users::Update < BrowserAction",
-          "./src/actions/users/delete.cr": "class Users::Delete < BrowserAction"
+          "./src/actions/users/delete.cr": "class Users::DeleteOperation < BrowserAction"
         should_create_files_with_contents io,
           "./src/actions/users/index.cr": %(get "/users"),
           "./src/actions/users/show.cr": %(get "/users/:user_id"),
--- a/spec/tasks/gen/resource_spec.cr
+++ b/spec/tasks/gen/resource_spec.cr
@@ -16,7 +16,7 @@ describe Gen::Resource::Browser do
           "./src/actions/users/create.cr": "class Users::Create < BrowserAction",
           "./src/actions/users/edit.cr": "class Users::Edit < BrowserAction",
           "./src/actions/users/update.cr": "class Users::Update < BrowserAction",
-          "./src/actions/users/delete.cr": "class Users::DeleteOperation < BrowserAction"
+          "./src/actions/users/delete.cr": "class Users::Delete < BrowserAction"
         should_create_files_with_contents io,
           "./src/actions/users/index.cr": %(get "/users"),
           "./src/actions/users/show.cr": %(get "/users/:user_id"),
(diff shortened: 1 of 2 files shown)


spec/tasks/gen/resource_spec.cr
```diff
@@ -16,7 +16,7 @@ describe Gen::Resource::Browser do
           "./src/actions/users/create.cr": "class Users::Create < BrowserAction",
           "./src/actions/users/edit.cr": "class Users::Edit < BrowserAction",
           "./src/actions/users/update.cr": "class Users::Update < BrowserAction",
-          "./src/actions/users/delete.cr": "class Users::Delete < BrowserAction"
+          "./src/actions/users/delete.cr": "class Users::DeleteOperation < BrowserAction"
```
jwoertink, line 19: This change should stay as it was. `Users::Delete` is the correct Action name

tasks/gen/templates/resource/actions/{{folder_name}}/delete.cr.ecr
```diff
@@ -1,4 +1,4 @@
-class <%= pluralized_name %>::Delete < BrowserAction
+class <%= pluralized_name %>::DeleteOperation < BrowserAction
```
jwoertink, line 1: This should be reverted. `class <%= pluralized_name %>::Delete` is the correct action name.

jwoertink (changes requested):
Hey! Thanks for starting another PR. This one will require changing how the action is executed.

Currently when you generate the action, you'll get generated code like this:

```crystal
UserQuery.find(user_id).delete
flash.success = "Deleted the record"
redirect Index
```

But that just calls `delete` directly on the model. Instead it should generate the proper `DeleteOperation` and implement that. It'll end up generating code like this:

```crystal
record = UserQuery.find(user_id)
User::DeleteOperation.destroy(record) do |operation, deleted|
  flash.success = "Deleted the record"
  redirect Index
end
```

You'll just have to figure out the dynamic calls that will generate the proper code, and then how to test that the correct code is generated for the correct model.

nodeMD:
Thank you @jwoertink! 💪 Really appreciate the guidance, it helps me understand how crystal and lucky is working  🙇 
I've pushed the changes. Please take a look and review 🙂

tasks/gen/templates/resource/actions/{{folder_name}}/delete.cr.ecr
```diff
@@ -1,7 +1,9 @@
 class <%= pluralized_name %>::Delete < BrowserAction
   <%= route("Delete") %> do
-    <%= query_class %>.find(<%= resource_id_method_name %>).delete
-    flash.success = "Deleted the record"
-    redirect Index
+    <%= underscored_resource %>  = <%= query_class %>.find(<%= resource_id_method_name %>)
+    <%= resource %>::DeleteOperation.destroy(<%= underscored_resource %>) do |operation, deleted|
```
jwoertink, line 4:
```suggestion
    <%= resource %>::DeleteOperation.destroy(<%= underscored_resource %>) do |_operation, _deleted|
```

Since these two variables aren't used in the block, adding the `_` in front will keep [Ameba](https://github.com/crystal-ameba/ameba) from throwing any errors.

jwoertink (changes requested):
Great work! Just one small change left that I just thought of, but the rest of it looks good. Thanks for doing this. I also noticed a few other things that I will open up some other issues on shortly 😄

nodeMD:
> Thanks for doing this. I also noticed a few other things that I will open up some other issues on shortly 😄

Thank you! 😃 Can't wait to give them a try 😁

jwoertink (approved):
🎉

PR #1507, opened 2021-05-28
Ability to render HTML pages with a non-200 status code

### LINKED ISSUE #1312: Ability to render status codes on an html response
**Describe the bug**
In a LuckyAction, I can do `plain_text 'ohai', status: 419` but I can't do `html Page, things, status: 419`

**Expected behavior**
The ideal DSL here isn't clear, and I don't know what a sane implementation would be. Here are some ideas:

```crystal
# could get lost among assignments
html Page, user: current_user, status: 419

# separate statement is verbose
status 419
html Page, user: current_user
```

jwoertink:
## Purpose
Fixes #1312

## Description
This PR adds a new `html_with_status` macro that will let you render an HTML page with a status that's not 200.

```crystal
class SomeAction::Index < BrowserAction

  get "/page" do
    html_with_status IndexPage, 244, some_arg: "hi"
  end
end
```

If you don't pass a number, you'll get an error message that looks like this:

Screen Shot 2021-05-28 at 11 15 23 AM

/cc. @robacarp 

## Checklist

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -19,6 +19,12 @@ class Rendering::Index < TestAction
   end
 end
 
+class Rendering::Show < TestAction
+  get "/rendering/:nothing" do
+    html_with_status IndexPage, 419, title: "Closing Time", arg2: "You don't have to go home but you can't stay here"
+  end
+end
+
 class Namespaced::Rendering::Index < TestAction
   route do
     html ::Rendering::IndexPage, title: "Anything", arg2: "testing multiple args"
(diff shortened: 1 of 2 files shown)


robacarp:
I can't object to this, but now that you've added the parameter to the `html` macro, it aligns better with the other rendering API methods. So why would I ever need to use `html_with_status`? Why not just call `html SomePage, 402`?

jwoertink:
You could still do `html SomePage, 402` if you wanted, but that `402` at a first glance might not make sense. Especially if you used some non-standard code like [LinkedIn](https://stackoverflow.com/questions/46214017/public-linkedin-profile-url-returns-server-status-code-999). I think `html_with_status SomePage, 999` at least gives you a clue as to what the `999` is referring to.

I'm using the `http_with_status` also as the escape hatch to keep with the "Lucky way" of saying "Here's how you do things normally, and here's the escape hatch for when normal doesn't work".

robacarp (approved):

matthewmcgarvey (approved):

PR #1508, opened 2021-05-28
Added new module to disable FLoC

### LINKED ISSUE #1486: Disable FLoC by default
About Google's FLoC: https://www.howtogeek.com/724441/what-is-googles-floc-and-how-will-it-track-you-online/

Example of how Laravel disabled it: https://github.com/laravel/framework/commit/c42988a551ffdae0e70aa15959969732265789c0

We probably want to do something similar to Laravel and set the appropriate header if it's missing. That way users can set it if they want.

jwoertink:
## Purpose
Fixes #1486

## Description
This adds a new module to optionally disable FLoC. Just include it in your app, and it will set the appropriate header.

```crystal
class BrowserAction < Lucky::Action
  include Lucky::SecureHeaders::DisableFLoC
end
```

## Checklist

### DIFF
--- a/spec/lucky/secure_headers_spec.cr
+++ b/spec/lucky/secure_headers_spec.cr
@@ -66,6 +66,14 @@ class SniffGuardRoutes::Index < TestAction
   end
 end
 
+class FLoCGGuardRoutes::Index < TestAction
+  include Lucky::SecureHeaders::DisableFLoC
+
+  get "/secure_path7" do
+    plain_text "test"
+  end
+end
+
 describe Lucky::SecureHeaders do
   describe "SetFrameGuard" do
     it "sets the X-Frame-Options header with sameorigin" do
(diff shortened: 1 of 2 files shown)


robacarp (approved):

matthewmcgarvey:
@jwoertink The connected issue was to disable FLoC by default. What are your thoughts on that?

jwoertink:
Reading around at a few articles, it seems like this is just a google experiment that probably won't last, or may change. I wouldn't be surprised if this is removed by the holidays 😂 

We can still do it by default by adding this module in to generated apps. Then if people want to opt in to this disable, they can just add the module in their BrowserAction for upgrades.

matthewmcgarvey (approved):

edwardloveall (approved):
I think it would be great to disable by default 🙂 It feels like the right thing to do and we want to be "batteries included" where it makes sense. I think it makes sense here!

jwoertink:
ok, issue on CLI opened up! 👍

PR #1509, opened 2021-05-28
Moving params in to HTTP::Server::Context 

### LINKED ISSUE #1473: Parsing Lucky::Params twice per request
**Describe the bug**

We affect performance by parsing the Lucky::Params twice

https://github.com/luckyframework/lucky/blob/79e7108a807ccd8573b5ead3b7bd418051c05c72/src/lucky/http_method_override_handler.cr#L19

https://github.com/luckyframework/lucky/blob/79e7108a807ccd8573b5ead3b7bd418051c05c72/src/lucky/param_helpers.cr#L5

**Expected behavior**

Should this be added to our context extensions? https://github.com/luckyframework/lucky/blob/master/src/lucky/context_extensions.cr

jwoertink:
## Purpose
Fixes #1473

This is my stab at solving #1473 

## Description
The original issue was that we were instantiating the params object twice. This lets us do it just once by moving it in to context.

The only strange thing I'm not too sure on is the whole part of setting `route_params` later. I don't think it affects anything here, but I for sure want another set of eyes on this one.

## Checklist

### DIFF
--- a/src/lucky/action.cr
+++ b/src/lucky/action.cr
@@ -4,6 +4,7 @@ abstract class Lucky::Action
   getter :context, :route_params
 
   def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
+    context.params.route_params = @route_params
   end
 
   abstract def call
--- a/src/lucky/context_extensions.cr
+++ b/src/lucky/context_extensions.cr
@@ -24,4 +24,10 @@ class HTTP::Server::Context
   def flash : Lucky::FlashStore
     @_flash ||= Lucky::FlashStore.from_session(session)
   end
+
+  @_params : Lucky::Params?
+
+  def params : Lucky::Params
+    @_params ||= Lucky::Params.new(request)
+  end
 end
--- a/src/lucky/param_helpers.cr
+++ b/src/lucky/param_helpers.cr
@@ -1,7 +1,5 @@
 module Lucky::ParamHelpers
-  @_params : Lucky::Params?
-
-  def params : Lucky::Params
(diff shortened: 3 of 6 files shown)


src/lucky/action.cr
```diff
@@ -4,6 +4,7 @@ abstract class Lucky::Action
   getter :context, :route_params
 
   def initialize(@context : HTTP::Server::Context, @route_params : Hash(String, String))
+    context.params.route_params = @route_params
```
jwoertink, line 7: It feels weird putting this here, but where else would it go?

src/lucky/param_helpers.cr
```diff
@@ -1,7 +1,5 @@
 module Lucky::ParamHelpers
-  @_params : Lucky::Params?
-
-  def params : Lucky::Params
-    @_params ||= Lucky::Params.new(context.request, @route_params)
+  memoize def params : Lucky::Params
+    context.params
   end
```
jwoertink, line 4: This is basically just so you can still call `params` in your actions without having to call `context.params`.

src/lucky/context_extensions.cr
```diff
@@ -24,4 +24,10 @@ class HTTP::Server::Context
   def flash : Lucky::FlashStore
     @_flash ||= Lucky::FlashStore.from_session(session)
   end
+
+  @_params : Lucky::Params?
+
+  def params : Lucky::Params
```
jwoertink, line 30: Side bonus, `params` is now available in all Components thanks to https://github.com/luckyframework/lucky/pull/1488

matthewmcgarvey (approved):
🎉

PR #1517, opened 2021-05-31
New skip_browsersync_checks flag on lucky watch

### LINKED ISSUE #1255: Alpine Linux support for print_browsersync_port_taken_error
I was experimenting with lucky, within an alpine container. This is for use within vscode remote containers/dev cycle. The `lucky dev` falls over during browsersync phase, within the [tasks/watch cycle](https://github.com/colindensem/lucky/blob/7c90587567b3ea224a2dc6fa02e80a528d7864d9/tasks/watch.cr#L78).  

Alpine ships with a slimmed down version of `ps` and `lsof`.  

For `ps` the `-p` and `command` flags aren't supported, instead you need `ps -o comm`

I think I've got a fix for this, but there are two possible approaches.

1. Alter the code to check for busybox os, run the alternative command (See below)
2. Add docs to detail installing `procps` && `lsof` when using alpine for full `ps` support

Possibly need both for the following code to work. Or simplify and don't try to show a list of processes, just say the port is in use. You'd still need the `lsof` package updating. 

Sudo code for option 1, still work in progress, first lines of crystal ;)
```
    private def browsersync_port_taken_command
      io = IO::Memory.new
      busybox_command = "ps -o comm `lsof -ti :#{BROWSERSYNC_PORT}`"
      non_busybox_command = "ps -p `lsof -ti :#{BROWSERSYNC_PORT}` -o command"
      is_busybox_os? ? busybox_command : non_busybox_command
    end

    private def is_busybox_os?
      io = IO::Memory.new
      Process.run("ls --help 2>&1 | grep BusyBox", output: io, error: STDERR, shell: true)
      !io.to_s.empty?
    end
```

Happy to produce a pr with some guidance as to direction/necessity. 

Background, I was only exploring alpine in case I take it to production, for slimmer images. It works fine out of the box using a `crystallang/crystal:0.35.1` image.
I've pushed an alpine branch on this [repo](https://github.com/colindensem/luckyframework-sandbox/tree/alpine-support).

jwoertink:
## Purpose
Fixes #1255

## Description
On some systems like Alpine Docker, the current browsersync checks may not work. You already have the option to just disable browsersync (removing the `--reload-browser` flag), but in the case where you still want Lucky to handle browsersync, this new `--skip-browsersync-checks` flag will skip checking if the port is available. 

Eventually, it would be nice to have commands that are fully cross-platform compatible as Crystal gets closer to being fully native on Windows. For now, this is the easy hack around it.

## Checklist

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -20,7 +20,7 @@ module LuckySentry
     @app_built : Bool = false
     @successful_compilations : Int32 = 0
 
-    def initialize(build_commands : Array(String), run_commands : Array(String), files : Array(String), @reload_browser : Bool)
+    def initialize(build_commands : Array(String), run_commands : Array(String), files : Array(String), @reload_browser : Bool, @skip_browsersync_checks : Bool)
       @build_commands = build_commands
       @run_commands = run_commands
       @files = files
@@ -63,7 +63,7 @@ module LuckySentry
     end
 
     private def browsersync_port_is_available? : Bool
-      if File.executable?(`which lsof`.chomp)
+      if !@skip_browsersync_checks || File.executable?(`which lsof`.chomp)
         io = IO::Memory.new
         Process.run("lsof -i :#{BROWSERSYNC_PORT}", output: io, error: STDERR, shell: true)
         io.to_s.empty?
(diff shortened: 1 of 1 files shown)


matthewmcgarvey (commented):
I'm hesitant about adding a rather obscure command line flag that disables a small piece of code.

What are your thoughts about documenting that Alpine Linux needs to install `lsof`? We would just document that they need to run `apk add lsof` if they want to run lucky in dev mode

matthewmcgarvey:
For reference, I googled "alpine linux lsof" and the second result is a stack overflow link where the question is about this and the answer was to just install lsof instead of relying on alpine's. https://stackoverflow.com/questions/54487561/can-lsof-show-open-ports-on-alpine-linux

jwoertink:
I'm cool with that 👍  I wasn't too keen on this either, but I don't use Alpine, so I don't know anything about it 🤷‍♂️

PR #1518, opened 2021-06-02
Add new multi_select_input tag

### LINKED ISSUE #777: How to use select_input with 'multiple' attribute
Hi,

First thanks for making this cool framework, I've really enjoyed using it!

My issue is with a many-to-many relationship and using a select field with the `multiple` attribute to assign many associations to one model.

I have the many to many relationship set up fine, the issue is with generating HTML/Form to save the associations.

I could not figure out from the docs how to do this, but this is what I have so far:

```crystal
# In my pages
select_input f.category_ids, name: "article:category_ids[]", multiple: "multiple" do
  options_for_select(f.category_ids, categories_for_select)
end

private def categories_for_select
  @categories.map do |category|
    {category.name, category.id}
  end
end
```

I was thinking that if I set up a virtual field called category_ids, and then parsed this data in an after_save hook, that would be a simple solution. Unfortunately I don't think it's possible

```crystal
# this is what I'm trying to do
class ArticleForm < Article::BaseForm
   after_save sync_categories
  virtual category_ids : Array(Int32)

  def sync_categories(article : Article)
    # do relationship syncing here
  end
end
```

It seems it's not possible to create a virtual field that's an array of integers. 

Is there a way to do this that I've missed in the docs?

jwoertink:
## Purpose
Fixes #777 

## Description
This adds a new select tag method `multi_select_input` which allows you to use Arrays and pass over multiple values with a select.

```crystal
# in your page
def render
  multi_select_input(operation.tags, class: "multi-select") do
    options_for_select(operation.tags, [{"Ruby", "ruby"}, {"Crystal", "crystal"}])
  end
end
```

## Notes

There is a caveat here in that Avram Operations do not support passing in Arrays from params just yet. If you were to use this as is right now, you'd have to update your actions a bit.

```crystal
post "/users" do
  SaveUser.create(params, tags: params.get_all("user:tags")) do |o, u|
      # ....
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/select_helpers_spec.cr
+++ b/spec/lucky/select_helpers_spec.cr
@@ -25,11 +25,6 @@ private class TestPage
     self
   end
 
-  def render_multi_select_options(field, options)
-    options_for_multi_select(field, options)
-    self
-  end
-
   def render_prompt(label)
     select_prompt(label)
     self
@@ -81,7 +76,7 @@ describe Lucky::SelectHelpers do
   end
 
   it "renders multi-select options" do
-    view.render_multi_select_options(form.tags, [{"One", "one"}, {"Two", "two"}, {"Three", "three"}])
+    view.render_options(form.tags, [{"One", "one"}, {"Two", "two"}, {"Three", "three"}])
       .html.to_s.should eq <<-HTML
       <option value="one" selected>One</option><option value="two" selected>Two</option><option value="three">Three</option>
       HTML
--- a/spec/lucky/select_helpers_spec.cr
+++ b/spec/lucky/select_helpers_spec.cr
@@ -14,8 +14,19 @@ private class TestPage
     self
   end
 
+  def render_multi_select(field)
+    multi_select_input field do
+    end
+    self
+  end
+
   def render_options(field, options)
     options_for_select(field, options)
+    self
+  end
+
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (commented):
Does this need a new method? What if we allowed specifying multiple on `select_input` and we rename the method for options that you made just to be an override?

jwoertink:
It doesn't necessarily need a new method. I was more thinking that since multi selects look and work a bit different, we could treat it like a separate tag. It makes it easier to grep for, and stand out in the code:

```crystal
select_input(op.thing, multiple: true, size: 5, class: "select-box") do
end

multi_select_input(op.thing, size: 5, class: "select-box") do
end
```

I do like the idea of just doing an override on the `options_for_select`. I still have a few more things to work out first. I'll think more on the method naming.

stephendolan:
Yeah, for some reason I kind of like the separate methods.

Single and multi selects look and behave differently enough that it feels similar to the separation of email and text input methods we already have.

jwoertink:
nope, not yet. After we figure out #1523 I'll come back to this. I want to still try the `options_for_select` with the method override.

matthewmcgarvey (approved):

PR #1521, opened 2021-06-05
Fix expose method name check

### LINKED ISSUE #1520: Recent commit broke expose
**Describe the bug**

Commit: https://github.com/luckyframework/lucky/commit/bea9b8e61aed9a931cef9a5031f93dee7a06b009

You can call expose two ways with a symbol or just the method name by itself.

[Our tests](https://github.com/luckyframework/lucky/blob/80289da12f4178cd206dba72696c8283cab04825/spec/lucky/expose_spec.cr#L6) only verify the former but using the latter no longer works.

You now get this error:

```bash
❯ crystal spec
Showing last frame. Use --error-trace for full trace.

In spec/lucky/expose_spec.cr:6:10

 6 | expose name
            ^---
Error: undefined macro method 'Call#ends_with?'
```

**To Reproduce**

Update the `expose_spec.cr` expose usage to use the method name instead of a symbol.

matthewmcgarvey:
## Purpose

Fix #1520 

## Checklist

### DIFF
--- a/spec/lucky/expose_spec.cr
+++ b/spec/lucky/expose_spec.cr
@@ -3,7 +3,7 @@ require "../spec_helper"
 include ContextHelper
 
 class OnlyExpose < TestAction
-  expose :name
+  expose name
 
   get "/expose" do
     html OnlyExposePage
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Ah, nice catch. 👍

PR #1523, opened 2021-06-08
Changing how inputs with Array attributes are generated.

jwoertink:
## Purpose
Related #1518 

## Description
We already treat the param names ending with `[]` as an array, but right now, there's no way to tell your input that it's an array without manually updating the name. This PR looks at the field, and if it's an Array type, we can append `[]` to the name. This required a few things to change since these would call [field.param](https://github.com/luckyframework/avram/blob/eb0d092828ad052ff6f3f2defa232198a91fc974/src/avram/attribute.cr#L26) which would just call `value.to_s` if it was nil. In the case of an array, this would make the value of the input `"[]"`.

Now we can keep track of how many times you call the attribute which allows us to set a different value for all of the array values you have.

```crystal
op.tags.each do |tag|
  span tag
  text_input(op.tags, value: tag)
end
span "Add new tag"
text_input(op.tags)
```

## Note

I've tagged this as "BREAKING CHANGE" because this will change how your forms will work when you have arrays. The other major deal is Avram still doesn't handle pulling in values from array params. 

## Checklist

### DIFF
--- a/spec/lucky_avram/input_helpers_spec.cr
+++ b/spec/lucky_avram/input_helpers_spec.cr
@@ -1,0 +1,31 @@
+require "../spec_helper"
+include ContextHelper
+
+private class TestPage
+  include Lucky::HTMLPage
+
+  def render
+  end
+end
+
+describe Lucky::InputHelpers do
+  it "generates the proper name for array attributes" do
+    view { |page|
+      page.text_input(array_attribute)
+      page.text_input(array_attribute)
+      page.text_input(array_attribute)
+    }.should contain <<-HTML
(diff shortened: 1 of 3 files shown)


jwoertink:
The use of the inputs is mostly arbitrary, but you could say that you are letting an admin create new tags on a product, and each tag gets its own text_input. This PR is mainly about passing in an Array attribute, and it generating an input name with `[]` appended on to it.

I could use `checkbox` too, but currently if you had tags = `["footwear", "on sale"]`, and passed that attribute to `checkbox`, you'd get

```html
<input type="checkbox" name="product:tags" value="[\"footwear\", \"on sale\"]">
```

Now with this PR, you can get:

```html
<input type="checkbox" name="product:tags[]" value="footwear">
<input type="checkbox" name="product:tags[]" value="on sale">
```

matthewmcgarvey (approved):

PR #1527, opened 2021-06-10
Working out Array param stuff

jwoertink:
## Purpose
Alternative to #1519

## Description
Going the route of #1519 could still work; however, that route requires that params are defined in separate objects. In larger projects, that could end up being lots of duplication and such.

This first commit takes some of that code used in params, but with the idea that we leave the param parsing in Avram Operations. The plus side to this path is there's no breaking changes. We could support a few additional types with Arrays which would make forms slightly more robust. The downsides are that we leave the avram/lucky coupling, testing params in Avram become a little more difficult as we now have to add more methods to paramable, and this path still doesn't get us closer to supporting model associations (i.e. accepts_nested_attributes_for, etc...)

Maybe this is just a stepping stone of progress towards making some of those downsides doable?

This PR now makes this possible:

```crystal
class Searches::Index < BrowserAction
  param tags : Array(String)

  get "/searches" do
    results = SearchStuff.run!(tags: tags)

    html IndexPage, results: results
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/params_spec.cr
+++ b/spec/lucky/params_spec.cr
@@ -82,11 +82,11 @@ describe Lucky::Params do
 
     params = Lucky::Params.new(request)
 
-    params.nested?(:user)
-    params.nested?(:user)
+    params.nested?(:user).should eq({"name" => "Paul", "age" => "28"})
+    params.nested?(:user).should eq({"name" => "Paul", "age" => "28"})
   end
 
-  it "works when parsing multipart params twice" do
+  it "works when parsing json params twice" do
     request = build_request body: {page: 1}.to_json,
       content_type: "application/json",
       fixed_length: true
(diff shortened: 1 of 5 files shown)


spec/lucky/params_spec.cr
```diff
@@ -443,6 +443,109 @@ describe Lucky::Params do
...
   end
 
+  describe "nested_arrays" do
+    it "gets nested arrays from form encoded params" do
+      request = build_request body: "user:name=paul&user:langs[]=ruby&user:langs[]=elixir",
+        content_type: "application/x-www-form-urlencoded"
+
+      params = Lucky::Params.new(request)
+
+      params.nested_arrays?(:user).should eq({"langs" => ["ruby", "elixir"]})
```
wontruefree, line 453: If this is `eq` not `includes` shouldnt `"name" => "paul"` be in the params hash?

jwoertink: This one was a little tricky. The method only returns the nested arrays, it doesn't return any other keys. This is similar to how the file param stuff works.

The reason why is because I either had to turn the non-array pairs in to `"name" => ["paul"]` to keep the single  `Hash(String, Array(String))` type, or, have the value be a union in which case calling methods on the values becomes a little more difficult. If we merged all of them in to a single method, then the type signature becomes pretty gnarly.

jwoertink:
I was originally going to look at also dealing with files and arrays of files in this, but then I started thinking that it's probably a super rare case where you would set a file that's not nested in your action. Any file handling you would do will most likely be done in an operation (which still doesn't work). The `param` macro in actions seems to be for query params most often. However, with this code setup, we could add that in for a separate PR if it's something people want.

matthewmcgarvey (approved):

PR #1532, opened 2021-07-01
Limit Lucky to Crystal 1.0.0

### LINKED ISSUE #1531: PasswordResets actions are broken in new generated apps
Specifically [this line](https://github.com/luckyframework/lucky/pull/1527/files#diff-36333e6eb8dfce0769c277ac82db57e16a878dfd239d3861f88faaee44a78c5eR413) from https://github.com/luckyframework/lucky/pull/1527 changed the generated param methods from a regular method to a `getter` macro using a block. The `getter` version allows for built-in memoizing for a small performance boost, but the generated methods from that don't set the type restriction on them. This means that when an abstract method requires a return type, it fails like this:

```
There was a problem expanding macro 'getter'

Code in macro 'param'

 6 | getter token : String do
     ^
Called macro defined in macro 'macro_140526316403104'

 117 | macro getter(*names, &block)

Which expanded to:

 >  7 |           @token : String?
 >  8 | 
 >  9 |           def token
                     ^----
Error: this method overrides Auth::PasswordResets::RequireToken#token() which has an explicit return type of String.

Please add an explicit return type (String or a subtype of it) to this method as well.
```

We might be able to just revert that line, and then see if there's any major performance hits on it. If not, then we can leave it un-memoized for now.

matthewmcgarvey:
## Purpose

Fixes #1531

## Checklist

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,7 +1,7 @@
 name: lucky
 version: 0.27.2
 
-crystal: ">=0.36.1, < 2.0.0"
+crystal: ">=1.0.0"
 
 authors:
   - Paul Smith <paulcsmith0218@gmail.com>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
         shard_file:
           - shard.yml
         crystal_version:
-          - 0.36.1
           - 1.0.0
         experimental:
           - false
@@ -45,7 +44,6 @@ jobs:
         shard_file:
           - shard.yml
         crystal_version:
-          - 0.36.1
           - 1.0.0
         experimental:
           - false


shard.yml
```diff
@@ -1,7 +1,7 @@
 name: lucky
 version: 0.27.2
 
-crystal: ">=0.36.1, < 2.0.0"
+crystal: ">=1.0.0"
```
matthewmcgarvey, line 4: I removed the upper limit because of this forum post https://forum.crystal-lang.org/t/artificial-shard-restrictions-version-2-0-0-considered-harmful/3319

stephendolan (approved):

jwoertink (approved):
Sweet! Yeah, lets do this with all of the other ones. I'm cool with removing the upper limit too. I originally copied that from another shard, but I saw that post too and it makes sense. Thanks for doing this 👍

PR #1536, opened 2021-07-04
Add ability to check that all routes maintain similar style (with escape hatches)

### LINKED ISSUE #1100: Add compile time checks to route macros to check for underscores or dashes
`get/put/post/delete/etc.` should check that there just underscores or just dashes. I'm not 100% sure how to make this configurable yet, but it needs to be one or the other. So you can say something like "enforce_only_underscores" or "enforce_only_dashes" somehow and it'll raise if you accidentally use the wrong separator in a path.

It should be dismissable with a `skip_path_checks` (or some other name) so you have an escape hatch if needed)

This is a complex one but will be very worth it

paulcsmith:
Closes #1100 

This caught a bunch of routes in our test suite that used dashes 🤪 So it works!

Need to add this to the default Lucky CLI template in a follow-up PR

Screen Shot 2021-07-04 at 11 56 37 AM

### DIFF
--- a/spec/lucky/action_pipes_spec.cr
+++ b/spec/lucky/action_pipes_spec.cr
@@ -33,7 +33,7 @@ end
 class Pipes::Skipped < InheritablePipes
   skip set_before_cookie, overwrite_after_cookie
 
-  get "/skipped-pipes" do
+  get "/skipped_pipes" do
     plain_text "Body"
   end
 end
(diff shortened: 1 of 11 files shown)


jwoertink (approved):
Yeah, this looks pretty simple!

paulcsmith:
Thanks for the review!

PR #1537, opened 2021-07-09
Add subdomain support

### LINKED ISSUE #1166: Native subdomain support
Thought there was already an existing issue, but I couldn't find one. I guess we've just talked about it in the gitter chat and it was never brought up here.

## The problem

Subdomains are pretty commonly used to separate pieces of a site. You may want to give each user a subdomain based on their username, add subdomains for different regions, etc.

With Rails they have the notion of constraints, which allow you to constrain a route using different parameters, including a subdomain. For a dynamic subdomain you just subclass `Constraint` and put your resolving logic in there.

## Solution?

I have a couple potential ideas for a solution. The first of which doesn't require many changes to the way things are already done. Consider the following action:

```crystal
class Users::Index < BrowserAction
  get "/users/:username" do
    # Check for an existing user
    html IndexPage, user: user
  end
end
````

A simple way to handle subdomains would be to just add an optional `sub` or `subdomain` parameter to the existing routing macros. So the if I wanted to change the above code to use a subdomain instead, maybe we could do something like this:

```crystal
class Users::Index < BrowserAction
  get "/", sub: ":username" do
    # Check for an existing user
    html IndexPage, user: user
  end
end
````

or for a non-wildcard type domain:

```crystal
class Store::Index < BrowserAction
  get "/", sub: "store" do
    # Render store.mysite.com
    html IndexPage
  end
end
```

Another potential solution would be to do the Rails'y thing and use the concept of constraints. This would take a bit more work, but could allow routes to be further constrained using validations. How this could work would take a bit of discussion, but it would also make routing in Lucky even more powerful.

matthewmcgarvey:
## Purpose

Resolves #1166 

## Description

Adds subdomain support to Lucky.

### For request requirement

```crystal
abstract class AdminAction < Lucky::Action
  include Lucky::Subdomain

  register_subdomain "admin" # requires all requests to be made through admin.example.com
end
```

### For subdomain access

```crystal
class Payments::Create < BrowserAction
  include Lucky::Subdomain
  register_subdomain ["tenant1", "tenant2", ...]

  post "/payments" do
    tenant = TenantQuery.new.name(subdomain).first
    ...
  end
end
```

## Checklist

### DIFF
--- a/spec/lucky/subdomain_spec.cr
+++ b/spec/lucky/subdomain_spec.cr
@@ -8,15 +8,21 @@ abstract class BaseAction < Lucky::Action
 end
 
 class Simple::Index < BaseAction
-  register_subdomain
+  require_subdomain
 
   get "/simple" do
     plain_text subdomain
   end
 end
 
+class OptionalSubdomain::Index < BaseAction
+  get "/optional" do
+    plain_text subdomain? || "none"
+  end
+end
+
 class Specific::Index < BaseAction
-  register_subdomain "foo"
+  require_subdomain "foo"
 
   get "/specific" do
     plain_text subdomain
@@ -24,7 +30,7 @@ class Specific::Index < BaseAction
 end
 
 class Regex::Index < BaseAction
-  register_subdomain /www\d/
+  require_subdomain /www\d/
 
   get "/regex" do
     plain_text subdomain
@@ -32,7 +38,7 @@ class Regex::Index < BaseAction
 end
 
 class Multiple::Index < BaseAction
-  register_subdomain ["test", "staging", /(prod|production)/]
+  require_subdomain ["test", "staging", /(prod|production)/]
 
   get "/multiple" do
     plain_text subdomain
@@ -44,6 +50,16 @@ describe Lucky::Subdomain do
     request = build_request(host: "foo.example.com")
     response = Simple::Index.new(build_context(request), params).call
     response.body.should eq "foo"
+  end
+
+  it "handles optional subdomain" do
+    request = build_request(host: "qa.example.com")
+    response = OptionalSubdomain::Index.new(build_context(request), params).call
+    response.body.should eq "qa"
+
+    request = build_request(host: "example.com")
+    response = OptionalSubdomain::Index.new(build_context(request), params).call
+    response.body.should eq "none"
   end
 
   it "raises error if subdomain missing" do
--- a/spec/lucky/subdomain_spec.cr
+++ b/spec/lucky/subdomain_spec.cr
@@ -1,0 +1,129 @@
+require "../spec_helper"
+
+include ContextHelper
+
+abstract class BaseAction < Lucky::Action
+  include Lucky::Subdomain
+  accepted_formats [:html], default: :html
+end
+
+class Simple::Index < BaseAction
+  register_subdomain
+
+  get "/simple" do
+    plain_text subdomain
+  end
+end
+
+class Specific::Index < BaseAction
+  register_subdomain "foo"
+
+  get "/specific" do
+    plain_text subdomain
+  end
+end
+
+class Regex::Index < BaseAction
+  register_subdomain /www\d/
+
+  get "/regex" do
+    plain_text subdomain
+  end
+end
+
+class Multiple::Index < BaseAction
+  register_subdomain ["test", "staging", /(prod|production)/]
+
+  get "/multiple" do
+    plain_text subdomain
+  end
+end
+
+describe Lucky::Subdomain do
+  it "handles general subdomain expectation" do
+    request = build_request(host: "foo.example.com")
+    response = Simple::Index.new(build_context(request), params).call
+    response.body.should eq "foo"
+  end
+
+  it "raises error if subdomain missing" do
+    request = build_request(host: "example.com")
+    expect_raises(Lucky::InvalidSubdomainError) do
+      Simple::Index.new(build_context(request), params).call
+    end
+  end
+
+  it "handles specific subdomain expectation" do
+    request = build_request(host: "foo.example.com")
+    response = Specific::Index.new(build_context(request), params).call
+    response.body.should eq "foo"
+  end
+
+  it "raises error if subdomain does not match specific" do
+    request = build_request(host: "admin.example.com")
+    expect_raises(Lucky::InvalidSubdomainError) do
+      Specific::Index.new(build_context(request), params).call
+    end
+  end
+
+  it "handles regex subdomain expectation" do
+    request = build_request(host: "www4.example.com")
+    response = Regex::Index.new(build_context(request), params).call
+    response.body.should eq "www4"
+  end
+
+  it "raises error if subdomain does not match regex" do
+    request = build_request(host: "4www.example.com")
+    expect_raises(Lucky::InvalidSubdomainError) do
+      Regex::Index.new(build_context(request), params).call
+    end
+  end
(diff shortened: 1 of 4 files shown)


src/lucky/subdomain.cr
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  def subdomain : String
+    {% raise "No subdomain available without calling `register_subdomain` first." %}
+  end
+
+  private def _fetch_subdomain : String?
+    host = request.hostname
+    return if host.nil? || IP_HOST_REGEXP.matches?(host)
```
wontruefree, line 41: Is there a way to use `Socket::IPAddress`. This will handle IPV6 and IPV4 and it looks like it does not use a Regex so I think it might be faster.

matthewmcgarvey: I did try to look for something in the stdlib and didn't find anything. I didn't see this class. Unfortunately the docs for it say this:

> IPAddress won't resolve domains, including localhost. If you must resolve an IP, or don't know whether a String contains an IP or a domain name, you should use Addrinfo.resolve instead.

And it looks like `Addrinfo.resolve` won't answer the question of whether or not the host uses an ip address.

wontruefree: That is a good point.  You might looks at this to check if this is an IP.  But it is not as nice as having using all stdlib tooling to fetch the sub domain.

https://crystal-lang.org/api/1.0.0/Socket.html#ip?(string:String)-class-method
https://play.crystal-lang.org/#/r/bicq

jwoertink: I agree with @wontruefree. This one may be a little tricky since IPV6 would get passed through, but since `Addrinfo.resolve` works on normal hosts, we may get a false positive. This `Socket::IPAddress` looks like it may work https://crystal-lang.org/api/1.0.0/Socket/IPAddress.html

If you pass in `"localhost"`, it'll raise an exception. The only thing I don't like is that the IP would actually be the edge case here which means we're raising an exception and rescuing 90% of the time which is probably a performance penalty 😝  ...

Maybe this is a "fine enough for now" scenario, and we just open a separate issue to figure out a better way?

src/lucky/subdomain.cr
```diff
@@ -0,0 +1,71 @@
...
+
+  alias Matcher = String | Regex | Bool | Array(String | Regex) | Array(String) | Array(Regex)
+
+  # Sets up a subdomain requirement for an action
+  #
+  # ```
+  # register_subdomain                                    # subdomain required but can be anything
+  # register_subdomain "admin"                            # subdomain required and must equal "admin"
+  # register_subdomain /(dev|qa|prod)/                    # subdomain required and must match regex
+  # register_subdomain ["tenant1", "tenant2", /tenant\d/] # subdomain required and must match one of the items in the array
```
jwoertink, line 21: I love all the different options. Thinking about how I currently use subdomains, I only need a subdomain on my routes when I'm in staging. In production, we don't use subdomains. Is that possible to do? I'm thinking when you do `qa.luckyframework.org`, that would only be on the QA environment, and not in production.. So we need to be able to say `register_subdomain "qa" if LuckyEnv.qa?` or whatever...

src/lucky/subdomain.cr
```diff
@@ -0,0 +1,71 @@
...
+      _fetch_subdomain.not_nil!
+    end
+
+    private def _match_subdomain
+      _match_subdomain({{ matcher }})
+    end
+  end
+
+  def subdomain : String
+    {% raise "No subdomain available without calling `register_subdomain` first." %}
```
jwoertink, line 36: 👍

spec/lucky/subdomain_spec.cr
```diff
@@ -0,0 +1,129 @@
...
+class Simple::Index < BaseAction
+  register_subdomain
+
+  get "/simple" do
+    plain_text subdomain
+  end
+end
+
+class Specific::Index < BaseAction
+  register_subdomain "foo"
```
jwoertink, line 19: Does this work for multiple subdomain too? Like `"staging.dashboard.foo"`

src/lucky/subdomain.cr
```diff
@@ -0,0 +1,71 @@
...
+        actual.to_s == expected_subdomain.to_s
+      else
+        expected_subdomain === actual
+      end
+    end
+
+    if result
+      continue
+    else
+      raise InvalidSubdomainError.new
```
jwoertink, line 68: Maybe this error could pass in the expected subdomain and what it tried to match against? That might make it better for reporting in Bugsnag or whatever.

jwoertink (commented):
I dig this direction, and I think it gives us some great flexibility. I left a few comments, but overall I'm happy with it.

src/lucky/subdomain.cr
matthewmcgarvey: If you follow the rails link I left on the constant, the regex is exactly how Rails handles subdomains. It doesn't extract a subdomain if it matches the regex. I guess I'm not seeing how this wouldn't work.

spec/lucky/subdomain_spec.cr
matthewmcgarvey: My current job uses "multiple subdomains" as well because we have a tenant subdomain and the environment subdomain so in staging a url might look like `tenant1.staging.example.com`.

This subdomain support really focuses on 1 subdomain. Without any special handling, the subdomain for my example would be `tenant1.staging` and it would be up to me to add special logic to parse it further.

Another way to handle that would be to configure the `tld_length` to be 2 only in staging so that the subdomain would become `tenant1` but I think that might be a sort of hack.

src/lucky/subdomain.cr
matthewmcgarvey: I have pushed an update to better clarify. The macro is now called `require_subdomain`.

In this situation you don't actually require a subdomain, you just do have a subdomain in certain scenarios. In those situations you wouldn't use `require_subdomain` but you can still check if a subdomain was provided by calling `subdomain?`. Calling `subdomain` will raise a compile time error though.

src/lucky/subdomain.cr
jwoertink: It looks like rails uses that second regex https://github.com/rails/rails/blob/afc6abb674b51717dac39ea4d9e2252d7e40d060/actionpack/lib/action_dispatch/http/url.rb#L9 which handles IPV6 

```
/(^[^:]+:\/\/)?(\[[^\]]+\]|[^:]+)(?::(\d+$))?/ =~ "fe80::aede:48ff:fe00:1122"
```

It seems they're only using it when building the host url though. So I think we can just ignore it for now. I've never actually seen anyone access a site using IPV6. Maybe API calls for hackers trying to find vulnerabilities? I think this is a "fine enough for now" scenario though.

spec/lucky/subdomain_spec.cr
jwoertink: Ok, yeah, that's what my sites do too. `staging.tenant1.example.com`. That's fine though. We just would document that on the site saying if you use deep nested subdomains, then you'll need to do your own split and parse.

src/lucky/subdomain.cr
jwoertink: Cool. That gives us a nice escape hatch then too.

jwoertink (approved):
Alright! I think this is ready to go. Can you just open an issue on the website to document this after it's all merged in? Thanks!

PR #1538, opened 2021-07-10
Ensure routes are unique

paulcsmith:
Tested by changing one of our routes in the specs to collide and it worked

Closes https://github.com/luckyframework/lucky/issues/1539

Screen Shot 2021-07-10 at 4 58 35 PM

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -148,10 +148,13 @@ module Lucky::Routable
   NORMALIZED_ROUTES = [] of Nil # types are not resolved in macros
 
   # :nodoc:
-  macro enforce_route_uniqueness(original_path)
+  macro enforce_route_uniqueness(method, original_path)
     # Regex for capturing the param part for normalization
+    #
+    # So "/users/:user_id" is changed to "/users/:normalized"
     {% normalized_path = original_path.gsub(/(\:\w*)/, ":normalized") %}
-    {% if already_used_route = NORMALIZED_ROUTES.find { |route| route[:normalized_path] == normalized_path } %}
+
+    {% if already_used_route = NORMALIZED_ROUTES.find { |route| route[:normalized_path] == normalized_path && route[:method] == method } %}
       {% raise <<-ERROR
       #{already_used_route[:original_path]} in '#{already_used_route[:action]}' collides with the route in '#{@type.name}'
 
@@ -181,6 +184,7 @@ module Lucky::Routable
       {% NORMALIZED_ROUTES << {
            normalized_path: normalized_path,
            original_path:   original_path,
+           method:          method,
            action:          @type.name,
          } %}
     {% end %}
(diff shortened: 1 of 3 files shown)


src/lucky/routable.cr
```diff
@@ -145,14 +145,51 @@ module Lucky::Routable
...
   end
 
+  NORMALIZED_ROUTES = [] of Nil # types are not resolved in macros
+
+  # :nodoc:
+  macro enforce_route_uniqueness(method, original_path)
+    # Regex for capturing the param part for normalization
+    #
+    # So "/users/:user_id" is changed to "/users/:normalized"
+    {% normalized_path = original_path.gsub(/(\:\w*)/, ":normalized") %}
```
paulcsmith, line 155: This way `/users/:id` and `/users/:user_id` are counted as the same route, since that's how the router tries to match them.

paulcsmith:
There are some edge cases with optional route parts, but overall this should catch *most* route collissions

src/lucky/skip_unique_route_check.cr
```diff
@@ -0,0 +1,6 @@
+# Include this in an action to skip route uniqueness checks.
+module Lucky::SkipUniqueRouteCheck
```
matthewmcgarvey, line 2: Not sure that it's necessary to provide this. If you skip the compile time check, you're still going to get a runtime error from the duplicate checks in lucky_router https://github.com/luckyframework/lucky_router/blob/d6020f0c404e033abb920cf29e34ec8d752525c2/src/lucky_router/matcher.cr#L50

jwoertink (approved):
Sweet! This is a great addition.

src/lucky/routable.cr
```diff
@@ -145,14 +145,51 @@ module Lucky::Routable
...
+
+    {% if already_used_route = NORMALIZED_ROUTES.find { |route| route[:normalized_path] == normalized_path && route[:method] == method } %}
```
kevinsjoberg, line 157: Not sure if this matter or not, but would it make sense (from a lookup perspective) to make `NORMALIZED_ROUTES` a `Hash` instead? The normalized path would be the key in this case.

paulcsmith: @kevinsjoberg I tried, but there are some issues with Hash/NameTuples in macro constants that I couldn't seem to get around :(

Also since it needs to match both method *and* normalized path I think this may actually work better. Since if we used the normalized path as key that would mean there is no way to lookup based on method. I suppose we could combine the method and path into one string key though 🤔 Maybe someone can do a cleanup pass later and can figure out how to use the Hash in a macro

src/lucky/skip_unique_route_check.cr
paulcsmith: Good point. We discussed this in chat and I think we can leave this here pre-1.0 (or at least for one version) since this uses different logic for checking duplicates. If it seems to not have issues then we can remove it!

Does that sound good?

paulcsmith: Created an issue to track this here: https://github.com/luckyframework/lucky/issues/1541

PR #1542, opened 2021-07-13
Store normalized routes in `Hash`

kevinsjoberg:
This is a follow up of https://github.com/luckyframework/lucky/pull/1538, more specifically [this](https://github.com/luckyframework/lucky/pull/1538#discussion_r667710183) discussion.

Routes are now stored in a `Hash` instead of an `Array`. I did some isolated testing on my own, and it seems to be working. @paulcsmith feel free to jump in here if I'm doing something weird. 🙂

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -145,7 +145,7 @@ module Lucky::Routable
     # no-op by default
   end
 
-  NORMALIZED_ROUTES = [] of Nil # types are not resolved in macros
+  private NORMALIZED_ROUTES = {} of _ => _
 
   # :nodoc:
   macro enforce_route_uniqueness(method, original_path)
@@ -153,8 +153,9 @@ module Lucky::Routable
     #
     # So "/users/:user_id" is changed to "/users/:normalized"
     {% normalized_path = original_path.gsub(/(\:\w*)/, ":normalized") %}
+    {% normalized_key = "#{method.id} #{normalized_path.id}" %}
 
-    {% if already_used_route = NORMALIZED_ROUTES.find { |route| route[:normalized_path] == normalized_path && route[:method] == method } %}
+    {% if already_used_route = NORMALIZED_ROUTES[normalized_key] %}
       {% raise <<-ERROR
       #{original_path} in '#{@type.name}' collides with the path in '#{already_used_route[:action]}'
 
@@ -173,7 +174,7 @@ module Lucky::Routable
       ERROR
       %}
     {% else %}
-      {% NORMALIZED_ROUTES << {
+      {% NORMALIZED_ROUTES[normalized_key] = {
            normalized_path: normalized_path,
            original_path:   original_path,
            method:          method,


src/lucky/routable.cr
```diff
@@ -145,16 +145,17 @@ module Lucky::Routable
...
   # :nodoc:
   macro enforce_route_uniqueness(method, original_path)
     # Regex for capturing the param part for normalization
     #
     # So "/users/:user_id" is changed to "/users/:normalized"
     {% normalized_path = original_path.gsub(/(\:\w*)/, ":normalized") %}
+    {% normalized_key = "#{method.id} #{normalized_path.id}" %}
 
-    {% if already_used_route = NORMALIZED_ROUTES.find { |route| route[:normalized_path] == normalized_path && route[:method] == method } %}
+    {% if already_used_route = NORMALIZED_ROUTES[normalized_key] %}
```
jwoertink, line 158: If `NORMALIZED_ROUTES[normalized_key]` doesn't exist, this still works? In runtime, you'd have to add the `?` to say it's nilable otherwise it'd throw an error. I guess it works different in macro land?

src/lucky/routable.cr
```diff
@@ -145,16 +145,17 @@ module Lucky::Routable
     # no-op by default
   end
 
-  NORMALIZED_ROUTES = [] of Nil # types are not resolved in macros
+  NORMALIZED_ROUTES = {} of String => Hash(Symbol, String)
```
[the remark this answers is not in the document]
jwoertink: Actually, for funsies... can you make this `{} of String => Nil` and see if it all still works?

jwoertink (commented):
I will say that this code actually reads nicer. It's easier to see what it's doing. I just want to make sure that there's no weird cases with Hash working differently in macros.

src/lucky/routable.cr
kevinsjoberg: Yeah. I was equally surprised by this. See https://github.com/crystal-lang/crystal/pull/10930. 🙂

src/lucky/routable.cr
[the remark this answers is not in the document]
kevinsjoberg: @jwoertink nice catch! 👍 Totally missed that (I guess I was in Ruby mode). This surprises me as well. Looks like it compiles fine, but will break at runtime if you reference `NORMALIZED_ROUTES`. The same is true for `{} of String => Nil`.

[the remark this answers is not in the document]
matthewmcgarvey: It is because the macro changes the type of the constant. The type you define there is only applied if you don't touch the constant.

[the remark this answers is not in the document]
kevinsjoberg: @matthewmcgarvey @jwoertink I've amended the original commit. However, I seem to be getting the following error if I try to print `NORMALIZED_ROUTES` like so:

```crystal
puts NORMALIZED_ROUTES.inspect
```

```console
Unhandled exception: can't execute `{normalized_path: "/api/v1/prefixed_get", original_path: "/api/v1/prefixed_get", method: :get, action: PrefixedActions}` at  (Exception)
  from /usr/share/crystal/src/log/main.cr:49:13 in '???'
  from src/lucky/router.cr:3:3 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:110:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:96:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:119:3 in 'main'
  from __libc_start_main
  from _start
  from ???
ERROR: 1
```

Not entirely sure why. The error seem to come from Crystal's `cleanup_transformer.cr`. More specifically, the `untyped_expression`  method. But in my small playground example, I'm not seeing the same thing.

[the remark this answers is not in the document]
kevinsjoberg: I've amended the original commit. I made `NORMALIZED_ROUTES` private and declared the type `{} of Nil => Nil` as it doesn't really matter for this use case.

src/lucky/routable.cr
```diff
@@ -173,7 +174,7 @@ module Lucky::Routable
       ERROR
       %}
     {% else %}
-      {% NORMALIZED_ROUTES << {
+      {% NORMALIZED_ROUTES[normalized_key] = {
            normalized_path: normalized_path,
            original_path:   original_path,
            method:          method,
```
jwoertink, line 180: My first thought was "we don't need most of this now". The `action` key is really the only value being used, but maybe it doesn't hurt to have the extra data?

kevinsjoberg: I was thinking the same thing, but I guess it's only present and compile time and not in the actual executable in any case? I also didn't want to change anything that wasn't completely necessary from the original implementation without first discussing it with the rest of you. 🙂

src/lucky/routable.cr
```diff
@@ -145,16 +145,17 @@ module Lucky::Routable
     # no-op by default
   end
 
-  NORMALIZED_ROUTES = [] of Nil # types are not resolved in macros
+  private NORMALIZED_ROUTES = {} of _ => _
```
jwoertink, line 148: I think we can keep this `Nil => Nil`. That's pretty common in other areas. I don't think we use the `_` often outside of maybe a Proc. Plus with the recent commit on Crystal, I'm not sure if the Underscore stuff will change or not.

straight-shoota: FWIW: Using underscore as type argument is IMO ideal because it clearly designates this constant as used at compile-time only.

jwoertink: If @straight-shoota is cool with it, then that's good enough for me!

jwoertink (approved):
Thanks for fixing this up!

matthewmcgarvey (approved):

PR #1543, opened 2021-07-16
Adding Crystal 1.1.0 to CI

jwoertink:
## Purpose
Adding Crystal 1.1.0 to the CI

PR #1544, opened 2021-07-16
Switch to oprypins github action for installing crystal

matthewmcgarvey:
Looks like they haven't released Crystal 1.1.0 for the alpine docker images https://github.com/luckyframework/lucky/pull/1543

I think switching to https://github.com/crystal-lang/install-crystal will mean that we won't have to wait going forward.


stephendolan (approved):
Yeah, I'm cool with this! I've been using the GitHub action to install Crystal on my shards for a bit now and haven't run into any issues.

PR #1554, opened 2021-07-23
Include Jeremy in sponsorship options

stephendolan:
Jeremy is a massive contributor to the framework and should be featured as a person who can be sponsored.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,3 @@
 # These are supported funding model platforms
 
-github: [paulcsmith]
+github: [paulcsmith, jwoertink]


.github/FUNDING.yml
```diff
@@ -1,3 +1,3 @@
 # These are supported funding model platforms
 
-github: [paulcsmith]
+github: [paulcsmith, jwoertink]
```
jwoertink, line 3: Thanks 😅  ❤️  Maybe we should just add everyone on the core team so it doesn't look like the "paul and jeremy show"? I don't want to give the wrong impression that people can support me, but not you. Not sure if @edwardloveall or @matthewmcgarvey have sponsor links... Maybe it'll do a silent fail but magically work if they add sponsor links?

edwardloveall: I'm okay without sponsorship. Thank you for thinking of me!

PR #1556, opened 2021-07-24
fix: change the secret key target

### LINKED ISSUE #1555: Secret key task compiles on run, but should be precompiled
https://github.com/luckyframework/lucky/blob/f2f86d639ae59be4149edbdaf9bb97348cf13c37/shard.yml#L30-L31

I think it's because the target here is called `lucky.secret_key` instead of `lucky.gen.secret_key`

RhnSharma:
## Purpose
Fixes #1555
Hi @jwoertink, how does this look?
Let me know if this needs any changes.
Thanks

## Checklist

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -27,7 +27,7 @@ targets:
     main: src/precompiled_tasks/gen/component.cr
   lucky.gen.task:
     main: src/precompiled_tasks/gen/task.cr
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
This looks great! Thanks so much for doing this.

PR #1568, opened 2021-08-13
Added more type options for status in the html_with_status macro

### LINKED ISSUE #1558: Can we use an HTTP::Status enum in this method?
https://github.com/luckyframework/lucky/blob/a928465ed27856600f88802712bd40120d13087a/src/lucky/renderable.cr#L79-L81

If we can, then lets add that enum and manually call `.value` on it.

bnjamin:
The `html_with_status` macro now also accepts Symbols and `HTTP::Status` as the status argument.

## Purpose
Fixes #1558

## Description
I also added the option to use symbols, no worries they are typesafe, so you will get a compile time error in case of typos.
I think it reads better with a symbol, maybe i'm weird like that 🤷 

```crystal
html_with_status EditPage, :unprocessable_entity, operation: operation, post: post
```
## Checklist

### DIFF
--- a/src/lucky/renderable.cr
+++ b/src/lucky/renderable.cr
@@ -69,6 +69,8 @@ module Lucky::Renderable
 
   # Render an HTMLPage with a status other than 200
   #
+  # The status can either be a Number, a HTTP::Status, or a Symbol that corresponds to the HTTP::Status.
+  #
   # ```
   # class SecretAgents::Index < BrowserAction
   #   get "/shhhh" do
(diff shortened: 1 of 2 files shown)


src/lucky/renderable.cr
```diff
@@ -69,29 +69,43 @@ module Lucky::Renderable
...
+  # See Crystal's
+  # [HTTP::Status](https://crystal-lang.org/api/latest/HTTP/Status.html)
+  # enum for more available http status codes.
   macro html_with_status(page_class, status, **assigns)
-    {% if !status.is_a?(NumberLiteral) %}
+    {% if status.is_a?(NumberLiteral) %}
+      html {{ page_class }}, _with_status_code: {{ status }}, {{ **assigns }}
+    {% elsif status.is_a?(SymbolLiteral) %}
+      html {{ page_class }}, _with_status_code: HTTP::Status::{{ status.upcase.id }}.value, {{ **assigns }}
+    {% elsif status.resolve.is_a?(NumberLiteral) %}
```
bnjamin, line 89: This is for the `HTTP::Status` enum case. When in "macro land" status is a `Crystal::Macros::Path`. At first i tried `status.is_a?(HTTP::Status)` however that would return false.

jwoertink: Ah, interesting. So the `Crystal::Macros::Path#resolve` will turn that enum in to an Integer?

src/lucky/renderable.cr
```diff
@@ -69,29 +69,43 @@ module Lucky::Renderable
...
   # ```
+  # See Crystal's
+  # [HTTP::Status](https://crystal-lang.org/api/latest/HTTP/Status.html)
+  # enum for more available http status codes.
   macro html_with_status(page_class, status, **assigns)
-    {% if !status.is_a?(NumberLiteral) %}
+    {% if status.is_a?(NumberLiteral) %}
+      html {{ page_class }}, _with_status_code: {{ status }}, {{ **assigns }}
+    {% elsif status.is_a?(SymbolLiteral) %}
+      html {{ page_class }}, _with_status_code: HTTP::Status::{{ status.upcase.id }}.value, {{ **assigns }}
```
jwoertink, line 88: Nice! I didn't know you could call `upcase` right on the symol at the macro level like that.

jwoertink (approved):
Sweet! Thanks for this. I dig it :+1:

src/lucky/renderable.cr
bnjamin: That's right, so it will actually work with all Enums with the current implementation, maybe we should add a check to see if the `Path` starts with `HTTP::Status`

src/lucky/renderable.cr
bnjamin: It was also a first for me

src/lucky/renderable.cr
jwoertink: That might not be a bad idea. You wouldn't want to accidentally send out a status of `4` :joy:  Then again.. maybe you do? :thinking:  The whole reason for this method was more of an escape hatch since for the most part, you'd want to just use the `html` macro...

bnjamin: I added the check, if you really wanted to sent out a status of `4` you could just pass a `Number`. Here be dragons 🐉.
Maybe it won't be an uncommon escsape hatch for long, the `Turbo` library from `Hotwired` expects a `422 unprocessable entity` when a form fails with validation errors.

jwoertink: Sweet. Sounds good. Thanks!

jwoertink (approved):

PR #1581, opened 2021-09-23
Removing compile-time catches for html_with_status

### LINKED ISSUE #1575: html_with_status macro breaks on built-in Error page
Using the `html_with_status` macro in our Errors action breaks because that page `needs status : Int32` to be passed over. Since this macro has a status argument, doing this throws an error:

```
html_with_status Errors::ShowPage, 404, message: message, status: status

In src/actions/errors/show.cr:53:5
web          |
web          |  53 | html_with_status Errors::ShowPage, 404, message: message, status: status
web          |       ^---------------
web          | Error: no argument named 'message'
```

My guess is because the `status: status` matches a named arg, so the compiler thinks `message` should too..

jwoertink:
## Purpose
Fixes #1575 

## Description
If your status code comes from a variable, this was failing because it was looking for literal objects only. This was originally to ensure you were passing in something proper since the macro arguments don't catch type-safety.

Removing those compile-time catches allows us to just fallback to a "whatever you passed in". If it's a Number, then it will pass that over. If it's a variable that returns a Number, then it will be fine. Anything else, and you'll end up getting a "No overload matches" error from `TextResponse`.

## Checklist

### DIFF
--- a/spec/lucky/action_rendering_spec.cr
+++ b/spec/lucky/action_rendering_spec.cr
@@ -21,7 +21,8 @@ end
 
 class Rendering::Show::WithStatus < TestAction
   get "/rendering/:nothing" do
-    html_with_status IndexPage, 419, title: "Closing Time", arg2: "You don't have to go home but you can't stay here"
+    gather_status = 419
+    html_with_status IndexPage, gather_status, title: "Closing Time", arg2: "You don't have to go home but you can't stay here"
   end
 end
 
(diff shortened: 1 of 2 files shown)


spec/lucky/action_rendering_spec.cr
```diff
@@ -21,7 +21,8 @@ end
 
 class Rendering::Show::WithStatus < TestAction
   get "/rendering/:nothing" do
-    html_with_status IndexPage, 419, title: "Closing Time", arg2: "You don't have to go home but you can't stay here"
+    gather_status = 419
+    html_with_status IndexPage, gather_status, title: "Closing Time", arg2: "You don't have to go home but you can't stay here"
```
robacarp, line 25: 😆

robacarp (approved):

PR #1582, opened 2021-09-23
Fix compile error

### LINKED ISSUE #1579: Passing optional param to `Lucky::Routable.with` leads to compile error
https://github.com/luckyframework/lucky/blob/4c89e1280eddeed1fca83960d86ed2f2c90da465/src/lucky/routable.cr#L314

Can we do without the type restriction of `optional_path_params` here?

I have the following route definition:
```crystal
class Refunds::Create < BrowserAction
  post "/receipts/?:receipt_id/refunds" do
    # ...
  end
end
```

Calling `Refunds::Create.with(receipt_id: 2)` leads to a compile error. The call does not match the method definition of `Lucky::Routable.route` above, causing it to fall through to the next definition of `Lucky::Routable.route`, which is set up to raise a compile error:

```
> 3 |     
> 4 | 
> 5 |     add_route(:get, "/receipts/?:receipt_id/refunds", Refunds::Create)
           ^
Error: Invalid call to Refunds::Create.route

Optional arguments:

- receipt_id

For more information, refer to https://luckyframework.org/guides/http-and-routing/link-generation.
```

Calling `Refunds::Create.with(receipt_id: "2")` works. (Note: "2" here is a `String`). This behaviour is inconsistent with all other params, since they can be passed to `Lucky::Routable.route` or `Lucky::Routable.with` without explicitly converting to `String`.

akadusei:
Closes #1579.

Calls to `Lucky::Routable.route` failed to compile when passed optional path param values other than strings.

## Checklist

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -311,7 +311,7 @@ module Lucky::Routable
 
     # optional path variables are nilable
     {% for param in optional_path_params %}
-      {{ param.gsub(/^\?:/, "").id }} : String? = nil,
+      {{ param.gsub(/^\?:/, "").id }} = nil,
     {% end %}
     anchor : String? = nil
       ) : Lucky::RouteHelper


jwoertink (approved):
sweet! Thanks for fixing this, and catching the issue.

Issue #1524, opened 2021-06-09
A lucky project dependency is causing ameba to show up in lib folder


matthewmcgarvey:
Maybe not a big deal, but I noticed in my project I generated with lucky cli that ameba is showing up in the `lib/` folder even though it's not a dependency and is not a transitive dependency either. When I run `shards prune` it is removed but added back as soon as I run `shards install` again.

<img width="463" alt="CleanShot 2021-06-08 at 20 28 01@2x" src="https://user-images.githubusercontent.com/17329408/121274246-16522b00-c898-11eb-8e60-bd3c56ccc1a6.png">

jwoertink:
Weird... Would this be a shards bug? Or maybe all development dependencies always show up in lib unless you do `shards install --production`?

jwoertink:
Looking in to this, it turns out it's Avram... Though, I'm still not sure why. Removed Avram from the list, and it doesn't show up. Ameba is a dev dependency of Avram, but it's also a dev dependency of ALL the Lucky shards. i created a new project with just Avram, and Ameba shows up. Then I made a project with just Habitat, and it doesn't show up....

jwoertink:
Found the issue. Turns out this also exists in Lucky, and Carbon. The issue is that when the `postinstall` runs for each of these shards, it calls `shards build` from those directories. That will run a `shards install` within that directory which includes development dependencies.

To fix this, we need to add `--without-development` to each of these `shards build`.

PR #1589, opened 2021-10-02
Prevent ameba from being installed during postinstall

jwoertink:
## Purpose
Fixes https://github.com/luckyframework/lucky/issues/1524

## Description
When Lucky is included as a dependency of an app, and you run `shards install`, it will call Lucky's `postinstall` script. That will then run `shards build` from within the lucky directory causing it to install all shards including development dependencies.

This PR tells it to only build the targets, but don't build development dependencies.

## Checklist

### DIFF
--- a/script/precompile_tasks
+++ b/script/precompile_tasks
@@ -3,7 +3,15 @@
 # Precompile tasks if not in production and not in CI
 # Also allow skipping precompilation with an ENV var
 if [ "$LUCKY_ENV" != "production" ] && [ -z "$CI" ] && [ -z "$SKIP_LUCKY_TASK_PRECOMPILATION" ]; then
-    shards build
+    # This is to prevent Lucky's development dependencies being installed
+    # from an external source like running `shards install` from a Lucky app.
+    # Install shards and build shard targets.
+    if [ -z "$BUILD_WITHOUT_DEVELOPMENT" ]; then
+        shards build
+    else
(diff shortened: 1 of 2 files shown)


script/precompile_tasks
```diff
@@ -3,7 +3,15 @@
...
 if [ "$LUCKY_ENV" != "production" ] && [ -z "$CI" ] && [ -z "$SKIP_LUCKY_TASK_PRECOMPILATION" ]; then
-    shards build
+    # This is to prevent Lucky's development dependencies being installed
+    # from an external source like running `shards install` from a Lucky app.
+    # Install shards and build shard targets.
+    if [ -z "$BUILD_WITHOUT_DEVELOPMENT" ]; then
+        shards build
+    else
+        shards build --without-development
+    fi
```
robacarp, line 13: Since the typical use case here would be to build without development, should this `if` be reversed so that the flag doesn't need to be passed in unless you actually want the ancillary binaries?

Do we even need the flag here? Assuming that I actually want the ameba binary I can add that to my own shards.build stanza for the project.

jwoertink: My bash skills suck 😅  it should probably be flipped

@edwardloveall and I were talking about this, and we figured if you're working on Lucky directly, maybe there's a time you want to run that script with any dev deps included? I think the 99% case though would be without.

PR #1593, opened 2021-10-19
Use shards build for release task

### LINKED ISSUE #1564: Change the release task to use shards build
https://github.com/luckyframework/lucky/blob/f1c83e132c53054a0aab5b21265691579df3835e/tasks/build/release.cr#L12

Instead of this just building the `start_server` and creating a binary called `start_server`, we should instead run `shards build --release` which will look at your shard.yml targets and build those. This means if you have workers and such, those would also get built.

donovanglover:
Fixes #1564.

### DIFF
--- a/tasks/build/release.cr
+++ b/tasks/build/release.cr
@@ -9,7 +9,7 @@ class Build::Release < LuckyTask::Task
   property error_io : IO = STDERR
 
   def call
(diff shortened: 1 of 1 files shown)


jwoertink:
Hey @donovanglover just wanted to check in. Looks like some specs are failing, and there's one last suggestion to round this out. Let me know if you have any questions. Thanks!

PR #1598, opened 2021-10-29
Use the proper boolean attributes in HTML form tags.

### LINKED ISSUE #1560: Use the proper bool attribute on opption selected
https://github.com/luckyframework/lucky/blob/a928465ed27856600f88802712bd40120d13087a/src/lucky_avram/ext/lucky/select_helpers.cr#L19

I remember reading somewhere that `selected="true"` is actually an invalid value, though most browsers will still render it properly. We should just convert this to the standard boolean attributes that we use.

jwoertink:
## Purpose
Fixes #1560

## Description
We had been using `checked="true"` or `selected="true"` which according to [these docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes) is invalid.

> To be clear, the values "true" and "false" are not allowed on boolean attributes.

## Checklist


src/lucky_avram/ext/lucky/input_helpers.cr
```diff
@@ -65,12 +65,13 @@ module Lucky::InputHelpers
                unchecked_value : String,
                checked_value : String,
                **html_options) : Nil forall T
+    attrs = [] of Symbol
     if field.param == checked_value
-      html_options = merge_options(html_options, {"checked" => "true"})
+      attrs = attrs | [:checked]
```
jwoertink, line 70: this one could technically call `<<`, but I just kept it consistent with the other forms in this file.

src/lucky_avram/ext/lucky/input_helpers.cr
```diff
@@ -80,7 +81,7 @@ module Lucky::InputHelpers
   def checkbox(field : Avram::PermittedAttribute(Bool), attrs : Array(Symbol), **html_options) : Nil
     unchecked_value = "false"
     if field.value
-      html_options = merge_options(html_options, {"checked" => "true"})
+      attrs = attrs | [:checked]
```
jwoertink, line 84: Can't use `<<` here because arguments are passed by reference. Which means that when we pass `EMPTY_BOOLEAN_ATTRIBUTES`, it would pass that to other methods causing ALL radios/checkboxes to be marked as checked

src/lucky_avram/ext/lucky/input_helpers.cr
```diff
@@ -211,6 +212,7 @@ module Lucky::InputHelpers
       "value" => input_value(field),
     }.merge(input_overrides)
     update_array_id_counter!(field)
+    attrs.uniq!
```
jwoertink, line 215: just in case so you don't accidentally send over duplicate attrs...

matthewmcgarvey (approved):

src/lucky_avram/ext/lucky/input_helpers.cr
robacarp: HTML is pretty fast and loose with it's attributes, especially in the age of React and other front end frameworks which make up their own elements...are we sure there is no use case for passing the same attribute more than once to a tag?

jwoertink: This would only prevent a tag being rendered like `<input ... checked checked required required />`. I'm sure browsers would just ignore it, so it's probably not required, but in the case of some random browser that breaks due to it, I figured why not? 🤷‍♂️

PR #1604, opened 2021-11-03
Making the watcher less greedy

### LINKED ISSUE #1421: Making the watcher less greedy?
The current `lucky watch` task runs [this loop](https://github.com/luckyframework/lucky/blob/b6f68bdc78b0b6831e801cac3ba0c97f83159e81/tasks/watch.cr#L222) that looks for the modification time to change on the files, and once it sees one that changes, it kicks off a compilation.

In some cases, you might have 2 or more crystal files open in your editor, and you need to save all of them. Right now your options are to save (kick off compile), save file 2 (kick off second compile), and save file 3 (kick off third compile)... or you just kill your whole running dev server to save all your files, and then wait for the reboot of lucky and the asset stuff (plus workers in my case).

This may be a long shot, but what would be cool is if we had a way to detect that a compile was happening, and if we try to kick off another compile, we kill the first one where it stands. Another option could be to wait like 2 seconds before starting the compile... though, I'm not big on that idea because in a normal app case, that slows down your compile turn-around.

skojin:
make watcher abort compilation (1) and compile again (2) when source file before compilation (1) is completed

fixes #1421

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -12,6 +12,7 @@ module LuckySentry
   class ProcessRunner
     include LuckyTask::TextHelpers
 
+    getter build_processes = [] of Process
     getter app_processes = [] of Process
     property successful_compilations
     property app_built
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
This is awesome! I'm stoked about this. Just gave it a shot in one of my apps and it worked great. Thanks for taking this on ❤️

robacarp:
@skojin wow, excellent

PR #1605, opened 2021-11-03
fix watch recompile message logic

skojin:
show "Recompiling..." when current build is aborted
and "Compiling..." if previous build was completed

this PR fix bug in previous PR https://github.com/luckyframework/lucky/pull/1604

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -153,7 +153,7 @@ module LuckySentry
     end
 
     def restart_app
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
ah! Good catch. Eventually I'd like to come back to https://github.com/luckyframework/lucky/pull/1067 and have it showing the spinner like we do in other areas. I think I'll open an issue to revisit that now that this is done

skojin:
I like the spinner

PR #1610, opened 2021-11-12
Adding a request_id

### LINKED ISSUE #950: Disjointed logging with lots of requests
If you have a page that takes a while to load, but also a bunch of async ajax calls on that page, you may see some disjointed logging in development where the calls come in at different times.

Something similar to this
```text
GET /favicon.ico
 ▸ Sent 200 (1.33ms)
 ▸ Sent 200 (721.66ms)
 ▸ Query   SELECT * FROM users

GET /api/reports
GET /dashboard
```

This is most likely because each request is done in it's own thread, and the logging is done on that thread. One solution could be to add a request ID to each log so you can match up which lines go to which request. 

ref: https://gitter.im/luckyframework/Lobby?at=5daab34be3646f24c75192ee

jwoertink:
## Purpose
Fixes #950

## Description
Each HTTP request can now have a unique identifier that you can use for tracking that request. Maybe store it in the DB, or use it in logging.

A new handler was added to assign the new `context.request_id`. This handler comes with a config for setting it however you want. More than likely, you'll just use `UUID.random.to_s`, but really you can set it however you want.

```crystal
# config/server.cr

Lucky::RequestIdHandler.configure do |settings|
  settings.set_request_id = ->(context : HTTP::Server::Context) {
    UUID.random.to_s
  }
end
```

Once you add this, the pretty log formatter adds the request ID in. It looks like this

Screen Shot 2021-11-11 at 7 39 52 PM

If you don't want the request id generated, or logged, the `set_request_id` is nil by default, and it all will be ignored. This also means that you can opt in per environment. So if you needed something different in production, or whatever... 

There are still a few questions I have unanswered...
* Should we log this on every single line? I only added it to the two lines I cared about (See #950 )
* Should we add this to all of the different log formatters?

## Checklist


src/lucky/pretty_log_formatter.cr
```diff
@@ -54,6 +54,12 @@ struct Lucky::PrettyLogFormatter < Dexter::BaseFormatter
         arrow.colorize.dim
       end
     end
+
+    private def add_request_id : Nil
+      if id = local_context["request_id"]
+        io << " (#{id.colorize.dim})"
```
jwoertink, line 60: What would be awesome here is if we could colorize the `id` to be a different color each time it changed... So when you look at the terminal, you see all the blue IDs or all the red ones and can easily tell which goes where... but I have no clue how we can do that unless the color was set much much higher and globally

robacarp: The color is an interesting idea. Here's one way to accomplish that. At first spin I had to pull in `big` -- I feel like this could also just use i64 by splicing the uuid in some way before converting it to an int.

```crystal
require "uuid"
require "big"
require "colorize"

uuid = UUID.random
hex = uuid.hexstring.to_big_i(16) 
color = (hex % 256).to_u8
puts "uuid: #{uuid.to_s.colorize(Colorize::Color256.new(color))}"
```

jwoertink: oh snap! That's awesome 😂  I'm gonna try that.

jwoertink: Yup! Looks like `hex = uuid.hexstring[0..6].to_i64(16)` works too. Not sure if the slice needs to be larger or not, but that seems to work...

robacarp: You're certainly discarding a lot of the entropy...which would be a problem if this was about anything more than just colors on a screen 😆 :shipit:

matthewmcgarvey (approved):

src/lucky/pretty_log_formatter.cr
jwoertink: Just realized that we can't guarantee the `id` will be UUID. With this setup, you have the ability to make the request ID anything you want as long as it's a string. It was a nice idea though 😂

jwoertink:
Ok, I'm gonna just merge this in as-is. It's easy enough to add more to it for the following release if we want.

PR #1611, opened 2021-11-13
Add previous_page and next_page methods to paginator

wout:
## Purpose
Adds `previous_page` and `next_page` methods to `Lucky::Paginator`.

(Sorry for not opening an issue first, but this seems like such an obvious addition).

## Description
Currently, there are only the `path_to_next` and `path_to_previous` methods, returning a path including all query params. There's no way to get the next/previous page _number_ other than calculating it based on the values returned by the `paginator.page` and `paginator.total` methods.

I'm working on an API and I need pagination with the full url and without any additional query parameters that may be present. To build the url, I need the previous and next page numbers rather than the paths.

## Checklist

### DIFF
--- a/spec/paginator/paginator_spec.cr
+++ b/spec/paginator/paginator_spec.cr
@@ -92,6 +92,24 @@ describe Lucky::Paginator do
     end
   end
 
+  describe "#previous_page" do
+    it "returns the next page" do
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Looks good to me!

matthewmcgarvey (approved):

PR #1614, opened 2021-11-17
when kill build process on code change, kill child process too

### LINKED ISSUE #1613: Compiler get OutOfMemory on many recompilations
**Describe the bug**
When we kill compiler on code change, the child compiler process not died
because we kill parent `sh` process started with `Process.new(command, shell: true)`
and if change code to quickly we get OutOfMemory error

**To Reproduce**
Change code and save it, to get many "Recompiling..." in console

**Expected behavior**

**Screenshots/code**
one `/bin/sh` and many child `asdf exec crystal`
Screen Shot 2021-11-17 at 16 51 46

skojin:
fixes #1613

kill child process explicitly with `kill $(ps -o pid= --ppid $$)` snippet from https://unix.stackexchange.com/a/124148

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -161,7 +161,11 @@ module LuckySentry
 
     private def stop_all_processes
       @build_processes.each do |process|
-        process.terminate unless process.terminated?
+        unless process.terminated?
+          # kill child process, because we started build process with shell option
+          Process.run("pkill -P #{process.pid}", shell: true)
+          process.terminate
+        end
       end
       @app_processes.each do |process|
         process.terminate unless process.terminated?


jwoertink:
Nice! Thanks for finding and fixing this. Do you happen to know the cross-platform compatibility of this? My guess is Alpine is probably the only one that may have an issue; however, I think we have some other Alpine issues anyway. If that's the case, then not a huge deal.

skojin:
mac fail on kill `ps -o pid= --ppid PID` command

jwoertink:
hmm, yeah, looks like it fails on mac. Also, it looks like the `-o` flag is for display? 

According to `man ps`

> -o      Display information associated with the space or comma separated list of keywords specified.

Since you're not actually passing any keywords, maybe that flag isn't needed?

skojin:
`pkill -P 16864` works on ubuntu and mac

jwoertink (approved):
Nice work! This looks better. That random cookie jar failure is unrelated. I've seen that fail randomly sometimes.

PR #1616, opened 2021-11-20
Replace avram parsing with ParamParser

### LINKED ISSUE #1383: Lucky depends on Avrams type extensions for param parsing

matthewmcgarvey:
## Purpose

Fixes #1383 

## Description

Where we were relying on Avram's parsing, this adds a new `Lucky::ParamParser` that handles it a bit differently than Avram does. Instead of adding methods to the types, there's one module with methods (with the same name) for the various types. Logically, though, it works the same. This new way is a bit nicer in that everything for parsing is in one place and not spread out in multiple files.

I did _not_ include parsing for Enums and JSON. Unfortunately it is not possible to have a method generic for all enums and I didn't think JSON was necessary though I think it would be easy enough to add.

I have a potential solution for the enum thing though. What if `param` could be provided a block that gets the string value and it returns the parsed value matching their variable or nil if it couldn't be parsed?

```crystal
param role : User::Role do |param|
  User::Role.parse?(param)
end
```

Just a thought if we feel like it's absolutely necessary to support that.

## Checklist

### DIFF
--- a/spec/lucky/param_parser_spec.cr
+++ b/spec/lucky/param_parser_spec.cr
@@ -1,0 +1,127 @@
+require "../spec_helper"
+
+enum TimeComponent
+  Year
+  Month
+  Day
+  Hour
+  Minute
+  Second
+end
+
+struct FormattedTime
+  property label : String
+  property value : String
+  property components : Array(TimeComponent)
+
+  def initialize(@label : String, @value : String, excluded_components : Array(TimeComponent) = [] of TimeComponent)
+    @components = [
+      TimeComponent::Year,
+      TimeComponent::Month,
+      TimeComponent::Day,
+      TimeComponent::Hour,
+      TimeComponent::Minute,
+      TimeComponent::Second,
+    ] - excluded_components
+  end
+end
+
+macro numbers_tests(klass, input, output)
+  describe "parse {{ klass }}" do
+    it "turns string into number" do
+      Lucky::ParamParser.parse({{ input }}, {{ klass }}).should eq({{ output }})
+    end
+
+    it "returns nil if param not number" do
+      Lucky::ParamParser.parse("abc", {{ klass }}).should be_nil
+    end
+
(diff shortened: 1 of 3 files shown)


spec/lucky/param_parser_spec.cr
```diff
@@ -0,0 +1,127 @@
...
+enum TimeComponent
+  Year
+  Month
+  Day
+  Hour
+  Minute
+  Second
+end
+
+struct FormattedTime
```
matthewmcgarvey, line 12: All of this was directly taken from Avram https://github.com/luckyframework/avram/blob/9a5a921bb05cc3031d9436d2fdc6927606314876/spec/type_extensions/time_spec.cr#L3-L27

@stephendolan saved me from having to figure out all these tests myself 😅

src/lucky/routable.cr
```diff
@@ -479,17 +479,17 @@ module Lucky::Routable
         {% end %}
       end
 
-      result = {{ base_type }}.adapter.parse(val)
+      result = Lucky::ParamParser.parse(val, {{ base_type }})
```
matthewmcgarvey, line 482: This is the line where we are switching from Avram code to this new module

jwoertink (approved):
Nice work!

PR #1620, opened 2021-11-22
Extract lucky_avram and remove avram dependency

matthewmcgarvey:
This is one of the last parts to making Avram optional in Lucky.
The code removed from here will be moved to avram https://github.com/luckyframework/avram/pull/772
The moved code will be accessible to apps by adding additional requires and the lucky_cli change can be found here https://github.com/luckyframework/lucky_cli/pull/711

This whole thing is a breaking change but in a minor way. Currently, having `require "lucky"` in your code pulled in all this code and also brought in avram. To get the same functionality, code will need to be updated to

```crystal
require "lucky"
require "avram"
require "avram/lucky"
```

And then the tasks will also need to be imported in your `tasks.cr` file

```crystal
require "avram/lucky/tasks"
```


jwoertink (approved):
Thanks for getting this started. I'm going to move forward with it now. This will allow people to use Lucky without a DB, or even use a different ORM if they'd like with a proper escape hatch. I think this will also make testing the lucky-avram bridge a lot easier with this code being closer to home.

PR #1622, opened 2021-11-27
Combine server bind_tcp and listen

matthewmcgarvey:
`HTTP::Server#listen` provides an overload that does exactly what we are doing on two lines.

https://crystal-lang.org/api/1.0.0/HTTP/Server.html#listen(host:String,port:Int32,reuse_port:Bool=false)-instance-method

### DIFF
--- a/src/lucky/base_app_server.cr
+++ b/src/lucky/base_app_server.cr
@@ -20,8 +20,7 @@ abstract class Lucky::BaseAppServer
 
   # :nodoc:
   def listen : Nil
-    server.bind_tcp host, port
-    server.listen
+    server.listen(host, port)
   end
 
   # :nodoc:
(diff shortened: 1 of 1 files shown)


src/lucky/base_app_server.cr
```diff
@@ -20,8 +20,7 @@ abstract class Lucky::BaseAppServer
 
   # :nodoc:
   def listen : Nil
-    server.bind_tcp host, port
-    server.listen
+    server.listen(host, port)
```
jwoertink, line 23: Why do we even define this here? We end up overriding it in the generated apps https://github.com/luckyframework/lucky_cli/blob/master/src/web_app_skeleton/src/app_server.cr.ecr#L28-L32 which includes the extra arg for the reuse_port... 🤔  Seems like we could just make this method abstract, then do this cleanup on the CLI repo. What do you think?

matthewmcgarvey: Good catch, I made the update here and opened https://github.com/luckyframework/lucky_cli/pull/699

jwoertink (approved):

PR #1638, opened 2022-01-04
Remove custom error on default_format macro

matthewmcgarvey:
Standard error message is just as helpful, if not more so, than our custom one.

## Before

Screen Shot 2022-01-03 at 10 31 56 PM

## After

Screen Shot 2022-01-03 at 10 32 06 PM


jwoertink (approved):
Yeah, that's fair enough 👍

PR #1639, opened 2022-01-04
Improve error message creating route links

### LINKED ISSUE #1637: Cleaner Lucky::Action.route errors
In the spirit of recent discussion about simplifying macros and making errors more clear, I started looking back into where we have existing complications. For instance in https://github.com/luckyframework/lucky/pull/1372 we added a new method with a compiler error to help people avoid confusion when using `Lucky::Action.route`. We are still seeing people get confused about it, so I went to see if I could help make it even better. I started looking into moving the method and was about to make a PR with it and decided just to double check, to run the error without the overload to get a screenshot of the default error message and show how my change made it "better". I was totally surprised to find out that the default error was WAY better than what we have.

## The Caveat

The problem is... the better default error message only appears if we remove `Lucky::Action.with`. If it's not taken out, the error message is still confusing.

## Example

This is the line with the error ("foo" shouldn't be there but technically it's duplicated arguments since `no_default` is the first argument)

```crystal
ParamsWithDefaultParamsLast.route("foo", no_default: "Yay!")
```

### Specifying default incorrectly
<details>
<summary>Current</summary>
<br>
Screen Shot 2022-01-03 at 9 56 18 PM
</details>

<details>
<summary>Without custom `Lucky::Action.route` compiler error</summary>
<br>
Screen Shot 2022-01-03 at 9 57 11 PM
</details>

<details>
<summary>Without `Lucky::Action.with`</summary>
<br>
Screen Shot 2022-01-03 at 9 58 25 PM
</details>

There's multiple other ways to trigger a compilation error and all of them are clearly pointing at the correct line instead of within macro code. But are we willing to accept the cost? DUN-DUN-DUUUUN!!!

matthewmcgarvey:
Resolves #1637 

Check out the connected issue for screenshots. Instead of removing `Lucky::Action.with`, I set up the function to require the expected arguments the same way as `Lucky::Action.route` which then calls `route` with them. This means it is not a breaking change.

I did remove the compiler error if you call `Lucky::Action.with` with no arguments. You get an error that doesn't actually point you to where you called it.

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -336,23 +336,13 @@ module Lucky::Routable
       {% end %}
       anchor : String? = nil
         ) : Lucky::RouteHelper
+        \{% begin %}
         route(
-          {% for param in path_params %}
-            {% temp = param.gsub(/:/, "").id %}
-            {{ temp }}: {{ temp }},
-          {% end %}
-
-          {% for param in PARAM_DECLARATIONS %}
-            {{ param.var }}: {{ param.var }},
-          {% end %}
-
-          # optional path variables are nilable
-          {% for param in optional_path_params %}
-            {% temp = param.gsub(/^\?:/, "").id %}
-            {{ temp }}: {{ temp }},
-          {% end %}
-          anchor: anchor
+          \{% for arg in @def.args %}
+            \{{ arg.name }}: \{{ arg.internal_name }},
+          \{% end %}
         )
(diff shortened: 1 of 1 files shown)


matthewmcgarvey:
@jwoertink Yep, the only functionality removed was the custom error we had if you called `.with` with no args.

jwoertink (approved):
Let's do it 🚀

Issue #1025, opened 2020-03-05
Add 'previous_url' helper


paulcsmith:
> This depends on other work I still need to document, so please don't start it yet :)

This links to the path in the referer header or uses a fallback path.

```crystal
link_to_back "Go back", fallback: Users::Index
```

matthewmcgarvey:
I knew I had seen this before. I believe this functionality was added here: https://github.com/luckyframework/lucky/commit/6a8132dc999945440a17850d55731cffb83308a3

It's `redirect_back` instead of `link_to_back`

jwoertink:
The `link_to_back` would be different. This would generate `<a>` tag that goes to maybe `history.goback()` or whatever that thing is, or maybe whatever the referrer was. It would be used in pages. The `redirect_back` would be used in Actions.

paulcsmith:
Yeah it'd prob use the referrer and require a fallback. I've changed my mind on the API. I don't want a million link helpers. Also this is a lot less felxible because maybe you want a URL that goes back that will be used in a button or a JS handler?

So here is my new proposal:

```crystal
a "Back", href: previous_url(fallback: Users::Index)
```

So `previous_url` uses the referrer (pretty much just like `redirect_back` and requires a `fallback` argument. It just returns a String so can be used pretty much anywhere.

This can be added now since the page has access to `context` by default.

Thoughts? I think this would be super helpful and the API is a lot more flexible and super easy to implement (just one method!)

jwoertink:
Oh, I think I like that much better. You're right about the JS stuff.

PR #1641, opened 2022-01-06
Adds previous_url method to url_helpers

colinkey:
## Purpose
Adds the `previous_url` helper as described in #1025 

## Description
If the `Referer` header is set, this method should return the url from the header. A user should pass a Lucky::Action or Lucky::RouteHelper as a fallback in instances where the Referer header is not set. 

Functionally, it works very similar to the `redirect_back` method added in #1168 but is suitable to use within a Lucky::Page

## Checklist

### DIFF
--- a/spec/lucky/url_helpers_spec.cr
+++ b/spec/lucky/url_helpers_spec.cr
@@ -146,15 +146,39 @@ describe Lucky::UrlHelpers do
       end
     end
   end
+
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (approved):
Nice work 🎉

jwoertink (approved):
🙌  perfect! Thanks so much for contributing 😄

PR #1644, opened 2022-01-08
Update Lucky::BaseHTTPClient to make requests to app directly

matthewmcgarvey:
## Purpose

In Lucky apps, we recommend using `Lucky::BaseHTTPClient` to make request specs against your running application. It's a simple proxy to an underlying `HTTP::Client` that allowed using you actions to make requests. In the quest to make our specs as quick as possible, this removes the need to start up the actual server in a different thread for request specs. It takes the middleware of your app can calls them directly.

In a very unofficial test, I saw ~10% speed improvement from the request specs because of this.

This _might_ also aid us in doing something similar for flow specs.

In a normal app, you are generated an `ApiClient` that looks like this

```crystal
class ApiClient < Lucky::BaseHTTPClient
  def initialize
    super
    headers("Content-Type": "application/json")
  end

  def self.auth(user : User)
    new.headers("Authorization": UserToken.generate(user))
  end
end
```

and a separate setup file boots your app in a fiber.

With this change, only one line will be added to your `ApiClient` and you won't need your app booted anymore

```crystal
class ApiClient < Lucky::BaseHTTPClient
  app AppServer.new # <--- instructs the client to connect to the app directly

  def initialize
    super
    headers("Content-Type": "application/json")
  end

  def self.auth(user : User)
    new.headers("Authorization": UserToken.generate(user))
  end
end
```

### DIFF
--- a/spec/lucky/base_http_client_spec.cr
+++ b/spec/lucky/base_http_client_spec.cr
@@ -1,18 +1,10 @@
 require "../spec_helper"
 
-server = TestServer.new(test_server_port)
-
-spawn do
-  server.listen
-end
-
-Spec.before_each do
-  TestServer.reset
-end
-
 class HelloWorldAction < TestAction
+  accepted_formats [:plain_text]
+
   post "/hello" do
-    plain_text "unused"
+    plain_text "world"
   end
 end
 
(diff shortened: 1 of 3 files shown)


jwoertink (commented):
Nice! This looks pretty great. One question... What happens with Flow specs? Don't those need the running server in order to connect through the browser? I know you wanted to make some flow specs run without the need for the browser, but there'll still be cases where you might test javascript clicks and such. Maybe we just consider that to be the edge case with docs on how to boot the server for those?

matthewmcgarvey:
@jwoertink this has no connection to flow specs. The underlying selenium library makes all requests to update the browser. I'm thinking about adding functionality to flow using something like this, though.

jwoertink:
Right, but what I mean is if you disabled the running spec server, then Flow specs that need to visit a specific page like `localhost:5000/sign_in` wouldn't work, right? So for those cases, we couldn't suggest removing this thing
https://github.com/luckyframework/lucky_cli/blob/45f35290dae282610ea6aa5012b31861c792a03e/src/web_app_skeleton/spec/setup/start_app_server.cr.ecr#L1-L5

I'm still down for this change. I just want to make sure it's clear to people that there may still be cases where you'll need the running server. When we generate new apps, we may still have to include that bit of code.

jwoertink (approved):

matthewmcgarvey:
O I see what you mean! Yes, you can't remove that until we get flows using an in-memory integration as well.

PR #1646, opened 2022-01-11
Add Crystal 1.3.0 to the CI

jwoertink:
## Purpose
Now that Crystal 1.3.0 is out, make sure we test against that. I'm also bumping the Crystal versions to be the last in those version series. I figure it's probably better to test against 1.2.2 than 1.2.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,14 +15,12 @@ jobs:
           - shard.yml
         crystal_version:
           - 1.0.0
-          - 1.1.1
-          - 1.2.2
           - 1.3.0
         experimental:
           - false
         include:
           - shard_file: shard.edge.yml
-            crystal_version: 1.1.1
+            crystal_version: 1.3.0
             experimental: true
           - shard_file: shard.yml
             crystal_version: nightly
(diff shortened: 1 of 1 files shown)


robacarp (approved):

.github/workflows/ci.yml
```diff
@@ -15,13 +15,14 @@ jobs:
           - shard.yml
         crystal_version:
           - 1.0.0
-          - 1.1.0
-          - 1.2.0
+          - 1.1.1
+          - 1.2.2
+          - 1.3.0
```
matthewmcgarvey, line 20: 🙏 Please, sir...can we remove `1.1.1` and `1.2.2` from these check?

grepsedawk (approved):

.github/workflows/ci.yml
jwoertink: 😂  yeah, this is what I was asking about before. So do we only keep 2 in here plus the nightly? Whatever the lowest version we support, and whatever the highest version is? In this case 1.0.0, and 1.3.0?

matthewmcgarvey: That's my opinion. I think it's unlikely we'd break anything on the in-between versions and it will cut down on CI time

jwoertink: oh actually... for the formatting we only really need to include latest. I'm gonna drop the 1.0 here too

PR #1650, opened 2022-01-17
Turbolinks legacy module

rmarronnier:
Follow-up / Replace https://github.com/luckyframework/lucky/pull/1648
Following various discussions about the fate of Turbolinks support in Lucky, this PR extract Turbolinks specific code that was hardcoded in `redirectable.cr` to a single module file `turbolinks_action_support.cr`.

This module can be included into a `Lucky::Action` to regain Turbolinks support.

This is a necessary step for Lucky to be able to support alternatives (like Turbo Drive for example) without monkeypatching `redirectable.cr`

Right now, in the current state of this PR, an existing Lucky project needs to include `Lucky::TurboLinksActionSupport` into every related action or into an "umbrella" action to keep Turbolinks support like this spec example :

```Crystal
class RedirectTurboLinksAction < TestAction
  include Lucky::TurboLinksActionSupport

  get "/redirect_turbolinks_test" do
    plain_text "does not matter"
  end
end
```

I'm open to more elegant / efficient refactoring / naming / ...  changes.

## Checklist


src/lucky/redirectable.cr
```diff
@@ -122,36 +122,20 @@ module Lucky::Redirectable
...
 
   # :nodoc:
   def redirect(to page_instead_of_action : Lucky::HTMLPage.class, **unused_args)
     {% raise "You accidentally redirected to a Lucky::HTMLPage instead of a Lucky::Action" %}
   end
 
-  private def store_turbolinks_location_in_session(path : String)
-    cookies.set(:_turbolinks_location, path).http_only(true)
-    # this cookie read at Lucky::RedirectableTurbolinksSupport
+  private def special_ajax_redirect(path, status) : Lucky::TextResponse
```
matthewmcgarvey, line 133: This is really not a special ajax redirect by default. It's only on the turbolinks module that the ✨ added. Maybe rename this and the overrides just `handle_redirect` or something to that effect.

src/lucky/action.cr
```diff
@@ -1,5 +1,7 @@
 require "./*"
 
+# require "./support/turbolinks/action/*"
```
matthewmcgarvey, line 3: Can be removed?

matthewmcgarvey (changes requested):
This is looking good! Really glad to see that you were able to move all the turbolinks stuff into the same module. My concern is that I just _know_ people will miss that they need to include the module. I think I'd prefer that we still include the turbolinks module for this next release. The new turbo module, though, would just override the `special_ajax_redirect` method (or whatever it might be renamed to) and have an empty method for `set_turbolinks_location_header_from_session` so that if the turbo module and turbolinks module are present, the turbo one takes precedent. And I think it would probably be best to switch to including the turbo module in generated projects (so in LuckyCli) instead of the `Lucky::Action` so that this situation can be solved outside of updating lucky directly in the future. Make sense?

src/lucky/redirectable.cr
rmarronnier: Yeah, I wasn't very sure about the naming of this... And now I think to try to only overload a subpart of the `def redirect(to path : String, status : Int32 = 302) : Lucky::TextResponse` might obscure things later on. Dumb code is better than smart code or something :-)

So what do you think about (removing `special_ajax_redirect` and overloading the whole `redirect` method) :

``` Crystal
# In redirectable.cr

  def redirect(to path : String, status : Int32 = 302) : Lucky::TextResponse
    # flash messages are not consumed here, so keep them for the next action
    flash.keep
    Lucky::TextResponse.new(context, "", "")
  end
  ```
  And
  
``` Crystal
# In turbolinks_action_support.cr

# Overloads Lucky::Redirectable redirect's method
  def redirect(to path : String, status : Int32 = 302) : Lucky::TextResponse
    # flash messages are not consumed here, so keep them for the next action
    flash.keep
   if ajax? && request.method != "GET"

      # do not enable form disabled elements for XHR redirects, see https://github.com/rails/rails/pull/31441
      context.response.headers.add "X-Xhr-Redirect", path

      Lucky::TextResponse.new(context,
        "text/javascript",
        %[Turbolinks.clearCache();\nTurbolinks.visit(#{path.to_json}, {"action": "replace"})],
        status: 200)
    else
      if request.headers["Turbolinks-Referrer"]?
        store_turbolinks_location_in_session(path)
      end
      # ordinary redirect
      Lucky::TextResponse.new(context, "", "")
    end
    Lucky::TextResponse.new(context, "", "")
  end

private def store_turbolinks_location_in_session(path : String)
    cookies.set(:_turbolinks_location, path).http_only(true)
  end

  macro included
    before set_turbolinks_location_header_from_session
  end

  private def set_turbolinks_location_header_from_session
    if turbolinks_location = cookies.get?(:_turbolinks_location)
      cookies.delete(:_turbolinks_location)
      # change browser address bar at last request, see https://github.com/turbolinks/turbolinks#following-redirects
      response.headers["Turbolinks-Location"] = turbolinks_location
    end
    continue
  end
  ```

src/lucky/action.cr
rmarronnier: Of course, good catch !

rmarronnier:
> This is looking good! Really glad to see that you were able to move all the turbolinks stuff into the same module. My concern is that I just _know_ people will miss that they need to include the module. I think I'd prefer that we still include the turbolinks module for this next release. The new turbo module, though, would just override the `special_ajax_redirect` method (or whatever it might be renamed to) and have an empty method for `set_turbolinks_location_header_from_session` so that if the turbo module and turbolinks module are present, the turbo one takes precedent. And I think it would probably be best to switch to including the turbo module in generated projects (so in LuckyCli) instead of the `Lucky::Action` so that this situation can be solved outside of updating lucky directly in the future. Make sense?

It makes perfect sense : the "double-overriding" was the missing piece of the puzzle for a non-breaking change path. Thank you  :-)

src/lucky/redirectable.cr
matthewmcgarvey: I think that's fine 👍

rmarronnier:
I pushed the discussed changes and I renamed the module back to `Lucky::RedirectableTurbolinksSupport`, which is more accurate (doesn't change behavior outside of `redirectable.cr`).

Ready for (final ?) review.

matthewmcgarvey (approved):
This PR got tiny! Nice work 🎉

PR #1651, opened 2022-01-17
Allowing for rounding up on hours.

### LINKED ISSUE #1649: time_ago_in_words always rounds down
```crystal
puts time_ago_in_words(110.minutes.ago) #=> an hour
```

It currently returns on hours because this is technically only 1 hour, but it ignores the other 50minutes. I would expect this to return "almost 2 hours" instead.

https://github.com/luckyframework/lucky/blob/e1a6adf8552676884e67528293d3142032387b4b/src/lucky/page_helpers/time_helpers.cr#L25

We do already make a special concession for minutes past 45 where we say "about an hour" if it's been 46minutes or more

https://github.com/luckyframework/lucky/blob/e1a6adf8552676884e67528293d3142032387b4b/src/lucky/page_helpers/time_helpers.cr#L77-L79

We could probably do the same here that if it's past X hour and 45 minutes, then just round up.

russ:
Fixes #1649

## Purpose
Allowing for a more expected return when a time is closer to the higher hour mark.

## Description
Just a small tweak to see if the time is closer to the next hour.

## Checklist


jwoertink (approved):
Nice!

PR #1661, opened 2022-01-31
More accurately parse json values for params

matthewmcgarvey:
## Purpose

In the cases where the value of a json field is a JSON object, it was inaccurately returning the string representation of the hash instead of the JSON object. So, instead of calling `.to_s` on `JSON::Any` we call `.to_json` with one caveat and that is String values. Calling `.to_json` on a string wraps the string in escaped quotes.


jwoertink (approved):
:+1:

PR #1665, opened 2022-02-10
Add changelog for X-Forwarded-Proto matching

grepsedawk:
I'm not really sure how you do changelog items, but this should be noted as a "more notable change than just a fixup"


jwoertink (approved):
Normally I just handle the changelog myself when I do the release. I'll just merge this in for now, but you won't need to worry about adding these changes for future PRs :smiley:

PR #1674, opened 2022-02-19
reworks the message verifier to remove the magic -- character.

### LINKED ISSUE #1595: Message Verifier fails when data has --
https://github.com/luckyframework/lucky/blob/74c049f7f8e1c31656310855b48fb4754d2dec23/src/lucky/support/message_verifier.cr#L14

https://github.com/luckyframework/lucky/blob/74c049f7f8e1c31656310855b48fb4754d2dec23/src/lucky/support/message_verifier.cr#L28

It's a super rare case, but `signed_message` could look like 

```
MTYzNThNzIyZi04YWODczYi1lZTkyNTJmMjcwMmE%3D--bNCi4Gn--fjh61AB9FVLZrEgzZN8c2DgTrw%3D
```

and when it does, the `digest` is wrong. We can account for this by passing in `2` as a second argument to `split`

`split("--", 2)` will split by the first occurrence of `--`, and only split in to 2 parts. [ref](https://crystal-lang.org/api/1.2.1/String.html#split(limit:Int32?=nil):Array(String)-instance-method)

jwoertink:
## Purpose
Fixes #1595

## Description
We were using a magic `--` to split the data in the token generated by the message verifier. In some rare instances, the data on the left or the right of the magic character could also contain `--` themselves. This caused issues when splitting on that. 

This fix removes the magic character completely and just uses some JSON instead.

/cc @robacarp 

## Checklist
* [ x - Lucky builds on docker with `./script/setup`
* [ x - All builds and specs pass on docker with `./script/test`

### DIFF
--- a/spec/lucky/support/message_verifier_spec.cr
+++ b/spec/lucky/support/message_verifier_spec.cr
@@ -6,4 +6,10 @@ describe Lucky::MessageVerifier do
     signed_message = verifier.generate("abc123")
     verifier.verified(signed_message).should eq("abc123")
   end
+
+  it "still works with some more complext data" do
+    verifier = Lucky::MessageVerifier.new("mFClXIwWbxfqJwnJ/rXXFK02kO5z8wY2P8mjozsEQDk=", :sha256)
+    signed_message = verifier.generate("#{Time.utc(2022, 1, 15, 10, 12).to_unix}:some_special_dude@hotmail.com:b211cbb5-3cc0-475a-9ebe-45f3fd2fe650")
+    verifier.verified(signed_message).should eq("1642241520:some_special_dude@hotmail.com:b211cbb5-3cc0-475a-9ebe-45f3fd2fe650")
+  end
 end
--- a/src/lucky/support/message_verifier.cr
+++ b/src/lucky/support/message_verifier.cr
@@ -11,7 +11,9 @@ module Lucky
     end
 
     def verified(signed_message : String) : String?
-      data, digest = signed_message.split("--", 2)
+      json_data = ::Base64.decode_string(signed_message)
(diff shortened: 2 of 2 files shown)


matthewmcgarvey (commented):
A spec needs to be added for a signed message encrypted the old way and show what the failure scenario is there. I'm assuming that on rollout of this change _everyone_ is going to be signed out from everybody's app which we've done before but I don't think it was ideal.

If we really wanted to, we could try both ways to decrypt but only encrypt in the new way so that it would be a smoother release.

robacarp:
This is great @jwoertink. I agree with @matthewmcgarvey that detecting and rolling forward auth tokens would smooth out the release. Box it into a method that can be marked as deprecated and not forgotten. I'm sure even then some people will get an invalid token error when it is _eventually_ removed, but that would help a ton.

jwoertink:
This will now try to decode legacy tokens, and then basically fallback to new tokens. I've added the deprecation notice:

```
❯ crystal spec/
In src/lucky/support/message_verifier.cr:22:22

 22 | data, digest = legacy_verified(signed_message)
                     ^--------------
Warning: Deprecated Lucky::MessageVerifier#legacy_verified. Legacy token verification will be removed in the next release.

A total of 1 warnings were found.
```

robacarp (approved):

matthewmcgarvey (changes requested):
Could we swap it to new way first? It seems like it'd be a lot of errors for apps to start having when they've upgraded

jwoertink:
Swapping the logic was a little weird, but specs passed. After we remove the deprecated legacy stuff, we should be able to clean this up more. Let me know if you guys have other suggestions for this.

PR #1675, opened 2022-02-19
Added remote_ip to request object

### LINKED ISSUE #1643: Error: undefined method 'address' for Socket::UNIXAddress (compile-time type is Socket::Address+)
**Describe the bug**
Type mismatch for `Socket::UNIXAddress`

**To Reproduce**
`ip_address = context.request.remote_address.try(&.address) || "N/A"`

from https://luckyframework.org/guides/http-and-routing/http-handlers#remoteiphandler

**Screenshots/code**
image

 - Lucky version 0.29.0
 - Crystal version Crystal 1.2.2 (2021-11-10)
 - OS: MacOS

jwoertink:
## Purpose
Fixes #1643

## Description
If you need access to the client's IP address, using `remote_address` can be troublesome since it casts to `Socket::Address`, and not all `Socket::Address` subclasses contain an `address` method. So you usually end up with some code like this:

```crystal
context.request.remote_address.as?(Socket::IPAddress).try(&.address) || "127.0.0.1"
```

But now you can just do

```crystal
context.request.remote_ip || "127.0.0.1"
```

In addition, I've also added a new habitat config option `ip_header_names` option. This allows you to configure the header names of where the client IP may come from. e.g. `X-Real-Ip`, `Forwarded-For`, etc...

## Checklist

### DIFF
--- a/src/charms/request_extensions.cr
+++ b/src/charms/request_extensions.cr
@@ -1,0 +1,9 @@
+class HTTP::Request
+  # This is an alternative to `remote_address`
+  # since that casts to `Socket::Address`, and not all
+  # subclasses have an `address` method to give you the value.
+  # ```
+  # request.remote_address.as?(Socket::IPAddress).try(&.address)
+  # ```
+  property remote_ip : String?
+end
--- a/src/lucky/remote_ip_handler.cr
+++ b/src/lucky/remote_ip_handler.cr
@@ -10,17 +10,26 @@
 class Lucky::RemoteIpHandler
   include HTTP::Handler
 
+  Habitat.create do
+    setting ip_header_names : Array(String) = ["X_FORWARDED_FOR"]
+  end
+
   def call(context)
     context.request.remote_address = fetch_remote_ip(context)
+    context.request.remote_ip = context.request.remote_address.as?(Socket::IPAddress).try(&.address.presence)
     call_next(context)
   end
 
   private def fetch_remote_ip(context : HTTP::Server::Context) : Socket::Address?
     request = context.request
 
-    if x_forwarded = request.headers["X_FORWARDED_FOR"]?.try(&.split(',').first?).presence
+    remote_ip = settings.ip_header_names.compact_map { |header|
+      request.headers[header]?.try(&.split(',').first?).presence
+    }.first?
+
+    if remote_ip
       begin
-        Socket::IPAddress.new(x_forwarded, 0)
+        Socket::IPAddress.new(remote_ip.to_s, 0)
       rescue Socket::Error
         # if the x_forwarded is not a valid ip address we fallback to request.remote_address
         request.remote_address
--- a/spec/lucky/remote_ip_handler_spec.cr
+++ b/spec/lucky/remote_ip_handler_spec.cr
@@ -9,6 +9,7 @@ describe Lucky::RemoteIpHandler do
 
       run_remote_ip_handler(context)
       context.request.remote_address.should eq nil
+      context.request.remote_ip.should eq nil
     end
 
     it "returns the X_FORWARDED_FOR address" do
@@ -18,7 +19,9 @@ describe Lucky::RemoteIpHandler do
       context = build_context(request)
 
       run_remote_ip_handler(context)
+      context.request.remote_address.should be_a(Socket::IPAddress)
       context.request.remote_address.should eq Socket::IPAddress.new("1.2.3.4", 0)
+      context.request.remote_ip.should eq("1.2.3.4")
     end
 
     it "returns request.remote_address if X_FORWARDED_FOR is not valid" do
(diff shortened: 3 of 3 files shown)


src/lucky/remote_ip_handler.cr
```diff
@@ -10,17 +10,26 @@
 class Lucky::RemoteIpHandler
   include HTTP::Handler
 
+  Habitat.create do
+    setting ip_header_names : Array(String) = ["X_FORWARDED_FOR"]
```
jwoertink, line 14: Not completely sold on this name, but naming is hard....

matthewmcgarvey (approved):

src/charms/request_extensions.cr
```diff
@@ -0,0 +1,9 @@
+class HTTP::Request
+  # This is an alternative to `remote_address`
+  # since that casts to `Socket::Address`, and not all
+  # subclasses have an `address` method to give you the value.
+  # ```
+  # request.remote_address.as?(Socket::IPAddress).try(&.address)
+  # ```
+  property remote_ip : String?
```
robacarp, line 8: Since this is a string, I wonder if there's a way this optional can be avoided? 

For the sake of argument, what if it defaulted to "0.0.0.0" if detection failed? What about an empty string?

src/lucky/remote_ip_handler.cr
```diff
@@ -10,17 +10,26 @@
...
     call_next(context)
   end
 
   private def fetch_remote_ip(context : HTTP::Server::Context) : Socket::Address?
     request = context.request
 
-    if x_forwarded = request.headers["X_FORWARDED_FOR"]?.try(&.split(',').first?).presence
+    remote_ip = settings.ip_header_names.compact_map { |header|
+      request.headers[header]?.try(&.split(',').first?).presence
+    }.first?
```
robacarp, line 28: I see where you're going with the list of header names, but in practice this will never need to be a list for any one deployment. Here you're iterating and running the entire map on every request.

Is there a way to move this decision so it can be run on demand rather than on every request?

Or perhaps the logic of determining what header might actually contain the value could be done just once?

robacarp (changes requested):
In addition to my note about performance, there are some security implications to this magic. Remote IP detection isn't "guaranteed" or "safe" by any means, but that doesn't stop tons of project from using it to safeguard their admin dashboards, etc.

A safe load balancer implementation of the forwarded-for paradigm ensures that the header is wiped for all traffic, and then sets it correctly if the remote ip address is usable. Nobody should be able to set the header _besides_ the load balancer.

- If Lucky always (or even by default) searches for these headers, then it becomes trivial to spoof a request address by simply injecting a header to all requests. 
- If Lucky checks multiple headers, then correctly configuring a load balancer becomes more difficult. All possible headers need to be wiped, then the one "safe" header can be set.
- If the application isn't actually deployed behind a load balancer, Lucky should not be checking these headers at all.

jwoertink:
Instead of being an array, should it just be the single string value for the header? It would make sense that once you figure out which header is being passed back, it'll always be that same one. I've ran in to an issue where the LB config has changed, but I guess in that case, you'd just update the app to use a different header and do a new deploy.

src/charms/request_extensions.cr
jwoertink: I think we can do that. Not sure if "0.0.0.0" is ok default... but at least an empty string just to avoid it being nilable.

jwoertink:
@robacarp which part are you thinking for the opt-in?

The `remote_address` has been in Lucky for a few versions now, and it's always used the `X_FORWARED_FOR` header. So that would stay as the default, but just give you the option to change it. The only new bit would be the `remote_ip` which would just be an extra property on request.

robacarp:
My preference would be to have the whole thing be opt-in, but I understand why that might not be a desirable change to make. There are a bunch of ways an incorrectly configured x-forwarded-for system can be abused, and having a web server which reads the header without the proxy server in front of it is one of them.

I spent some time poking at what Rack does for this. A notable difference is that it doesn't return just one value, but an array of values. They take the added precaution of filtering out a default list of IP ranges, too. There's a lot more code [in this rack module](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L153-L158) than I expected.

jwoertink:
Ah, ok. I see what you're saying. In that case, we'd need to remove this 

https://github.com/luckyframework/lucky_cli/blob/5649c9e78c9c359710347ef0e91fa87c04072ea6/src/web_app_skeleton/src/app_server.cr.ecr#L6

Then we could add docs on adding it in if you need. I would eventually like to make it more robust like they do with the Rack stuff. IPSpoofing checks and all that...

jwoertink:
Got sent this article: https://adam-p.ca/blog/2022/03/x-forwarded-for/

One interesting thing is we currently have this line:
https://github.com/luckyframework/lucky/blob/4cb2108f3e05acaadc6c948a07b2838154e8871a/src/lucky/remote_ip_handler.cr#L21

which uses the first (left most) IP, but according to this article, we should be using the last (right most)
> When deriving the “real client IP address” from the X-Forwarded-For header, use the rightmost IP in the list.

It also mentions to use the last instances of the header, which I think Crystal is doing..
> make sure to use the last instance of that header.

I think maybe we can include a link to this article in the comments. We make the default `X-Forwarded-For`, with the option to change it, but it's really up to you to figure out how to setup your reverse proxy and/or load balancer to pass back what you need. Then this will just handle the 80% use case.

matthewmcgarvey (approved):

PR #1678, opened 2022-03-16
Add Lucky::CookieJar#[] and #[]=

grepsedawk:
Adds easier access to cookies by adding the typical key/value syntax
for cookies get/set operations

In the future, this could potentially even implement all of Enumerable({K, V}), but
these methods are super helpful as is to provide hash-like access to cookies


matthewmcgarvey (changes requested):
The types need to be specified. Otherwise looks good

grepsedawk:
> The types need to be specified

Ah cool, I thought being a direct alias the implicit types would be fine.
Will change that.

matthewmcgarvey (approved):

PR #1682, opened 2022-04-06
UPGRADE_NOTES for 0.30.0

jwoertink:


jwoertink:
alright, so this feels weird, but there's actually not much you need to do to upgrade this... I think this might be all. I want to wait until after @stephendolan has a chance to update LuckyDiff before merging

PR #1690, opened 2022-04-19
use shards to automate setting of VERSION

stakach:
no more manual management of version constant

### DIFF
--- a/src/lucky/version.cr
+++ b/src/lucky/version.cr
@@ -1,3 +1,7 @@
 module Lucky
-  VERSION = "0.30.1"
+  macro set_version
+    VERSION = {{ `shards version "#{__DIR__}"`.chomp.stringify.downcase }}
+  end
+
+  set_version
 end


jwoertink:
You know, I actually did this on LuckyEnv https://github.com/luckyframework/lucky_env/blob/7b10812688ec682fb86e9eeae640a63bc4500fd4/src/lucky_env.cr#L6 but I was curious as to whether this increases compile time or not... Since the Lucky ecosystem has so many shards, if they all did this, does that increase the compile time to a noticeable difference? Or is it like an extra 2ms for all of them?

stakach:
I changed to using the macro to maintain compatibility with [wyhaines/defined.cr](https://github.com/wyhaines/defined.cr#version-checks)
Definitely not a noticeable difference in compile time, with multiple shards running this and OS caches still being hot the execution is extremely fast - although my dev laptop is a beast 😅

jwoertink (approved):
Thanks for the PR!

PR #1725, opened 2022-08-25
just testing Vite

jwoertink:
Just using this so we have a branch to point to for now


jwoertink:
You can now use Vite with Lucky if you need as of this PR https://github.com/luckyframework/lucky/pull/1729

Issue #1724, opened 2022-08-22
Evaluate future of mix integration


jwoertink:
We originally integrated laravel-mix as default because when Lucky was first being built in 2017, webpack was the top choice, and mix provided really great configurations to wrap webpack.

Now in 2022, Laravel has deprecated mix https://laravel.com/docs/9.x/mix in favor of using Vite https://vitejs.dev/ This seems like a pretty decent choice seeing that it's using esbuild under the hood, and comes wrapped up with a lot of nice tooling like rollup and such. However, on my initial test of this, I was getting javascript compiler crashes, and found some of the integration docs to be very confusing.

I got this error with almost no javascript in the app. I let it sit for about ~5min and then I would get this.

```
// assets       | <--- JS stacktrace --->
// assets       |
// assets       | FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
// assets       |  1: 0xafedf0 node::Abort() [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  2: 0xa1814d node::FatalError(char const*, char const*) [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  3: 0xce795e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  4: 0xce7cd7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  5: 0xeb16b5  [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  6: 0xeb21a4  [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
// assets       |  7: 0xec0617 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/home/jeremy/.asdf/installs/nodejs/16.6.1/bin/node]
[...]

PR #1729, opened 2022-09-04
Experimenting with optional vite integration. 

jwoertink:
## Purpose
Ref #1724

## Description
This is an experiment to test if we can keep mix integration while allowing for vite to be used optionally. In theory, you would just need to update your `load_manifest()` call to point to `./public/manifest.json`, and a second `use_vite: true` arg. The actual setup for [Vite](https://vitejs.dev/) would still need to be done, but you can use https://github.com/jwoertink/vite_lucky/ as a sort of reference.

## Checklist

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -7,7 +7,7 @@ include RoutesHelper
 
 Pulsar.enable_test_mode!
 
-Lucky::AssetHelpers.load_manifest
+# Lucky::AssetHelpers.load_manifest
 
 Spec.before_each do
   ARGV.clear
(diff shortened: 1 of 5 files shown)


spec/spec_helper.cr
```diff
@@ -8,6 +8,7 @@ include RoutesHelper
 Pulsar.enable_test_mode!
 
 Lucky::AssetHelpers.load_manifest
+Lucky::AssetHelpers.load_manifest("./public/vite-manifest.json", use_vite: true)
```
jwoertink, line 11: In practice you would never run this more than once, but this was the only way I could test it since it happens at compile-time.

Issue #1652, opened 2022-01-22
route 


grepsedawk:
**Describe the bug**

**To Reproduce**

> Example:

> 1. Generate a Lucky project '...'
> 2. Generate a route with '...'
> 3. See error 'paste error message'

**Expected behavior**

**Screenshots/code**
If applicable, add screenshots/code samples to help explain your problem.

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock):
 - Crystal version (`crystal --version`):
 - OS:

**Additional context**

matthewmcgarvey:
<img width="661" alt="Screen Shot 2022-09-13 at 12 29 22 AM" src="https://user-images.githubusercontent.com/17329408/189816577-1b460e47-77aa-4887-86a2-99ac2ccfe50b.png">

Seems that the standard compiler error has changed now with Crystal v1.5.0.

Thinking about what would be expected and using my picture as an example... ideally it should point to the call to `link` saying that you can't pass `LinkHelpers::Show` and point at where `link` was called. I can't think of a way to accomplish that with the way things are right now. The most obvious think to reach for would be to add a `route` method to actions that have required parameters but I think that would have to provide a super generic error message and still not point to the line we want it to.

To think outside of the current implementation, I think we'd have to have three different interfaces.
1. `Lucky::RouteHelper` that's what you get when you provide parameters
2. `Lucky::Action` the base... params could be required or not
3. `Lucky::ParamsNotRequiredAction` layered onto the base `Lucky::Action`... params are not required

When generating the class, you'd extend the action with `Lucky::ParamsNotRequiredAction` if there are no required params. That module would give a `route` method that would be removed from `Lucky::Action`. At least in the `Lucky::LinkHelpers` module you'd replace the methods that take in `Lucky::Action` with `Lucky::ParamsNotRequiredAction`. That way, if you pass in an action that requires params without providing them, it would fail to compile exactly where we want it to. The only issue is, it wouldn't quite explain itself as we'd like so we'd likely have to provide an implementation of the methods with `Lucky::Action` and provide a clearer error message.

This is all just off the top of my head, so I have no idea if it'd even work. I'll look into it more as I'm interested in it right now.

PR #1734, opened 2022-09-13
Better compile time error for link helper usage and actions that require params

matthewmcgarvey:
## Purpose

Fixes https://github.com/luckyframework/lucky/issues/1652

## Description

This is an attempt at making link helper compile time errors better when passing in a `Lucky::Action` that needs params.

Screen Shot 2022-09-13 at 1 01 09 AM'

🙏 It's really late. I got the code working, but I hope you can see that the error message is a joke at the moment. Please help me with that. I am even wondering if we can access the action class in the error message to print out what is required? That might be a stretch too far though. Another thing I didn't look into is if this is useful anywhere else. I'd check anywhere we currently use `Lucky::Action.class`

### DIFF
--- a/spec/lucky/link_helpers_spec.cr
+++ b/spec/lucky/link_helpers_spec.cr
@@ -8,6 +8,10 @@ class LinkHelpers::Create < TestAction
   post "/link_helpers" { plain_text "foo" }
 end
 
+class LinkHelpers::Show < TestAction
+  get "/link_helpers/:id" { plain_text "foo" }
+end
+
 private class TestPage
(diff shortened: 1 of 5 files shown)


jwoertink:
love it. ship it! also, do better :joy:

jwoertink:
The code looks ok, but I would like to give this a shot with Crystal 1.6 first to see if the errors are going to change anyway. Actually... maybe it won't matter since it's not really the link that's wrong, but the action.. :thinking:  I'll think about an error message in the meantime.

PR #1736, opened 2022-09-27
Close the reponse after a redirect to avoid double redirects.

### LINKED ISSUE #1735: Investigate closing the response during redirect
Coming in Crystal 1.7, the HTTP Response will have a built-in `redirect` method https://github.com/crystal-lang/crystal/pull/12526 I noticed on this PR that after setting the status and header `close` is called.... I always see these IO errors in my Lucky apps and now I'm wondering if it's because we don't close the response after doing a redirect.

We can see about just adding a `context.response.close` after the status and header are added, then after Crystal 1.7 is out, we can slowly deprecate away the old code.

https://github.com/luckyframework/lucky/blob/e167dc59093742c4baa72038fbc2b682f3c36f70/src/lucky/redirectable.cr#L121-L126

jwoertink:
## Purpose
Fixes #1735 

## Description
Coming up in Crystal 1.7, HTTP response will have a redirect method built-in. That method calls `close` after setting the Location header. I think this might also fix this elusive bug https://forum.crystal-lang.org/t/help-solving-io-error/130 

## Checklist

### DIFF
--- a/spec/lucky/action_redirect_spec.cr
+++ b/spec/lucky/action_redirect_spec.cr
@@ -150,9 +150,7 @@ describe Lucky::Action do
     action.redirect to: "/somewhere"
     should_redirect(action, to: "/somewhere", status: 302)
 
-    expect_raises(IO::Error) do
-      action.redirect to: "/somewhere-else"
-    end
+    action.context.response.closed?.should eq(true)
   end
 end
 
--- a/spec/lucky/action_redirect_spec.cr
+++ b/spec/lucky/action_redirect_spec.cr
@@ -144,6 +144,16 @@ describe Lucky::Action do
     flash = Lucky::FlashStore.from_session(response.context.session)
     flash.success.should eq("Keep me!")
   end
+
+  it "closes the response after a redirect" do
+    action = RedirectAction.new(build_context, params)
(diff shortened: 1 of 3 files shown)


jwoertink:
I'm gonna hold this open for a day. That nightly should have worked, but maybe they need to push an update....

spec/lucky/action_redirect_spec.cr
```diff
@@ -144,6 +144,16 @@ describe Lucky::Action do
...
   end
+
+  it "closes the response after a redirect" do
+    action = RedirectAction.new(build_context, params)
+    action.redirect to: "/somewhere"
+    should_redirect(action, to: "/somewhere", status: 302)
+
+    expect_raises(IO::Error) do
+      action.redirect to: "/somewhere-else"
+    end
```
robacarp, line 155: This is a weird test methodology. It feels like it might lead to a false negative, where this test passes but for the wrong reason. Is there a hook to ask the action if the response is closed somehow?

jwoertink: I could do `action.context.response.closed?.should eq(true)`. :man_shrugging:

PR #1743, opened 2022-09-29
Redirect to fallback when referrer is current page

### LINKED ISSUE #1742: `#previous_url` stuck in a loop
**Describe the bug**
`previous_url` does not do the expected behavior when referrer is the current page itself.

**Scenario**
Let's say you're in `/users` and you clicked a button that sends a POST request to `/users/create` which redirects back to `/users`. The Referrer header would be set as "/users" causing the `previous_url` to not work as expected.

jcads:
## Purpose
fixes #1742 

## Description

## Checklist


jwoertink:
Nice! Thanks for the fix :tada:

PR #1761, opened 2022-11-15
Add macro to inline SVG files

wout:
## Purpose
Adds a macro to inline SVG files at compile time.

## Description
SVG logos and icons can be inlined with the `inline_svg` macro in pages and components:

```crystal
button
  inline_svg("buttons/round.svg")
end
```
This will:
- inline the SVG file from `src/svgs` at compile-time (path configurable with an annotation)
- strip the XML declaration, comments newlines and superfluous whitespace
- optionally (and by default) strip styling attributes so the icons can be styled with CSS (also configurable with an annotation)
- raise a compile-time exception when an SVG file is missing
- add a `data-inline-svg` attribute with the path of the icon (e.g. `<svg data-inline-svg="buttons/round.svg">`)

Icons can then be styled with:

```css
[data-inline-svg] {
  fill: none;
  stroke: #222;
  stroke-width: 2px;
}
```

Or:

```css
[data-inline-svg="buttons/round.svg"] {
  fill: #f03;
}
```

I realise Lucky does not use annotations a lot, but I think they are helpful in this case. The settings defined by the annotations are reasonable defaults, I think, but if necessary, users can jump in and override the defaults.

This PR implements #1760.

## Checklist

### DIFF
--- a/spec/lucky/text_helpers/svg_inliner_spec.cr
+++ b/spec/lucky/text_helpers/svg_inliner_spec.cr
@@ -1,0 +1,24 @@
+require "./text_helpers_spec"
+
+@[Lucky::SvgInliner::Path("spec/fixtures")]
+module Lucky::SvgInliner
+end
+
+class TextHelperTestPage
+  def with_inlined_svg
+    inline_svg("lucky_logo.svg")
+  end
+end
+
+describe Lucky::SvgInliner do
+  describe ".inline_svg" do
+    it "inlines an svg in a page" do
+      inlined_svg = view.tap(&.with_inlined_svg).render
+      inlined_svg.should start_with %(<svg data-inline-svg="lucky_logo.svg")
+      inlined_svg.should_not contain %(<?xml version="1.0" encoding="UTF-8"?>)
+      inlined_svg.should_not contain %(<!-- lucky logo -->)
+      inlined_svg.should_not contain "\n"
+      inlined_svg.should_not contain %(fill="none" stroke="#2a2a2a" class="logo")
+    end
+  end
+end
--- a/spec/lucky/text_helpers/svg_inliner_spec.cr
+++ b/spec/lucky/text_helpers/svg_inliner_spec.cr
@@ -8,6 +8,10 @@ class TextHelperTestPage
   def with_inlined_svg
     inline_svg("lucky_logo.svg")
   end
+
+  def with_inlined_and_styled_svg
+    inline_svg("lucky_logo.svg", false)
+  end
 end
 
 describe Lucky::SvgInliner do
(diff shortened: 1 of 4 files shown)


wout:
Forgot to mention I'll add the necessary documentation on the website as well when this is approved.

spec/lucky/text_helpers/svg_inliner_spec.cr
```diff
@@ -0,0 +1,37 @@
...
+      inlined_svg.should_not contain %(fill="none" stroke="#2a2a2a" class="logo")
+    end
+
+    it "inlines an svg in a page with its original styling attributes" do
+      inlined_svg = view.tap(&.with_inlined_and_styled_svg).render
+      inlined_svg.should start_with %(<svg data-inline-svg-styled="lucky_logo.svg")
+      inlined_svg.should contain %(fill="none" stroke="#2a2a2a" class="logo")
+      inlined_svg.should_not contain %(<?xml version="1.0" encoding="UTF-8"?>)
+      inlined_svg.should_not contain %(<!-- lucky logo -->)
+      inlined_svg.should_not contain "\n"
```
robacarp, line 34: I like that this strips out un-needed stuff from the XML. Should these `should_not` cases be broken out into an `it "strips out comments"` and `it "strips out the xml header"` set of tests? As it is, if something comes up and someone needs to modify the filter, they don't have a test which directly correlates to that feature.

src/lucky/page_helpers/svg_inliner.cr
```diff
@@ -0,0 +1,27 @@
+annotation Lucky::SvgInliner::Path
+end
+annotation Lucky::SvgInliner::StripRegex
+end
+
+@[Lucky::SvgInliner::Path("src/svgs")]
+@[Lucky::SvgInliner::StripRegex(/(class|fill|stroke|stroke-width|style)="[^"]+" ?/)]
```
robacarp, line 7: Can you help me understand what's going on with these annotations? It seems like a roundabout way to accomplish arguments to the macro.

robacarp (commented):
I love this idea. I use a lot of SVGs and have lamented that I can't easily inline them as part of compilation.

spec/lucky/text_helpers/svg_inliner_spec.cr
wout: Good point. I'll add separate specs for those cases separately.

src/lucky/page_helpers/svg_inliner.cr
wout: Sure. I think having those parameters passed as arguments would be tedious. Let's say the user wants to use a different directory for the SVG files, then it suffices to override the annotation in an initializer:

```crystal
# config/svg_inliner.cr
@[Lucky::SvgInliner::Path("src/icons")]
module Lucky::SvgInliner
end
```

As opposed to (for every single svg):

```crystal
inline_svg("menu-social/twitter.svg", false, "src/icons")
inline_svg("menu-social/discord.svg", false, "src/icons")
```

I prefer to use annotations instead of constants because they're more flexible. You wouldn't be able to override a constant in the same way.

Does that answer your question? Or am I missing the point?

robacarp: Oh interesting, I hadn't considered that. I'm not opposed. Another possibility for this would be to have a Habitat config parameter. @jwoertink @stephendolan thoughts?

wout: Tried that first, but settings from Habitat are not available at compile-time. As far as I know, the only way to allow compile-time configuration is by using annotations of constants.  And using constants is not favourable because it's impossible to override them.

I think these are sensible defaults, so in 99% of the cases, it wouldn't be necessary to change them. The annotations are just an escape hatch.

jwoertink (approved):
I think it looks good :+1:

src/lucky/page_helpers/svg_inliner.cr
robacarp: @wout yeah, you present a use case I can't argue with. Thank you for walking me through it. Let's go!

robacarp (approved):

PR #1762, opened 2022-12-14
Accommodate nil values correctly in JSON param fetching

stephendolan:
## Purpose
Describe the feature or issue and link to the related issue.
If no issue has been opened about this, be sure to open an issue first to discuss the need for this PR.

## Description

## Checklist

### DIFF
--- a/src/lucky/params.cr
+++ b/src/lucky/params.cr
@@ -411,7 +411,11 @@ class Lucky::Params
     nested_key_json = parsed_json[nested_key]? || JSON::Any.new({} of String => JSON::Any)
 
     nested_key_json.as_h.each do |key, value|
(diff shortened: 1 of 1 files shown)


jwoertink:
This is probably fine. I think to really test it, we would have to tell Avram to point to main branch, then write the spec in Avram that tests this whole deal.

jwoertink (approved):

stephendolan:
@jwoertink I actually pulled this functionality down into the helper so there's no inconsistency when this is used by other methods in the same file. It still fixes the API client failure I was seeing and passes our test suite. 🥳

Agreed that we should test this in some way, but I'm not even sure where to start :(

jwoertink:
Good enough for me! Thanks for fixing it :rocket:

PR #1763, opened 2022-12-27
don't capture blocks in the tags so you can still make custom wrapper…

### LINKED ISSUE #1755: Don't capture block in form_for helper
**Describe the bug**

https://github.com/luckyframework/lucky/blob/8d0681a086ff0985e58a319f9ded9002a17eb4e8/src/lucky/tags/form_helpers.cr#L14-L16

By capturing the block, people using that method cannot do a yield. If they use the other overload that does not capture the block, they can.

**To Reproduce**

Create a spec that uses a method like:

```crystal
def form_wrapper(action : Lucky::Action.class)
  form_for action do
    yield
  end
end
```
And it should work as expected, but it fails with an error saying that you can't call `yield` in a captured block.

 - Lucky version (check in shard.lock): 1.0.0-rc1

jwoertink:
## Purpose
Fixes #1755

## Description
Stops capturing blocks when passing down method overloads so you can create your own custom wrapper around some elements.

## Checklist

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -61,7 +61,7 @@ dependencies:
 development_dependencies:
   ameba:
     github: crystal-ameba/ameba
-    version: ~> 1.2.0
+    version: ~> 1.1.0
 
 scripts:
   postinstall: BUILD_WITHOUT_DEVELOPMENT=true script/precompile_tasks
--- a/shard.yml
+++ b/shard.yml
@@ -61,7 +61,7 @@ dependencies:
 development_dependencies:
   ameba:
     github: crystal-ameba/ameba
-    version: ~> 1.1.0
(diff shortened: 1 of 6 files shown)


shard.yml
```diff
@@ -61,7 +61,7 @@ dependencies:
 development_dependencies:
   ameba:
     github: crystal-ameba/ameba
-    version: ~> 1.1.0
+    version: ~> 1.2.0
```
jwoertink, line 64: I had to update ameba to get it to run locally for some reason...

jwoertink: reverted because it tanks on the CI... :thinking: so I either get to run it locally, or on CI, but not both...

PR #1769, opened 2022-12-30
Revamp Accept header handling

### LINKED ISSUE #986: Lucky couldn't figure out what format the client accepts. The client's Accept header: 'image/*;q=0.8'
I have the following action for handing the favicon. (I need to have it as its own action instead of it falling back to asset handing because I have have a catch all route `/:category` elsewhere in the application):

```crystal
class Favicon::Show < BrowserAction
  include Auth::AllowGuests

  before cache_publicly(1.day)

  get "/favicon.ico" do
    file "public/favicon.ico", disposition: "inline"
  end
end
```

When I access a page in Mobile Firefox on iOS and it requests the favicon, I see this error.

```
GET /favicon.ico
 ▸ Handled by Favicon::Show
 ▸ Exception Lucky couldn't figure out what format the client accepts.

    The client's Accept header: 'image/*;q=0.8'

You can teach Lucky how to handle this header:

    # Add this in config/mime_types.cr
    Lucky::MimeType.register "image/*;q=0.8", :custom_format

Or use one of these headers Lucky knows about:

    text/html, application/json, text/json, application/jsonrequest, text/javascript, application/xml, application/rss+xml, application/atom+xml, application/x-yaml, text/yaml, text/csv, text/css, text/calendar, text/plain, multipart/form-data, application/x-www-form-urlencoded

 (Lucky::UnknownAcceptHeaderError)
  from lib/lucky/src/lucky/request_type_helpers.cr:0:7 in 'determine_clients_desired_format'
  from lib/lucky/src/lucky/request_type_helpers.cr:44:41 in 'clients_desired_format'
  from src/actions/favicon/show.cr:6:3 in 'call'
  from lib/lucky/src/lucky/renderable.cr:97:16 in 'perform_action'
  from lib/lucky/src/lucky/route_handler.cr:10:7 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/error_handler.cr:15:5 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/flash_handler.cr:5:5 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/session_handler.cr:5:5 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/log_handler.cr:17:5 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/http_method_override_handler.cr:11:5 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/handler.cr:26:7 in 'call_next'
  from lib/lucky/src/lucky/force_ssl_handler.cr:37:7 in 'call'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/request_processor.cr:48:11 in 'process'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server/request_processor.cr:22:3 in 'process'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server.cr:497:5 in 'handle_client'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/http/server.cr:463:13 in '->'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/fiber.cr:255:3 in 'run'
  from /home/wmoore/.crenv/versions/0.31.1/share/crystal/src/fiber.cr:48:34 in '->'
  from ???
```

I tried adding `Lucky::MimeType.register "image/x-icon", :ico` to the `config/mime_types.cr` as suggested but the error still occurs.

**Lucky:** 0.18.0
**Crystal:** 0.31.0

### LINKED ISSUE #1766: Lucky does not handle Accept header properly
**Describe the bug**

Lucky incorrectly rejects a request as not acceptable, even thought the `Accept` header has a `*/*` entry.

**To Reproduce**

1. Create a lucky app with an API endpoint that inherits from the default `ApiAction`
2. Make a request to the endpoint with an `Accept` header with this value: `*/*; q=0.5, application/xml` 

**Expected behavior**

The request is handled.

**Actual behaviour**

The request is rejected with status 406 Not Acceptable. The error indicates that the client wants XML so this should be added to the accepted formats. However, this is not true the `*/*` entry means the client accepts anything, including the configured format, `:json`.

```
web          |  ▸  Lucky::NotAcceptableError
web          |
web          |      The request wants :xml, but Api::Payments::Webhook does not accept it.
web          |
web          |      Accepted formats: json
web          |
web          |      Try this...
web          |
web          |        ▸ Add :xml to 'accepted_formats' in Api::Payments::Webhook or its parent class.
web          |        ▸ Make your request using one of the accepted formats.
```

 - Lucky version (check in shard.lock): 1.0.0.rc1
 - Crystal version (`crystal --version`): Crystal 1.6.2
 - OS: Linux

**Additional context**

1. The client that is making this request is [Stripe](https://stripe.com/) triggering a webhook.
2. Adding `:xml` to `accepted_formats` allows working around the issue, but should not be necessary
3. This seems related to the other issue I raised a few years ago #986

wezm:
## Purpose

This PR revamps the handling of the `Accept` header to be less adhoc and more in line with [the HTTP specification](https://httpwg.org/specs/rfc9110.html#field.accept).

Fixes #1766
Fixes #986 

## Description

Since I have run into two `Accept` related issues over the years I decided to take a stab at improving the way it's handled. The changes achieve the above purpose by parsing the header according to the grammar in the spec and ordering the media ranges by their weight/qfactor. Lastly it supports wildcards when matching in order to find the client's preferred format.

There are two main concepts introduced:

1. `AcceptList`, which is a list of `MediaRange` sorted by qfactor
2. `MediaRange`, which represents a parsed and valid media range. Example media ranges are `*/*`, `image/*`, `text/html`, `text/plain; charset=UTF-8`, and `text/plain; charset=utf-8; q=0.8`.

With these two classes a matching algorithm is implemented where media ranges that the client accepts are matched against the accepted formats of the action with fallback behaviour that depends on whether the client has a `*/*` wildcard pattern or not.

## Checklist

I've left the commits I made as-is but would be open to squashing them if this was preferred.

### DIFF
--- a/src/lucky/mime_type.cr
+++ b/src/lucky/mime_type.cr
@@ -132,9 +132,12 @@ class Lucky::MimeType
     # Find a matching accepted format by accept list priority
     def find_match(known_formats : Hash(MediaType, Format), accepted_formats : Array(Symbol), default_format : Symbol) : Symbol?
       # If we find a match in the things we accept then pick one of those
-      self.list.each do |media_range|
-        if match = known_formats.find { |media, format| accepted_formats.includes?(format) && media_range.matches?(media) }
-          return match[1]
+      formats_in_common = known_formats.select { |_media, format| accepted_formats.includes?(format) }
+      unless formats_in_common.empty?
+        self.list.each do |media_range|
+          if match = formats_in_common.find { |media, _format| media_range.matches?(media) }
+            return match[1]
+          end
         end
       end
 
@@ -151,7 +154,7 @@ class Lucky::MimeType
         return nil
       end
 
-      # Otherwise the the client accepts anything so use the default format
+      # Finally the client accepts anything so use the default format
       default_format
     end
 
--- a/src/lucky/mime_type.cr
+++ b/src/lucky/mime_type.cr
@@ -39,6 +39,10 @@ class Lucky::MimeType
     accept_header_formats[accept_header_substring] = format
   end
 
+  def self.deregister(accept_header_substring : AcceptHeaderSubstring) : Nil
+    accept_header_formats.delete(accept_header_substring)
+  end
+
   # :nodoc:
   def self.determine_clients_desired_format(request, default_format : Symbol, accepted_formats : Array(Symbol))
     DetermineClientsDesiredFormat.new(request, default_format, accepted_formats).call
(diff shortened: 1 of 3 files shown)


Blacksmoke16:
If you guys are okay with another dependency, could think about using https://athenaframework.org/Negotiation/. Like how it's used in [invidious](https://github.com/iv-org/invidious/pull/2324).

jwoertink:
Wow! Nice work on this @wezm. I'm not familiar with the header spec, but it looks like you've done quite a bit here. I'll need to take some time to understand what all is going on.

Thanks for the suggestion @Blacksmoke16. I'm also not opposed to an extra dependency if it means we get tons of flexibility.

I guess my main question here is, if I only want my action to support HTML, and no other format, if someone tries to pass an Accept header for anything other than HTML, will this still raise an exception? It looks like it does, but my brain is still waking up.

wezm:
> I guess my main question here is, if I only want my action to support HTML, and no other format, if someone tries to pass an Accept header for anything other than HTML, will this still raise an exception? It looks like it does, but my brain is still waking up.

Yes it should. I aimed to preserve all the existing behaviour and all the existing tests continue to pass without edit. In addition the tests I added for the issues I raised also pass now.

src/lucky/mime_type.cr
```diff
@@ -24,7 +35,7 @@ class Lucky::MimeType
   register "application/x-www-form-urlencoded", :url_encoded_form
 
   def self.known_accept_headers : Array(String)
-    accept_header_formats.keys
+    accept_header_formats.keys.map { |media| media.to_s }
```
jwoertink, line 38:
```suggestion
    accept_header_formats.keys.map(&.to_s)
```

jwoertink:
Yeah, looking through this, it feels a lot more flexible than what we have now. I'm down to give it a shot!

robacarp:
This is awesome, the accepts header has needed work for a while and has bitten me several times.

jwoertink (approved):

Issue #1299, opened 2020-11-06
Recursive splat expansion error when mounting Shared::Field


matthewmcgarvey:
**Describe the bug**
When creating a form, we attempted to move a `Shared::Field` that did not pass a block (it relied on the default text_input override) to the first input above the email field. We started getting a compilation error like:
```
mount Shared::FieldErrors, attribute
^----
Error: recursive splat expansion: String, Avram::PermittedAttribute(String | Nil), ...
```

It only happens when the first field is using the override that provides a block.

**To Reproduce**

Here is a branch that has the failing code in a spec. I did try to simplify it, but I'm sure it could be made more straight forward.  https://github.com/matthewmcgarvey/lucky/tree/recursive-splat-expansion

The diff:
https://github.com/luckyframework/lucky/compare/master...matthewmcgarvey:recursive-splat-expansion

**Expected behavior**

Any ordering of mounted fields should work.

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock): v0.24.0
 - Crystal version (`crystal --version`): v0.35.1
 - OS: MacOS

matthewmcgarvey:
So the problem goes away if I move the methods into the component or I inline the `mount` code. There is a bug on crystal for this error message, could I have stumbled onto this problem? https://github.com/crystal-lang/crystal/issues/7164

matthewmcgarvey:
Verified to still be an issue on Crystal v0.36.1

snadon:
I was able to remove my workaround with Lucky 0.29 and Crystal 1.3.2. Not sure when the problem went away though.

matthewmcgarvey:
<img width="634" alt="Screen Shot 2022-01-27 at 8 19 56 PM" src="https://user-images.githubusercontent.com/17329408/151475862-6588ddf4-52a3-4ca5-98d3-47100b109f4f.png">

Verified to still be an issue on Crystal v1.3.0 😢

asterite:
What's the full error with --error-trace?

matthewmcgarvey:
<img width="698" alt="Screen Shot 2022-01-29 at 1 05 52 PM" src="https://user-images.githubusercontent.com/17329408/151674168-4a8ef26a-859a-4e32-becf-a35bf781a647.png">

flipsasser:
I'm still bedeviled by this in 2022. Anyone have an ideas?

jwoertink:
[...]

PR #1771, opened 2023-01-08
Code cleaning

jwoertink:
## Purpose
Mainly just cleaning up some code, and getting the latest ameba setup

## Description
We needed to update to the latest ameba which now takes the stance that `.not_nil!` shouldn't be used. Also after reading the recent articles about incremental compilation, it got me thinking about return types. I started adding a lot more; however, I skipped quite a few ...

There seems to be some cases where you don't put a return type like `initialize`, but it also got me thinking on other types like `to_s`. If the method is technically an override of a higher object / reference, and it doesn't have the return type in the stdlib, should it have it here? Probably not... There's also cases where a method is never intended to use the return value in which case we just say it returns `Nil`. What's the best way to determine which methods these are? 

This can be an ongoing effort... some of these made perfect sense, but there's a few that are "iffy". If any stand out, we can remove it and think about it for later.


jwoertink:
ok... so ameba now seems to require Crystal 1.5 or later (maybe 1.6?) But also, this issue https://github.com/luckyframework/lucky/issues/1299 seems to have been fixed by using Crystal 1.6 or later. Maybe the next release just makes that the minimum with 1.7 releasing today?

robacarp (approved):
easy cleanup, nice

Issue #1827, opened 2023-09-04
Investigating performance hit around building action URLs


jwoertink:
Ref: https://twitter.com/AkitaOnRails/status/1697651647273427429?s=20

TL:DR; It seems that just using `SomeAction.url` causes a pretty significant hit to performance when doing a stress test. 

Here's a snippet of code from a source repo where this was tested.
https://github.com/akitaonrails/rinhabackend-lucky-crystal-api/blob/000a86d0e1d9bdee245c7873f2f0440ed44c7cf8/src/actions/api/pessoas/create.cr#L19-L20

There's a lot of significant moving parts in this area, so it's possible the root cause could be located in some other shard, but we will use this issue to drop any notes or discoveries.

/cc. @akitaonrails

jwoertink:
I noticed the `self.route()` method uses a lot of string concatenation. I tried replacing that with a string builder, and that actually made it slower when benchmarking 10k urls.

Here's a basic action with the `route` method overridden and all the macros stripped out.

```crystal
require "lucky"

Lucky::RouteHelper.configure do |settings|
  settings.base_uri = "http://localhost:3000"
end

class Products::Show < Lucky::Action
  get "/products/:id" do
    plain_text "testing"
  end

  def self.route(id, anchor : String? = nil) : Lucky::RouteHelper
    path = path_from_parts(id)
    query_params = {} of String => String

    unless query_params.empty?
      path += "?#{HTTP::Params.encode(query_params)}"
    end

    anchor.try do |value|
      path += "##{URI.encode_www_form(value)}"
    end

    Lucky::RouteHelper.new :get, path
  end
end

Products::Show.with(1).url
```

matthewmcgarvey:
I think we want to switch from `URI.encode_www_form` to `URI.encode_path` and we can build the path and only call that method once instead of calling it for each path segment (and add it to the base url after). That's my guess as to the slowdown. Of course, if it _is_ the encoding, then we can assume that the path helper will never be as fast as what he did with the raw string, but of course his way doesn't handle invalid characters in the URL.

([doc link for the recommended method](https://crystal-lang.org/api/1.9.2/URI.html#encode_path%28string%3AString%29%3AString-class-method))

matthewmcgarvey:
Link to where I'm talking about: https://github.com/luckyframework/lucky/blob/3f15bc80cc5d26489dce76ffb5d2486720223ac9/src/lucky/routable.cr#L350-L365

[...]

PR #1829, opened 2023-09-06
small performance improvement when generating a route url.

jwoertink:
## Purpose
Ref #1827

## Description
Still investigating what can be done on #1827 for a bit more speed. I don't think we can match the performance of returning a raw string in an action, but little bits like this might help get it closer. This PR utilizes a method overload for `encode_www_form` that takes an IO to reduce the allocations. Here's a benchmark:

```
old 563.31  (  1.78ms) (± 0.65%)  7.17MB/op   1.54× slower
new 866.54  (  1.15ms) (± 0.95%)  3.37MB/op        fastest
```

This is from generating 10k URLs using `Users::Show.with(1).url`

## Checklist

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -338,7 +338,6 @@ module Lucky::Routable
       \{% end %}
     end
 
-
     private def self.path_from_parts(
         {% for param in path_params %}
           {{ param.gsub(/:/, "").id }},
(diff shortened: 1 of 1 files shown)


jkthorne:
one of these days a 33% increase will seem like a giant leap forward

jwoertink:
Looks like this patch makes a pretty big difference on apps under stress load.

PR #1854, opened 2024-03-24
optimize route building when there's query params.

### LINKED ISSUE #1831: Optimize this section of building a path
With https://github.com/crystal-lang/crystal/pull/13798 merged in, Crystal 1.10 will now have the ability to pass an IO in to this method

https://github.com/luckyframework/lucky/blob/0bcc91f729c17c414796929d35e3f99aa8686334/src/lucky/routable.cr#L299

This `Action.route` method will need a small refactor, and using some macro version checks with the `compare_version` macro, but if we make `path_from_parts` take an IO, then we can pass that IO to `HTTP::Params.encode` and `URI.encode_www_form` which will be a small performance boost when generating any routes or paths

jwoertink:
## Purpose
Fixes #1831

## Description
Crystal 1.10 added a method overload for `HTTP::Params.encode` that takes an IO. This lets us use a single IO to build a route when there's query params.

There doesn't seem to really be a difference when there's no query params involved:

```
# SomeAction.with(id: 1).path
old   1.11k (901.42µs) (± 0.13%)  2.91MB/op        fastest
new   1.10k (906.13µs) (± 0.20%)  2.91MB/op   1.01× slower
```

But when you use query params, it helps

```
# SomeAction.with(id: 1, page: 1, per_page: 10).path
old 459.21  (  2.18ms) (± 1.57%)  6.87MB/op   1.30× slower
new 598.36  (  1.67ms) (± 1.12%)  4.27MB/op        fastest
```

Using a Crystal version below 1.10.0 will show a deprecation warning

```
[Deprecated] Please update your Crystal version "1.9.2". Using Lucky with a version below 1.10.0 is deprecated.
```

## Checklist

### DIFF
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -300,6 +300,7 @@ module Lucky::Routable
         unless query_params.empty?
           io << '?'
           {% if compare_versions(Crystal::VERSION, "1.10.0") < 0 %}
+            {% @type.warning("[Deprecated] Please update your Crystal version #{Crystal::VERSION}. Using Lucky with a version below 1.10.0 is deprecated.") %}
             io << HTTP::Params.encode(query_params)
           {% else %}
             HTTP::Params.encode(io, query_params)
--- a/src/lucky/routable.cr
+++ b/src/lucky/routable.cr
@@ -282,29 +282,37 @@ module Lucky::Routable
     {% end %}
     anchor : String? = nil
     ) : Lucky::RouteHelper
-      path = path_from_parts(
-        {% for param in path_params %}
-          {{ param.gsub(/:/, "").id }},
+      path = String.build do |io|
+        path_from_parts(
+          io,
+          {% for param in path_params %}
+            {{ param.gsub(/:/, "").id }},
+          {% end %}
+          {% for param in optional_path_params %}
+            {{ param.gsub(/^\?:/, "").id }},
+          {% end %}
+        )
+        query_params = {} of String => String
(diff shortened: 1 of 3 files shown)


robacarp (approved):

src/lucky/routable.cr
```diff
@@ -282,29 +282,37 @@ module Lucky::Routable
...
-      {% for param in PARAM_DECLARATIONS %}
-        # add query param if given and not nil
-        query_params["{{ param.var }}"] = {{ param.var }}.to_s unless {{ param.var }}.nil?
-      {% end %}
-      unless query_params.empty?
-        path += "?#{HTTP::Params.encode(query_params)}"
-      end
+        unless query_params.empty?
+          io << '?'
+          {% if compare_versions(Crystal::VERSION, "1.10.0") < 0 %}
```
robacarp, line 302: Should this emit a deprecated message somehow? Code like this probably shouldn't live forever.

jwoertink: True. The deprecation is an annotation though.. Oh, I could just move the call in to a separate method with the deprecation. Ok, I'll look at doing that.

Blacksmoke16: As of `1.9.0` there is `ASTNode#warning` you can use to handle it manually. `@[Deprecated]` basically does the same thing.

jwoertink: oh! I didn't know. That's great. Thanks for the tip @Blacksmoke16 :raised_hands:

PR #1855, opened 2024-03-24
Adding the compile time to the watcher for development

### LINKED ISSUE #1851: Add the compile time to the watcher in development
I like seeing how long my assets take to build. Would be neat to see how long the compilation takes as well.

Maybe our "Done" message could just mimic the one in green here?

image

jwoertink:
## Purpose
Fixes #1851

## Description

This adds in the time span of compiling your app when it boots. 

Here's what it looks like
image

## Checklist

### DIFF
--- a/tasks/watch.cr
+++ b/tasks/watch.cr
@@ -152,8 +152,10 @@ module LuckySentry
 
     @app_built : Bool = false
     @successful_compilations : Int32 = 0
+    @build_started : Time::Span
 
     def initialize(@build_commands : Array(String), @run_commands : Array(String), @files : Array(String), @reload_browser : Bool, @watcher : Watcher?)
+      @build_started = Time.monotonic
     end
 
     private def build_app_processes_and_start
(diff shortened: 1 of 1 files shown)


tasks/watch.cr
```diff
@@ -262,7 +264,14 @@ module LuckySentry
       if build_success
         self.app_built = true
         create_app_processes
-        puts "#{" Done ".colorize.on_cyan.black} compiling"
+
+        elapsed_time = Time.monotonic - @build_started
+        message = String.build do |io|
+          io << " DONE ".colorize.on_cyan.black
+          io << " Compiled successfully in #{elapsed_time}".colorize.cyan
```
robacarp, line 271: What about calling distance_of_time_in_words or something on this so it's a little easier on the eyes?

robacarp (approved):

tasks/watch.cr
jwoertink: oh, that's a good idea. I was thinking I didn't want to write out a whole deal, but I forgot I could just use that.

jwoertink: Alright, made these changes and updated the screenshot. It looks a lot better.

robacarp: excellent!

PR #1866, opened 2024-04-16
Adds Date response header.

### LINKED ISSUE #1863: Add Date header to responses
I saw this issue on Kemal https://github.com/kemalcr/kemal/issues/668

and it looks like the spec says almost all responses should contain a Date header

```
Origin servers MUST include a Date header field in all responses,
   except in these cases:

      1. If the response status code is 100 (Continue) or 101 (Switching
         Protocols), the response MAY include a Date header field, at
         the server's option.

      2. If the response status code conveys a server error, e.g. 500
         (Internal Server Error) or 503 (Service Unavailable), and it is
         inconvenient or impossible to generate a valid Date.

      3. If the server does not have a clock that can provide a
         reasonable approximation of the current time, its responses
         MUST NOT include a Date header field. In this case, the rules
         in [section 14.18.1](https://datatracker.ietf.org/doc/html/rfc2616#section-14.18.1) MUST be followed.
```

We don't currently set that, and it should be easy to add in, so might as well!

jwoertink:
## Purpose
Fixes #1863

## Description
Apparently all servers should respond with a Date header. Kemal just added it in, so why not? :smile: 

## Checklist

### DIFF
--- a/src/lucky/text_response.cr
+++ b/src/lucky/text_response.cr
@@ -28,6 +28,7 @@ class Lucky::TextResponse < Lucky::Response
     end
     context.response.content_type = content_type
     context.response.status_code = status
+    context.response.headers.add "Date", HTTP.format_time(Time.utc)
     gzip if should_gzip?
     context.response.print(body) if should_print?
   rescue e : IO::Error


robacarp (approved):
🌴

Issue #1868, opened 2024-04-19
Improve the error output when newbie user forget settings LUCKY_ENV to production.


zw963:
**Describe the bug**

When user build a static binary, then copy file into server for deploy, if user  no correct setting `LUCKY_ENV`  to `production`, will get error message like following:

Unhandled exception: Expected config file for the watcher at ./config/watch.yml (Exception)

Which raised from line 20 of config/server.cr. (please check following screenshot)

![image](https://github.com/luckyframework/lucky/assets/549126/39d34bb1-8eca-4242-a534-d805472dccaf)

**Expected behavior**
I consider the better way is, to told user, if forget settings the correct `LUCKY_ENV`  to production?

From the perspective of a new user (it's been a long time since I used Lucky), this is much more useful.

**Screenshots/code**

```
root@test1:~/college# bin/college
Unhandled exception: Expected config file for the watcher at ./config/watch.yml (Exception)
  from college/lib/lucky/src/lucky/server_settings.cr:43:7 in 'yaml_settings_file'
  from college/lib/lucky/src/lucky/server_settings.cr:33:33 in 'settings'
  from college/lib/lucky/src/lucky/server_settings.cr:11:25 in 'host'
  from college/config/server.cr:20:21 in '__crystal_main'
  from /home/zw963/Crystal/share/crystal/src/crystal/main.cr:129:5 in 'main_user_code'
  from /home/zw963/Crystal/share/crystal/src/crystal/main.cr:115:7 in 'main'
  from /home/zw963/Crystal/share/crystal/src/crystal/main.cr:141:3 in 'main'
```

**Versions (please complete the following information):**
 - Lucky version (check in shard.lock): 1.1.0 
 - Crystal version (`crystal --version`): 1.12.1
 - OS: Arch

Thank you.

PR #1896, opened 2024-06-03
add tip to set LUCKY_ENV to production

snadon:
## Purpose
Fixes https://github.com/luckyframework/lucky/issues/1868

## Checklist

### DIFF
--- a/src/lucky/server_settings.cr
+++ b/src/lucky/server_settings.cr
@@ -40,8 +40,15 @@ module Lucky::ServerSettings
     if File.exists?(YAML_SETTINGS_PATH)
       File.read YAML_SETTINGS_PATH
     else
-      raise "Expected config file for the watcher at #{YAML_SETTINGS_PATH}. \
-      Did you forget to set LUCKY_ENV to production?"
+      <<-ERROR
+      Expected config file for the watcher at #{YAML_SETTINGS_PATH}.
+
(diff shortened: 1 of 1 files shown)


src/lucky/server_settings.cr
```diff
@@ -40,7 +40,8 @@ module Lucky::ServerSettings
     if File.exists?(YAML_SETTINGS_PATH)
       File.read YAML_SETTINGS_PATH
     else
-      raise "Expected config file for the watcher at #{YAML_SETTINGS_PATH}"
+      raise "Expected config file for the watcher at #{YAML_SETTINGS_PATH}. \
+      Did you forget to set LUCKY_ENV to production?"
```
jwoertink, line 44: I like this. I wonder if maybe we can make it look like the other exceptions so it stands out a bit?

```
raise <<-ERROR
Expected config file for the watcher at #{YAML_SETTINGS_PATH}.

Try this...

  ▸ If this is Production, be sure to set LUCKY_ENV=production
  ▸ If this is Development, ensure the #{YAML_SETTINGS_PATH} file exists

ERROR
```

I'm open to wording, but just tossing out the idea.

jwoertink (approved):
:+1: Thanks!

jwoertink (approved):

PR #1900, opened 2024-06-15
Fixing Crystal 1.13 regression issues

### LINKED ISSUE #1872: Compile error using Crystal nightly
This just started in the past few days

image

jwoertink:
## Purpose
Fixes #1872 

## Description
Crystal 1.13 changes how `Nil` is written on the macro side. This broke some of the Lucky code because now `::Nil.id` is not the same as `Nil.id`.

## Checklist

### DIFF
--- a/src/lucky/assignable.cr
+++ b/src/lucky/assignable.cr
@@ -89,7 +89,9 @@ module Lucky::Assignable
       {% sorted_assigns = ASSIGNS.sort_by { |dec|
            has_explicit_value =
              dec.type.is_a?(Metaclass) ||
-               dec.type.types.map(&.id).includes?(Nil.id) ||
+               dec.type.types.any? { |t|
+                 (t.is_a?(Metaclass) || t.is_a?(ProcNotation) || t.is_a?(Generic)) ? false : t.names.includes?(Nil.id)
+               } ||
                !dec.value.is_a?(Nop)
            has_explicit_value ? 1 : 0
          } %}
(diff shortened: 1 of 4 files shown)


src/lucky/assignable.cr
```diff
@@ -89,7 +89,9 @@ module Lucky::Assignable
       {% sorted_assigns = ASSIGNS.sort_by { |dec|
            has_explicit_value =
              dec.type.is_a?(Metaclass) ||
-               dec.type.types.map(&.id).includes?(Nil.id) ||
+               dec.type.types.any? { |t|
+                 (t.is_a?(Metaclass) || t.is_a?(ProcNotation) || t.is_a?(Generic)) ? false : t.names.includes?(Nil.id)
```
jwoertink, line 93: Not a fan of doing this, but if we use `type.resolve` here we will have to force all Lucky apps to start adding a bunch of `require` statements at the top of all the files because of examples like..

```crystal
class PostPage < MainLayout
   # in macro land, Crystal doesn't know what "User" is yet. If we resolve this type at compile time, we will have to require it at the top of this file.
  needs user : User

  # ...
end
```

This starts to compound when you're talking about hundreds of actions, models, queries, operations, pages, etc...

src/lucky/assignable.cr
```diff
@@ -98,7 +100,7 @@ module Lucky::Assignable
           {% var = declaration.var %}
           {% type = declaration.type %}
           {% value = declaration.value %}
-          {% value = nil if type.stringify.ends_with?("Nil") && !value %}
+          {% value = nil if type.stringify.ends_with?("Nil") && value.nil? %}
```
jwoertink, line 103: I changed this because if for some reason you did `param thingable : Bool? = false` it would default the value to `nil`. Probably a non-issue, but why not? :man_shrugging:

src/lucky/routable.cr
```diff
@@ -260,7 +260,7 @@ module Lucky::Routable
     end
 
     {% params_with_defaults = PARAM_DECLARATIONS.select do |decl|
-         !decl.value.is_a?(Nop) || decl.type.is_a?(Union) && decl.type.types.last.id == Nil.id
+         !decl.value.is_a?(Nop) || decl.type.is_a?(Union) && decl.type.resolve.nilable?
```
jwoertink, line 263: Here `resolve.nilable?` *should* be ok :crossed_fingers:  .... but if it starts blowing up apps, then we know where to come back to..

PR #1908, opened 2024-08-11
Fix invalid query string for array params

### LINKED ISSUE #1880: Building a route with array params doesn't work
If you have an action that specifies an array param, you're not able to build a route from that action.

```crystal
class Reports::Query < ApiAction
  param codes : Array(String)

  get "/reports/query" do
    results = ReportQuery.new.codes(codes)

    json(ReportSerializer.for_collection(results))
  end
end

# then in your spec for this

it "doesn't work" do
  client = ApiClient.new
  response = client.exec(Reports::Query.with(codes: ["a", "b", "c"]))
  response.status_code.should eq(200) # but it's actually 400 here
end
```

The route that's built looks more like

```
/reports/query?codes=%5B%22a%22%2C+%22b%22%2C+%22c%22%5D
```

Which basically looks like

```
@query_params=URI::Params{"codes" => ["[\"a\", \"b\", \"c\"]"]},
```

akadusei:
## Purpose
Fixes #1880
Duplicate of #1907

## Description
In #1907, I assumed the square brackets (`[]`) in array query params should not be encoded. I believe it should, and this PR addresses that. If accepted, #1907 should be rejected.

## Checklist

### DIFF
--- a/spec/lucky/base_http_client_spec.cr
+++ b/spec/lucky/base_http_client_spec.cr
@@ -10,6 +10,16 @@ class HelloWorldAction < TestAction
   end
 end
 
+class ArrayParamAction < TestAction
+  accepted_formats [:plain_text]
+
+  param codes : Array(String)
+
+  post "/array_param" do
+    plain_text codes.join("--")
+  end
+end
+
 class MyClient < Lucky::BaseHTTPClient
   app TestServer.new
 end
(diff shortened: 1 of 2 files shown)


jwoertink:
Yeah, this makes sense for them to be encoded. The answer on [this stackoverflow](https://stackoverflow.com/questions/11490326/is-array-syntax-using-square-brackets-in-url-query-strings-valid) says that square brackets can appear in the host, but should be percent encoded in the query string. 

Looks like there's some weird failures going on. I don't *think* they're related to this PR. At least for sure not the exception page, but can you look at the cookie stuff? If it's unrelated, then we can fix that separately. 

Thanks for taking this on!

PR #1916, opened 2024-10-06
Adding a MaximumRequestSizeHandler

### LINKED ISSUE #1143: Have maximum request size by default
It would be great to check incoming requests to make sure they are below a certain size. Otherwise, a malicious client could hit the server with a huge request and cause a denial-of-service attack. Perhaps this could be an adjustable value in `config/server.cr`.

russ:
* Configurable to be on/off. Off by default.
* Configurable to set the maximum request size. Default is 1MB.

## Purpose
A configurable handler for setting a maximum size on requests.

## Description
Fixes #1143

## Checklist

### DIFF
--- a/spec/lucky/maximum_request_size_handler_spec.cr
+++ b/spec/lucky/maximum_request_size_handler_spec.cr
@@ -1,0 +1,51 @@
+require "../spec_helper"
+require "http/server"
+
+include ContextHelper
+
+describe Lucky::MaximumRequestSizeHandler do
+  context "when the handler is disabled" do
+    it "simply serves the request" do
+      context = build_small_request_context("/path")
+      Lucky::MaximumRequestSizeHandler.temp_config(enabled: false) do
+        run_request_size_handler(context)
+      end
+      context.response.status.should eq(HTTP::Status::OK)
+    end
+  end
+
+  context "when the handler is enabled" do
+    it "with a small request, serve the request" do
+      context = build_small_request_context("/path")
+      Lucky::MaximumRequestSizeHandler.temp_config(enabled: true) do
+        run_request_size_handler(context)
+      end
+      context.response.status.should eq(HTTP::Status::OK)
+    end
+
+    it "with a large request, deny the request" do
+      context = build_large_request_context("/path")
+      Lucky::MaximumRequestSizeHandler.temp_config(enabled: true) do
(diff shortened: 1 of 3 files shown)


russ:
Where should docs for this go?

spec/lucky/maximum_request_size_handler_spec.cr
```diff
@@ -0,0 +1,51 @@
...
+      end
+      context.response.status.should eq(HTTP::Status::OK)
+    end
+
+    it "with a large request, deny the request" do
+      context = build_large_request_context("/path")
+      Lucky::MaximumRequestSizeHandler.temp_config(enabled: true) do
+        run_request_size_handler(context)
+      end
+      context.response.status.should eq(HTTP::Status::PAYLOAD_TOO_LARGE)
```
jwoertink, line 31: seems this failed?

spec/spec_helper.cr
```diff
@@ -48,4 +48,8 @@ Lucky::ForceSSLHandler.configure do |settings|
   settings.enabled = true
 end
 
+Lucky::MaximumRequestSizeHandler.configure do |settings|
+  settings.enabled = false
+end
```
jwoertink, line 53: this is probably not needed since the default is `false`, right?

src/lucky/maximum_request_size_handler.cr
```diff
@@ -0,0 +1,40 @@
+class Lucky::MaximumRequestSizeHandler
+  include HTTP::Handler
+
+  Habitat.create do
+    setting enabled : Bool = false
+    setting max_size : Int32 = 1_048_576 # 1MB
```
jwoertink, line 6: :+1:

jwoertink (commented):
From what I can tell, this seems like it's fine. I'm not sure why the specs are failing on it though. As for docs, either at the top of the class file just above the class definition, or we just write them up on the website.

src/lucky/maximum_request_size_handler.cr
jwoertink: actually... if you wanted this to be say 10GB (video uploads), would this need to be Int64?

jwoertink (approved):
I think the failure is related to changes in ExceptionPage and probably unrelated. The rest looks good! Thanks :raised_hands:

PR #1917, opened 2024-10-06
Action based rate limiting.

### LINKED ISSUE #1865: Action based rate limiting
To protect a web app against abusive connections / requests (sub DDOS scale) several layers can help (nginx, [middleware](https://github.com/defense-cr/defense)), but stumbling on [Rails future rate limiting ](https://github.com/rails/rails/pull/50490), I'm quite jealous of the elegant DX of this solution.

I could clearly see this working in Lucky :

```crystal
class SignIns::New < BrowserAction
  include Auth::RedirectSignedInUsers
  rate_limit to: 50, within: 10.seconds

  get "/sign_in" do
    html NewPage, operation: SignInUser.new
  end
end
``` 

I can't implement this right now, I'm just putting it out there as a starting point for discussion / inspiration.

russ:
## Purpose
Adds a way to set per action rate limits.
Fixes #1865

## Description
Include the module and write a rate_limit method to define the limit.
The rate_limit_key method can be overridden if you want differeny key logic.

## Checklist

### DIFF
--- a/spec/lucky/rate_limit_spec.cr
+++ b/spec/lucky/rate_limit_spec.cr
@@ -1,0 +1,43 @@
+require "../spec_helper"
+
+include ContextHelper
+
+class RateLimitRoutes::Index < TestAction
+  include Lucky::RateLimit
+
+  get "/rate_limit" do
+    plain_text "hello"
+  end
+
+  private def rate_limit : NamedTuple(to: Int32, within: Time::Span)
+    {to: 1, within: 1.minute}
+  end
+end
+
+describe Lucky::RateLimit do
+  describe "RateLimit" do
+    it "when request count is less than the rate limit" do
+      headers = HTTP::Headers.new
+      headers["X_FORWARDED_FOR"] = "127.0.0.1"
+      request = HTTP::Request.new("GET", "/rate_limit", body: "", headers: headers)
+      context = build_context(request)
+
+      route = RateLimitRoutes::Index.new(context, params).call
+      route.context.response.status.should eq(HTTP::Status::OK)
+    end
+
+    it "when request count is over the rate limit" do
+      headers = HTTP::Headers.new
+      headers["X_FORWARDED_FOR"] = "127.0.0.1"
+      request = HTTP::Request.new("GET", "/rate_limit", body: "", headers: headers)
+      context = build_context(request)
+
+      10.times do
+        RateLimitRoutes::Index.new(context, params).call
+      end
+
+      route = RateLimitRoutes::Index.new(context, params).call
+      route.context.response.status.should eq(HTTP::Status::TOO_MANY_REQUESTS)
+    end
+  end
+end
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -48,4 +48,8 @@ Lucky::ForceSSLHandler.configure do |settings|
   settings.enabled = true
 end
 
+LuckyCache.configure do |settings|
+  settings.storage = LuckyCache::MemoryStore.new
+end
+
 Habitat.raise_if_missing_settings!
--- a/src/lucky/rate_limit.cr
+++ b/src/lucky/rate_limit.cr
@@ -1,0 +1,41 @@
+module Lucky::RateLimit
+  macro included
+    before enforce_rate_limit
+  end
+
+  abstract def rate_limit : NamedTuple(to: Int32, within: Time::Span)
+
+  private def enforce_rate_limit
+    cache = LuckyCache.settings.storage
+    count = cache.fetch(rate_limit_key, as: Int32, expires_in: rate_limit["within"]) { 0 }
+    cache.write(rate_limit_key, expires_in: rate_limit["within"]) { count + 1 }
+
+    if count > rate_limit["to"]
+      context.response.status = HTTP::Status::TOO_MANY_REQUESTS
+      context.response.headers["Retry-After"] = rate_limit["within"].to_s
+      plain_text("Rate limit exceeded")
+    else
+      continue
+    end
+  end
+
+  private def rate_limit_key : String
+    klass = self.class.to_s.downcase.gsub("::", ":")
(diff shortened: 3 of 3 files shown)


spec/spec_helper.cr
```diff
@@ -48,4 +48,8 @@ Lucky::ForceSSLHandler.configure do |settings|
   settings.enabled = true
 end
 
+LuckyCache.configure do |settings|
```
russ, line 51: I'm not a huge fan of this but I'm not sure how best to configure the storage at the individual spec level.

src/lucky/rate_limit.cr
```diff
@@ -0,0 +1,41 @@
...
+      continue
+    end
+  end
+
+  private def rate_limit_key : String
+    klass = self.class.to_s.downcase.gsub("::", ":")
+    "ratelimit:#{klass}:#{rate_limit_identifier}"
+  end
+
+  private def rate_limit_identifier : Socket::Address | Nil
```
russ, line 27: This code is a copy of the RemoteIP code. You can override the method in your action, but the default needs something and I think the IP makes sense. The specs pass in a HTTP::Request so I don't get `context.remote_ip`.

jwoertink: I think you should have access to `request.remote_ip`  https://github.com/luckyframework/lucky/blob/bddb10b79c35699bd51bba91897f52d0114fab3b/src/charms/request_extensions.cr#L1-L8

This is actually patched in Lucky. `context` doesn't have a `remote_ip` method, but you may be thinking of `context.request_id` which is something different. 

Now, with that said, it brings up the point that if you don't have the `RemoteIpHandler` in your middleware stack, that could cause issues... I wonder if there's a way that we could say you must have that handler in your stack in order to use this module? :thinking:

src/lucky/rate_limit.cr
```diff
@@ -0,0 +1,41 @@
...
+      context.response.status = HTTP::Status::TOO_MANY_REQUESTS
+      context.response.headers["Retry-After"] = rate_limit["within"].to_s
+      plain_text("Rate limit exceeded")
+    else
+      continue
+    end
+  end
+
+  private def rate_limit_key : String
+    klass = self.class.to_s.downcase.gsub("::", ":")
```
jwoertink, line 23:
```suggestion
    klass = {{ @type.stringify.downcase.gsub(/::/, ":") }}
```

I'm not sure if this matters here, but if we build it at compile time, then it won't need to compute on each request.

spec/lucky/rate_limit_spec.cr
```diff
@@ -0,0 +1,43 @@
...
+class RateLimitRoutes::Index < TestAction
+  include Lucky::RateLimit
+
+  get "/rate_limit" do
+    plain_text "hello"
+  end
+
+  private def rate_limit : NamedTuple(to: Int32, within: Time::Span)
+    {to: 1, within: 1.minute}
+  end
```
jwoertink, line 14: This is actually a pretty slick interface. It feels consistent with some other Lucky modules.

jwoertink (commented):
This is amazing :rocket:  Thanks for tackling this. I think we should definitely add LuckyCache as a dependency here to solve the spec failure. There's stuff in LuckyCache for doing view caching anyway, so it makes sense. I left a few other comments too.

src/lucky/rate_limit.cr
rmarronnier, line 1: Thanks so much for this !

I'd love to get a macro one liner with keeping the possibility to have a method coputing values at runtime like you proposed. 

Example : 
```crystal

class RateLimitRoutes::Index < TestAction
  include Lucky::RateLimit
  rate_limit to: 1, within: 1.minute

  get "/rate_limit" do
    plain_text "hello"
  end
end
```
The dynamic setup still works as before :
```crystal
class ComputedRateLimitRoutes::Index < TestAction
  include Lucky::RateLimit

  get "/computed_rate_limit" do
    plain_text "hello"
  end

  private def computed_rate_limit : NamedTuple(to: Int32, within: Time::Span)
    {to: 1, within: 1.minute}
  end
end

```

Here's what I got working locally :


```crystal

module Lucky::RateLimit
  macro included
    before enforce_rate_limit
  end

  macro rate_limit(**tuple)
    private def computed_rate_limit : NamedTuple(to: Int32, within: Time::Span)
      {{tuple}}
    end
  end

  abstract def computed_rate_limit : NamedTuple(to: Int32, within: Time::Span)

  private def enforce_rate_limit
    cache = LuckyCache.settings.storage
    count = cache.fetch(rate_limit_key, as: Int32, expires_in: computed_rate_limit["within"]) { 0 }
    cache.write(rate_limit_key, expires_in: computed_rate_limit["within"]) { count + 1 }

    if count > computed_rate_limit["to"]
      context.response.status = HTTP::Status::TOO_MANY_REQUESTS
      context.response.headers["Retry-After"] = computed_rate_limit["within"].to_s
      plain_text("Rate limit exceeded")
    else
      continue
    end
  end

  private def rate_limit_key : String
    klass = self.class.to_s.downcase.gsub("::", ":")
    "ratelimit:#{klass}:#{rate_limit_identifier}"
  end

  private def rate_limit_identifier : Socket::Address | Nil
    request = context.request

    if x_forwarded = request.headers["X_FORWARDED_FOR"]?.try(&.split(',').first?).presence
      begin
        Socket::IPAddress.new(x_forwarded, 0)
      rescue Socket::Error
        # if the x_forwarded is not a valid ip address we fallback to request.remote_address
        request.remote_address
      end
    else
      request.remote_address
    end
  end
end
```

I'm sure there are still improvements macro wise but that's the idea.

rmarronnier:
@russ : Do you plan on continuing you work on this ? If not, how would you feel if I pick it up ?

russ:
> @russ : Do you plan on continuing you work on this ? If not, how would you feel if I pick it up ?

@rmarronnier  I keep meaning to come back to this. Please, feel free to pick it up if you have the time. 😄

PR #1927, opened 2024-10-31
Add abstract definition for debug_message

garymardell:
## Purpose

To make it clearer that `debug_message` must be implemented on `Response`.

## Description

I am using lucky to build a microservice that accepts protobuf definitions over http so I created my own `ProtobufResponse` class to stream the protobuf response straight into the response io. While implementing it I discovered that you also have to implement `debug_message` as well as it is required by renderable:

https://github.com/luckyframework/lucky/blob/2c1433d20c3210109585f048af64d5a4d34cf3b3/src/lucky/renderable.cr#L194-L196

## Checklist

### DIFF
--- a/src/lucky/data_response.cr
+++ b/src/lucky/data_response.cr
@@ -31,7 +31,8 @@
 class Lucky::DataResponse < Lucky::Response
   DEFAULT_STATUS = 200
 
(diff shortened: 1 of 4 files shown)


jwoertink (approved):
Looks good to me! Thanks :+1:

PR #1942, opened 2025-03-31
Refactor `Lucky::UploadedFile#blank?` to check the actual file size

### LINKED ISSUE #1147: Suggestion: Change UploadedFile name to use HTTP::FormData::Part filename
I have an app where I upload an image and save it locally. I was saving the file using the Lucky::UploadedFile name attribute in my action and noticed the saved file was using the name of the form field and losing the extension name as well. I noticed tempfile was another attribute and attempted to use tempfile.name but got the same result as the temp file gets created using part.name as well.

I looked in the source code and figured out I could get the filename through UploadedFile.metadata.filename. Which is fine since now I know how to get the filename. However, I feel others could also run into this problem as well thinking UploadedFile.name will yield a filename.

So changing [this line](https://github.com/luckyframework/lucky/blob/5720e099143b224f98b903602b1cb7ef221bcf2b/src/lucky/uploaded_file.cr#L13) to use part.filename instead of part.name.

### LINKED ISSUE #1938: File upload discarded if filename not set
Avram discards file uploads in an operation if `filename` is not set in the `Content-Disposition` header:

```crystal
file_headers = HTTP::Headers{
  "Content-Disposition" => %q(form-data; name="file")
}

file_io = IO::Memory.new("Hello")
part = HTTP::FormData::Part.new(file_headers, file_io)
file = Lucky::UploadedFile.new(part)

UploadFile.create(
  # ...
  file: file
) do # ...
```

Inspecting `file.value` returns `nil` in the operation. This is resolved by adding a filename to the header:

```crystal
file_headers = HTTP::Headers{
  "Content-Disposition" => %q(form-data; name="file"; filename="hello.txt")
}
```

This is because `Avram::Attribute#value` returns `nil` if value is blank: https://github.com/luckyframework/avram/blob/55556c9baabbfc5fb4d748062f1c066cb261c71a/src/avram/attribute.cr#L68

...and `Lucky::UploadedFile#blank?` is `true` only if `filename` is set:

https://github.com/luckyframework/lucky/blob/5607fe69ce631569bed96376cc5f36654f95fcce/src/lucky/uploaded_file.cr#L44-L46

My concern is that `filename` is not a required parameter in the `Content-Disposition` header for multipart forms ([RFC 2388](https://datatracker.ietf.org/doc/html/rfc2388#section-4.4)), so the absence of filename does not mean the absence of a file. Therefore, using that to determine that a file is blank is probably problematic.

Ideally, there should be a way to check if the file IO itself is blank/empty. After a *tete-a-tete* with @jwoertink on discord, we arrived at the following:

- An uploaded file should always be considered as not blank, with or without a filename.
- Where there is no filename, Lucky should generate a random filename for it.
- Use `filename` (instead of `name`) for the tempfile that saves the uploaded file locally. See #1147 

Leaving this here to hear any thoughts.

CC @wout

akadusei:
## Purpose

Fixes #1938 
Closes #1147

## Description

(See linked issues)

## Checklist

### DIFF
--- a/src/lucky/uploaded_file.cr
+++ b/src/lucky/uploaded_file.cr
@@ -30,14 +30,13 @@ class Lucky::UploadedFile
     @tempfile.path
   end
 
-  # Tests if the file name is blank, which typically means no file was selected
-  # at the time the form was submitted.
+  # Tests if the file size is zero.
   #
   # ```
   # uploaded_file_object.blank? # => Bool
   # ```
   def blank? : Bool
-    filename.blank?
+    tempfile.size.zero?
   end
 
   @[Deprecated("`metadata` deprecated. Each method on metadata is accessible directly on Lucky::UploadedFile")]
--- a/src/lucky/uploaded_file.cr
+++ b/src/lucky/uploaded_file.cr
@@ -1,6 +1,8 @@
 # This class represents an uploaded file from a form
 class Lucky::UploadedFile
   private getter part : HTTP::FormData::Part
+
+  getter filename : String
   getter tempfile : File
 
   delegate name, creation_time, modification_time, read_time, size, to: @part
@@ -11,7 +13,9 @@ class Lucky::UploadedFile
   # provided HTTP::FormData::Part and the **tempfile** will
   # be assigned the body of the HTTP::FormData::Part
   def initialize(@part : HTTP::FormData::Part)
-    @tempfile = File.tempfile(@part.name)
+    @filename = @part.filename.presence || Random.new.hex(12)
+    @tempfile = File.tempfile(filename)
+
     File.open(@tempfile.path, "w") do |tempfile|
       IO.copy(@part.body, tempfile)
     end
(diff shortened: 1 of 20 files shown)


src/lucky/uploaded_file.cr
```diff
@@ -26,23 +30,13 @@ class Lucky::UploadedFile
...
-  # Tests if the file name is blank, which typically means no file was selected
-  # at the time the form was submitted.
+  # Tests if the file size is zero.
   #
   # ```
   # uploaded_file_object.blank? # => Bool
   # ```
   def blank? : Bool
-    filename.blank?
+    tempfile.size.zero?
```
jwoertink, line 39: Yeah, I think this makes more sense :+1:

jwoertink (approved):
I submitted a ticket on that CI failure https://github.com/crystal-lang/crystal/issues/15617 but the rest looks good :+1:

PR #1950, opened 2025-04-28
Fix missing nested param error when nested key is set but empty

### LINKED ISSUE #1949: `Lucky::MissingNestedParamError` when nested params are present but empty
I have an operation where all attributes are optional, so when the user does not fill in anything, the API client sends `{ category: {} }` in its request body. This raises `Lucky::MissingNestedParamError`, even though the param was set (only it's empty).

The `Lucky::Params#nested` checks if the resulting hash is empty (rather than if the nested key itself was set), and raises the error:

https://github.com/luckyframework/lucky/blob/e161fc52f88a2047e6aa699089bf1466426c7514/src/lucky/params.cr#L236-L243

The same issue occurs if you have an array attribute in an operation. Even if the nested param is not empty, you would still have to fill in the array attribute to avoid the error:

```crystal
    response = client.exec(Features::Create, feature: {
      choices: Array(String).new, # Without this, you get "Missing param key: 'feature' (Lucky::MissingNestedParamError)..."
      maximum: 99,
      minimum: 3,
      name: "Number of Kitchens",
      type: :integer
    })
```

I assume the same holds true for other `#nested...` methods.

In Avram, the `Avram::Attribute#extract` methods call `params.nested()` before getting the individual attribute from the resulting hash, and that's where the error is triggered: https://github.com/luckyframework/avram/blob/c1827c63b41782a19c34307bde8434ccd7623db1/src/avram/attribute.cr#L117

I'm wondering why we call `#nested` instead of `#nested?`. My expectation was that the operation proceeds and let the validations do their work, even if the nested key was not set at all.

@jwoertink suggested via discord "What if we just added a property to params that allows you to ignore empty or whatever?"

```crystal
def to_unsafe # or any name
  @allow_unsafe = true
  self
end

if !params.allow_unsafe? && nested?(key).keys.empty?
 raise
end
```

```crystal
SaveUser.create(params.to_unsafe) do |_,_|
end
```

Any more suggestions or ideas?

akadusei:
## Purpose
Fixes #1949

## Description
The previous implementations was correct and expected for nested arrays and files, so this PR changes only `#nested?` and `#nested`. The rest is just a refactor of the other `#nested_...` methods, to provide a consistent API.

## Checklist

### DIFF
--- a/spec/lucky/params_spec.cr
+++ b/spec/lucky/params_spec.cr
@@ -443,6 +443,14 @@ describe Lucky::Params do
       end
     end
 
+    it "does not raise if nested key is set but empty" do
+      request = build_request body: {user: NamedTuple.new}.to_json,
+        content_type: "application/json"
+
+      params = Lucky::Params.new(request)
+      params.nested(:user).should eq(Hash(String, String).new)
+    end
+
     it "returns empty hash if nested_params are missing" do
       request = build_request body: "",
         content_type: "application/x-www-form-urlencoded"
--- a/src/lucky/params.cr
+++ b/src/lucky/params.cr
@@ -264,24 +264,11 @@ class Lucky::Params
   # params.nested_array("missing") # Missing parameter: missing
   # ```
   def nested_arrays(nested_key : String | Symbol) : Hash(String, Array(String))
-    nested_params = nested_arrays?(nested_key)
-    if nested_params.keys.empty?
-      raise Lucky::MissingNestedParamError.new nested_key
-    else
-      nested_params
-    end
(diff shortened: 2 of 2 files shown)


jwoertink (commented):
Looks like a nice little refactor. Not sure what's with the failing spec, but if it's related to Crystal 1.10 specifically, we can look at bumping the min Crystal version.

akadusei:
I have no idea. Maybe try rerunning that failing job again? 

The `Dockerfile` and the `.crystal-version` file still reads "1.6.2", though.

jwoertink:
Ah, good call. It must have been some race condition. I'm gonna go through and update the min Crystal on all of the framework anyway, so this is fine for now. Thanks!

jwoertink (approved):

PR #1958, opened 2025-05-22
Avoid intermediate strings

akadusei:
## Purpose
Avoid intermediate strings. _No issue exists. Feel free to reject this PR_

## Description
This should improve performance. [Ref](https://crystal-lang.org/reference/1.16/guides/performance.html#dont-create-intermediate-strings-when-writing-to-an-io)

## Checklist

### DIFF
--- a/src/lucky/base_component.cr
+++ b/src/lucky/base_component.cr
@@ -34,6 +34,6 @@ abstract class Lucky::BaseComponent
     String.build do |io|
       view(io)
       render
-    end.to_s
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


src/lucky/base_component.cr
```diff
@@ -34,6 +34,6 @@ abstract class Lucky::BaseComponent
     String.build do |io|
       view(io)
       render
-    end.to_s
```
jwoertink, line 37: :joy:

src/lucky/tags/specialty_tags.cr
```diff
@@ -87,7 +87,7 @@ module Lucky::SpecialtyTags
       options.each_with_index do |key, value, index|
         attrs << ", " if index > 0
         attrs << Wordsmith::Inflector.dasherize(key.to_s) << "="
```
jwoertink, line 89: Maybe we should make all of the wordsmith methods take an optional IO so we can do `.dasherize(io, key)` :thinking:

jwoertink (approved):
Thanks for finding these!

PR #1970, opened 2025-06-17
 Supporting different javascript runtime options. npm, yarn, and bun.

russ:
## Purpose
Supporting different javascript runtime options. npm, yarn, and bun.
This is in tandem with https://github.com/luckyframework/lucky_cli/pull/889

## Description
I'm not a huge fan of how this is done. I would love to see a different approach.

## Checklist


jwoertink:
So the main question here is, if we just got rid of the browser-sync (which may not even be working, and have been broken for some time), then none of this becomes required. The swap out becomes a lot easier to handle.

We wouldn't lose the auto-reload on the Crystal side because that's still handled by the websocket reloader which is default anyway. Though, I'm not sure if any of the javascript or css files are watched and reloaded on change with that...

jwoertink:
Yeah, I think this PR was only needed to work around browser-sync, but since we're basically ripping that out, then none of this should be needed.

Issue #1948, opened 2025-04-10
Configurable serialization


jwoertink:
Based on some comments from https://github.com/luckyframework/lucky/issues/1930

Right now all serialization assumes JSON. It's the most common, so it's a decent default, but if you need to do some alternate serialization like MsgPack, or XML or something, then it's basically hand-roll and implement your own serializers.

In https://github.com/luckyframework/lucky/pull/1947 the serialization was moved to a module, so now the question is, can we somehow make this configurable?

Could this be a registerable system like:

```crystal
Lucky::Serializable.register(JSON)
Lucky::Serializable.register(YAML)
Lucky::Serializable.register(MsgPack)

module Lucky::Serializable.
  macro register(type)
    def to_{{ type }}(io)
    end
  end
end
```

I have no idea how this would work, but I want to at least open a separate issue for discussion.

jwoertink:
Just for visibility I'm moving [my comment](https://github.com/luckyframework/lucky/pull/1975#issuecomment-3017034452) here.

After thinking about it, adding magic methods doesn't feel like the right direction. What would be easier and more transparent would be to just use some simple requires..

```crystal
require "lucky/serializable/json"

abstract struct BaseSerializer
  include Lucky::Serializable
end
```

```crystal
# lucky/serializable/json.cr
module Lucky::Serializable
  def to_json(io)
    render.to_json
  end
end
```

This pattern makes it easier to extend custom serializations (e.g. msgpack, yaml, etc...) just by including your conversion method in this module, then utilizing your conversion method in the action you need it in.

PR #1975, opened 2025-06-23
Serialization proposal

rmarronnier:
## Purpose

This is a working POC as an alternative solution to #1948 (it's not a configuration per se).

I've added several format as examples and let the md files as documentation (they can be removed, content extracted etc) to show clearly what this would look like.

I find the syntax neat and simple enough. What do you think ?

## Checklist

### DIFF
--- a/SERIALIZATION_CHANGELOG.md
+++ b/SERIALIZATION_CHANGELOG.md
@@ -3,8 +3,8 @@
 ## New Features
 
 ### 🎯 Multi-Format Serialization Support
-- Added support for **YAML**, **MsgPack**, and **CSV** formats alongside existing JSON
-- New rendering methods: `yaml()`, `msgpack()`, `csv()` with full status code support
+- Added support for **YAML**, **MsgPack**, **CSV**, and **XML** formats alongside existing JSON
+- New rendering methods: `yaml()`, `msgpack()`, `csv()`, `xml()` with full status code support
 - Content negotiation via `respond_with()` method based on HTTP Accept headers
 
 ### 🔧 Macro-Based Architecture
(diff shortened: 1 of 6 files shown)


jwoertink:
I keep thinking about this... I wonder if we can actually make it a lot more simple.... I was looking at this recent PR on Crystal https://github.com/crystal-lang/crystal/pull/15939/files and it got me thinking... What if your serialization was just based on a simple require?

```crystal
require "lucky/serializable/json"

abstract struct BaseSerializer
  include Lucky::Serializable
end
```

Then that would just have

```crystal
module Lucky::Serializable
  def to_json(io)
    render.to_json
  end
end
```

Then if you wanted your own custom serialization method, you would just open `Lucky::Serializable` and add your method. 

```crystal
module Lucky::Serializable
  def to_yaml(io)
    render.to_yaml
  end
end

class Data::Index < BrowserAction
  get "/data" do
    serializer = DataSerializer.new(1, 2, 3)
    if yaml?
       yaml(serializer)
     elsif json?
       json(serializer)
     else
       plain_text(serializer.render.to_s)
     end
  end
end
```

Lucky already has the built-in methods used for handling the action output. Serializers just need to know how to convert your `render` to some other output.

rmarronnier:
I agree with you : 
> module Lucky::Serializable
  def to_yaml(io)
    render.to_yaml
  end
end

makes so much sense. What I liked about this PR was especially the `render_format` macro to be able to easily declare a new render format (and the nice cool way to have` foobar @users` in the action).
I don't necessarily want, and won't fight for, to have new built-in formats (like yaml or csv) in `Lucky::Renderable` (it's more like examples of the powerful macro).
Feel free to steal code from this for a new PR to implement #1948

PR #1979, opened 2025-06-26
Add format detection from URL extensions

watzon:
## Purpose
Implements format detection from URL extensions as suggested in #771. This allows routes to automatically detect and handle different formats based on file extensions in the URL (e.g., `/reports/123.csv`).

## Description
This PR adds comprehensive format detection that extracts formats from URL extensions and makes them available to actions. The implementation follows Rails-like conventions where URL formats take precedence over Accept headers.

### Key Features:
- **Automatic Format Detection**: URLs like `/reports/123.csv` automatically detect the CSV format
- **Type-Safe Implementation**: Uses Crystal enums instead of symbols for compile-time type safety
- **Extensible System**: Custom formats can be registered at runtime
- **Backward Compatible**: No breaking changes to existing Lucky applications

### Example Usage:
```crystal
class Reports::Show < BrowserAction
  accepted_formats [:html, :csv, :json], default: :html
  
  get "/reports/:id" do
    # Automatically handles:
    # - /reports/123 (HTML)
    # - /reports/123.csv (CSV) 
    # - /reports/123.json (JSON)
    case clients_desired_format
    when :csv
      send_csv_file(report_data)
    when :json
      json({report: report_data})
    else
      html ShowPage, report: report
    end
  end
end

# Register custom formats
Lucky::FormatRegistry.register("PDF", "pdf", "application/pdf")
```

### Implementation Details:
- Format extraction happens in `RouteHandler` before route matching
- Path extensions are stripped for routing (e.g., `/reports/123.csv` matches `/reports/:id`)
- URL formats take precedence over Accept headers
- Comprehensive edge case testing (750+ tests passing)

## Checklist

### DIFF
--- a/src/lucky/route_handler.cr
+++ b/src/lucky/route_handler.cr
@@ -4,7 +4,21 @@ class Lucky::RouteHandler
   include HTTP::Handler
 
   def call(context : HTTP::Server::Context)
-    handler = Lucky.router.find_action(context.request)
+    original_path = context.request.path
+    
+    # Extract format from URL path and strip it for route matching
+    if url_format = Lucky::MimeType.extract_format_from_path(original_path)
+      context._url_format = url_format
+      # Create a modified request with format-stripped path for route matching
+      path_without_format = original_path.sub(/\.[a-zA-Z0-9]+(?:\?.*)?$/, "")
+      modified_request = context.request.dup
+      modified_request.path = path_without_format
+      lookup_request = modified_request
+    else
+      lookup_request = context.request
+    end
+    
+    handler = Lucky.router.find_action(lookup_request)
     if handler
       Lucky::Log.dexter.debug { {handled_by: handler.payload.to_s} }
       handler.payload.new(context, handler.params).perform_action
--- a/spec/lucky/format_edge_cases_spec.cr
+++ b/spec/lucky/format_edge_cases_spec.cr
@@ -1,0 +1,286 @@
+require "../spec_helper"
+
+include ContextHelper
+
+private class EdgeCaseFormatAction < TestAction
+  accepted_formats [:html, :json, :csv], default: :html
+  
+  get "/test/:id" do
+    plain_text "format: #{clients_desired_format}, id: #{id}"
+  end
+end
+
+describe "Format Detection Edge Cases" do
+  describe "URL parsing edge cases" do
+    it "handles multiple dots in filename correctly" do
+      # Should extract only the final extension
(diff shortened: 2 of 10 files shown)


src/lucky/route_handler.cr
```diff
@@ -4,7 +4,21 @@ class Lucky::RouteHandler
...
 
   def call(context : HTTP::Server::Context)
-    handler = Lucky.router.find_action(context.request)
+    original_path = context.request.path
+
+    # Extract format from URL path and strip it for route matching
+    if url_format = Lucky::MimeType.extract_format_from_path(original_path)
+      context._url_format = url_format
+      # Create a modified request with format-stripped path for route matching
+      path_without_format = original_path.sub(/\.[a-zA-Z0-9]+(?:\?.*)?$/, "")
```
jwoertink, line 13: I think you have it covered already, but I couldn't find in the specs... does this handle if there's a domain in the query string? e.g. `/path/report.csv?affiliate=site.com/path.html`

src/lucky/route_handler.cr
```diff
@@ -4,7 +4,21 @@ class Lucky::RouteHandler
...
+      modified_request = context.request.dup
```
jwoertink, line 14: Is there any considerations of memory here if say you're uploading a 5GB video file in the request? :grimacing:

jwoertink (commented):
Overall I think this is great. Seems like lots of good specs. Just a few small comments to consider.

src/lucky/route_handler.cr
watzon: I'm fairly certain the .dup here only performs a shallow copy, so it should be a matter of a few kilobytes and not the file content. For large uploads, the file should already be on disk as a temporary file, and only the IO reference should be copied.

src/lucky/route_handler.cr
watzon: Good catch. I'm fairly certain that the current regex should match extensions in query strings (like .com in ?affiliate=site.com), but it won't cause issues because we're only recognizing registered formats. So unless the user went in and had a custom format that matched the TLD format, it should be fine. However, I think I can improve the regex a little bit.

src/lucky/route_handler.cr
jwoertink: Ok cool. I asked in the Crystal discord for clarification, but we can roll with this for now and just keep in mind in case this comes up.

jwoertink (approved):
Thanks!

Issue #542, opened 2018-07-17
[Feature Request] Being able to define multiple routes on the same action


jwoertink:
I know I have weird requirements, but this has come up twice now for us. Both times we've found a way around it, but I would imagine at some point that may not be possible.

Imagine you have some routes like this:

```
/affiliates/jeremy
/affiliates/paul
/affiliates/russ

/affiliates/:id
```

The first three are static routes that all go to the same exact action which isn't a `Affiliates::Show` type action. The last one would be your standard `Affiliates::Show` action. 

In our case, we would send the first 3 to the `Affiliates::Show`, and then filter out the value of the params and do different things based on what it looks like. Instead, we'd rather do something like this:

```crystal
class Affiliates::Redirect < BrowserAction
  {% for affiliate in ["jeremy", "paul", "russ"] %}
    get "/affiliates/{{ affiliate.id }}" do
      # do special redirect linking for these affiliates
      # returns 301
    end
  {% end %}
end

class Affiliates::Show < BrowserAction
  get "/affiliates/:id" do
    # render affiliate page
    # returns 200
  end
end
```

Now, I get this would not only be difficult to add in, but would also cause some other issues like where would `Affiliates::Redirect.path` go? That one might be simple to solve with something like `Affiliates::Redirect.path(pattern: "/affiliates/jeremy")` or something along those lines.

Another solution might be something like this:

```crystal
get ["/affiliates/jeremy", "/affiliates/paul", "/affiliates/russ"] do |path|
  # path knows which one was called to get in here
end
```

The current work around would be something like this
```crystal
class Affiliates::Redirects::Jeremy < Affiliates::Redirects::Base
  get "/affiliates/jeremy" do
  end
end
```
However, in our case, we have an affiliate with a name beginning with a number, and we have 130+ of these (a lot of classes).

paulcsmith:
Hey @jwoertink Could you give an example of what the redirect would look like? 

I'm not totally sure I understand the use-case but maybe you could flip this around:

```crystal
get "/affiliates/:id" do
  case id
  when "paul"
  do_something_for_paul
  when "jeremy"
  do_something_for_jeremy
  else
     do_stuff_with_the_id
  end
end
```

Would that work? If not can you explain why? I think it would be cool to allow multiple paths for the same action, but I'd like to fully understand the use-cases before adding it

jwoertink:
That code example could technically work, though in our case, there would be 130+ `when` clauses. We ended up rolling with something like 

```crystal
[...]

PR #1998, opened 2025-10-31
Add route path aliases

wout:
## Purpose
This PR adds the possibility to create aliases for route paths as discussed in #1996.

## Description
I found another approach that is less invasive and requires a lot less changes to the code. It also doesn't affect performance when the feature is not used.

The way it works is to allow passing more than one path to the HTTP verb routing methods in actions:

```cr
class About::Index < BrowserAction
  get "/about", "/:locale/about" do
    # ...
  end
end
```

The first path is still the “main” path, is required, and will set up the action. Any subsequent paths are just aliases and don't trigger the call setup. Every route is added through the `add_route` macro so the aliases have all the same features you'd expect, including prefixes, optional params, etc.

I started using them in Fluck's website and they work great, so far without any side effects.

## Checklist

### DIFF
--- a/spec/lucky/action_spec.cr
+++ b/spec/lucky/action_spec.cr
@@ -206,6 +206,14 @@ class Tests::ActionWithPrefix < TestAction
   end
 end
 
+class Aliases::ActionWithPrefix < TestAction
+  route_prefix "/prefix"
+
+  get "/so_custom_aliased", "/:scope/so_custom_aliased" do
+    plain_text "doesn't matter"
+  end
+end
+
 class Tests::HtmlActionWithCustomContentType < TestAction
   get "/tests/new_action_with_custom_html_content_type" do
     html(Tests::IndexPage)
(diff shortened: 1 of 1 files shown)


jwoertink:
It's funny because it's basically https://github.com/luckyframework/lucky/issues/542 :joy:  

So I guess the main question is, how do you build the URL for one route vs the other? 

Something like this?

```crystal
About::Index.url #=> site.com/about
About::Index.with(locale: "nl").url #=> site.com/nl/about
```

wout:
> It's funny because it's basically #542 😂

So funny! 😄 I've never seen that issue, probably because it was before I discovered Lucky.

> So I guess the main question is, how do you build the URL for one route vs the other?
> 
> Something like this?
> 
> ```crystal
> About::Index.url #=> site.com/about
> About::Index.with(locale: "nl").url #=> site.com/nl/about
> ```

Precisely. That just works because it's using the `add_route` macro.

There is one thing I'd like to add to simplify routing. Currently in the example above the `locale` param can't be `nil` to route to the alias without the parameter in it:

```cr
About::Index.with(locale: nil).url #=> site.com/about
``` 

In case of localization, having this option would make it super easy to handle routing for the "main" locale (e.g. without `:locale`) and the additional locales where `:locale` is required. But that's for another PR. First I want to give this a spin to see how it works in reality.

PR #2002, opened 2025-11-07
Reorganize files to make HTML optional.

### LINKED ISSUE #1783: Investigate if Actions are the cause of long compilcation times
For our app we compile the main app binary, and a separate mosquito worker binary. Both binaries require the same app code with the exception that the worker binary doesn't require any actions. 

This is the difference in compilation time
image

I think to further investigate this, we just need to benchmark compiling an app, then compiling a separate app that doesn't use Actions, and see if there's truly a difference, or is it that we're doing something with our app?

My app has 28 action classes... Though it's a decent size, that's not really that many.

jwoertink:
## Purpose
Fixes #1783

## Description
No clue if this is a good idea or not :sweat_smile: This would be considered a breaking change, but in theory, should be easy to update existing apps....

The idea here is that when you have an API only Lucky app, and have no need for any sort of HTML pages/components, then by removing these files, you should see a (small) performance increase in compilation time.

In my very limited testing, I saw compilation time drop by several seconds. Granted, my app takes well over 2 minutes to compile, so it wasn't much... but I figure in the case of smaller apps, this may be the difference between getting distracted when writing specs, and being productive.

If all works correctly, you would just need to add this for existing apps that need HTML

```crystal
require "lucky"
require "lucky/html"
```

## Checklist


jwoertink:
Testing on 2 fresh apps. This one is an API only app where I've added the `require "lucky/html"` in and taken it out... but can you guess which one is which?

```
~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 5s
❯ time crystal build src/start_server.cr

real	0m4.518s
user	0m4.124s
sys	0m1.057s

~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build src/start_server.cr

real	0m4.516s
user	0m4.146s
sys	0m1.026s

~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build src/start_server.cr

real	0m4.851s
user	0m7.041s
sys	0m1.298s

~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build src/start_server.cr

real	0m4.360s
user	0m4.163s
sys	0m0.866s

~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build src/start_server.cr

real	0m4.603s
user	0m4.189s
sys	0m1.081s

~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build src/start_server.cr

real	0m4.800s
user	0m7.051s
sys	0m1.237s
```

Here's a fresh Web app. Both of these are pulling in this branch.

```
~/Sites/zzz_garbage is 📦 v0.1.0 via  v24.11.0 via 🔮 v1.18.2
❯ time crystal build src/start_server.cr

real	0m5.328s
user	0m7.195s
sys	0m1.309s

~/Sites/zzz_garbage is 📦 v0.1.0 via  v24.11.0 via 🔮 v1.18.2 took 5s
❯ time crystal build src/start_server.cr

real	0m5.099s
user	0m4.709s
sys	0m1.153s

~/Sites/zzz_garbage is 📦 v0.1.0 via  v24.11.0 via 🔮 v1.18.2 took 5s
❯ time crystal build src/start_server.cr

real	0m5.139s
user	0m4.750s
sys	0m1.176s

~/Sites/zzz_garbage is 📦 v0.1.0 via  v24.11.0 via 🔮 v1.18.2 took 5s
❯ time crystal build src/start_server.cr

real	0m5.064s
user	0m4.727s
sys	0m1.111s
```

Really, it feels almost negligible... at least in development... lastly, just for funsies, here's a release build comparison.

full app with HTML
```
~/Sites/zzz_garbage is 📦 v0.1.0 via  v24.11.0 via 🔮 v1.18.2 took 5s
❯ time crystal build --release src/start_server.cr

real	0m44.987s
user	0m42.913s
sys	0m2.043s
```

api app without HTML
```
~/Sites/zzz_garbage_api is 📦 v0.1.0 via 🔮 v1.18.2 took 4s
❯ time crystal build --release src/start_server.cr

real	0m40.773s
user	0m38.912s
sys	0m1.727s
```

src/lucky/paginator/components/bulma_nav.cr
```diff
@@ -1,9 +1,12 @@
+require "../../html/base_component"
+
 # Pagination component using Bulma Pagination styles
 #
 # https://bulma.io/documentation/components/pagination/
 class Lucky::Paginator::BulmaNav < Lucky::BaseComponent
   needs pages : Lucky::Paginator
 
+  @[Deprecated("Consider using Lucky::Paginator::SimpleNav or writing a custom component")]
```
wout: Is it an idea to put those front-end framework-specific components in a separate shard? Or maybe just document them?

jwoertink: I don't think so. I think just documenting them would be much better. Like recipes if you need. Then adding them into your project is easy and don't have to worry about outdated CSS framework versions.

PR #2015, opened 2026-02-15
Use bun for all front end concerns

wout:
## Changes
This is just the current state of the Bun implementation as a first draft. This works locally in development in the Fluck website app, and all tests pass as expected.

Changes needed in the Lucky app itself are:

**Procfile.dev**
```yaml
assets: bun run dev
```

**package.json**
```json
  "scripts": {
    "build": "bun lib/lucky/src/bun/bake.js",
    "dev": "bun run build --dev",
    "prod": "bun run build --prod"
  },
```

**src/components/shared/layout_head.cr**
```crystal
bun_reload_connect_tag
```

**Other**
- Run `bun run prod` for deployment 
- Add the shared `config/bun.json` file (optional, all fallbacks are in place)

## Questions
I'd like to add more tests but I'm a bit stuck. More specifically I'm a bit confused by the `Lucky::AssetBuilder::Mix` and `Lucky::AssetBuilder::Vite` classes. They only ever seem to be used in the test suite. What's their purpose? Are these preparations for the pluggable asset system?

What confuses me is that they are part of `src` but only used in `spec`. Of course I can add a `Bun` variant with a test suite, but I thought I'd ask first.

### DIFF
--- a/src/run_macros/asset_manifest_builder.cr
+++ b/src/run_macros/asset_manifest_builder.cr
@@ -27,6 +27,8 @@ struct AssetManifestBuilder
     @retry_after = ENV["LUCKY_ASSET_MANIFEST_RETRY_DELAY"]?.try(&.to_f) || 0.25
   end
 
+  # Tries to build a manifest from the chosen bundler and retries several times
+  # when it fails.
   def build_with_retry
     retry_or_raise_error unless File.exists?(@manifest_path)
 
@@ -37,6 +39,7 @@ struct AssetManifestBuilder
     end
   end
 
+  # Tracks retries and raises if maximum allowed retries are exceeded.
   private def retry_or_raise_error
     raise_missing_manifest_error unless retries < @max_retries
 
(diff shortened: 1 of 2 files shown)


wout:
Thanks! Bun feels really snappy and I love that it doesn't require any dependencies. Not even for the test suite.

A pluggable asset system is a great idea in itself I think, and adding the Bun counterparts wouldn't be a lot of work. But I'm not completely clear when it would be invoked. Would that be at compile time like the current `AssetManifestBuilder`? Because it would be great to modularise that part.

The only thing is that if it's kept around, the Vite implementation should be equally usable as the Mix and Bun versions. We could harvest the good parts from the LuckyVite shard which works, but is it worth the effort? It's not that any of use has the spare time to put in the effort into something that may not even be used at all 😄.

**UPDATE**:
For the flags on the `load_manifest` macro, we could rework that so it's less messy:
-  `load_manifest(use: :bun)` (default)
-  `load_manifest(use: :vite)`
-  `load_manifest(use: :mix)`

And then the frontend would require a bit of manual configuration which is inevitable I think.

src/run_macros/asset_manifest_builder.cr
```diff
@@ -0,0 +1,169 @@
...
+                  yarn run mix watch   # development build with watcher
+                  yarn run mix --production  # production build
+
+                ERROR
+              in .vite?
+                <<-ERROR
+                #{"Manifest not found:".colorize(:red)} #{@manifest_path}
+
+                #{"Make sure you have compiled your assets:".colorize(:yellow)}
+                  npx vite        # start development server
```
wout: I'm not sure about the build commands here. Should they still use yarn like with mix?

PR #1982, opened 2025-06-29
[Feature] Pluggable Asset Builder System

watzon:
This PR implements the framework portion of issue #1980, introducing a pluggable asset builder system that allows Lucky to work with different asset compilation tools while maintaining backwards compatibility with Laravel Mix.

### Asset Builder System (`src/lucky/asset_builder/`)

1. **Base Architecture**
   - Created abstract `Base` class defining the interface for asset builders
   - Defines `manifest_path`, `parse_manifest`, and `load_manifest` methods
   - Includes `normalize_key` helper for consistent asset path handling

2. **Mix Implementation** (`mix.cr`)
   - Implements Laravel Mix support for backwards compatibility
   - Simple key-value manifest parsing
   - Default manifest path: `./public/mix-manifest.json`

3. **Vite Implementation** (`vite.cr`)
   - Full support for Vite manifest format
   - Handles both production manifests and dev manifests (from vite-plugin-dev-manifest)
   - Intelligently detects manifest type based on structure
   - Filters out internal chunks (underscore-prefixed entries)
   - Default manifest path: `./public/.vite/manifest.json`

### Server Configuration (`src/lucky/server.cr`)
- Added `asset_build_system` setting with Mix as default
- Required all asset builder classes
- Maintains backwards compatibility - no configuration needed for existing projects

### Asset Helpers (`src/lucky/asset_helpers.cr`)
- Added experimental `load_manifest(manifest_file, use_vite)` overload
- Updated `load_manifest_from_build_system` macro to use correct Vite manifest path
- No changes to existing `asset()` or `dynamic_asset()` methods

### Macro Updates (`src/run_macros/generate_asset_helpers.cr`)
- Enhanced to handle both Vite production and dev manifest formats
- Detects dev manifest by presence of `url` and `inputs` properties
- Maintains compatibility with existing Mix manifest handling

### Vite Dev Manifest Support
The Vite implementation can parse dev manifests generated by vite-plugin-dev-manifest:
```json
{
  "url": "http://localhost:3001/",
  "inputs": {
    "app": "src/js/app.js",
    "styles": "src/css/app.scss"
  }
}
```

### Asset Path Normalization
All builders normalize asset paths consistently:
- Remove leading slashes
- Remove `assets/` prefix
- Strip `src/` prefix for Vite entries

## Testing

Comprehensive test coverage added:
- Unit tests for Mix and Vite builders
- Server integration tests
- Tests for both production and dev manifest formats
- Error handling and edge cases

## Backwards Compatibility

[...]

PR #2017, opened 2026-02-25
Remove the pluggable asset system

jwoertink:
## Purpose
Reverting most of #1982

## Description
We originally wanted some built-in way to tell your Lucky app which asset build system you were going to use. Then based on that, we could internally decide how to build things out.

However, once that was merged in, and we started working out Vite, things got super messy and complicated and broken real fast.

After all that, @wout came up with this absolutely brilliant idea to use `bun` as the default (https://github.com/luckyframework/lucky/pull/2015). I wanted to do this anyway, and it ended up being a much better solution anyway.

So now I'm reverting the bulk of #1982 for the time being just so we can move on and do a release without this stuff getting in the way. Eventually we can take a second stab at a pluggable system.


robacarp (approved):
Pluggable systems [are an incredible lift](https://github.com/mosquito-cr/mosquito/issues/67), ask me [how I know](https://github.com/amberframework/granite). It's far more sustainable to support one path deeply, and then have an escape hatch for when folks want to jump off the beaten path.

PR #2024, opened 2026-04-02
Fixing spelling mistakes.

jwoertink:
Using Ameba to catch some typos. A few are intentional in the specs, so I added those to the ignore file.


jwoertink:
The nightly CI failure is unrelated and will be fixed by Crystal soon, but the windows CI failure (which is also unrelated) seems to be a flaky spec. I've seen it fail randomly before.

PR #2032, opened 2026-04-07
Make sure css is always reloaded

wout:
## Purpose
Makes sure CSS is always fresh when using libraries like Turbo, HTMX, or AlpineAjax.

## Description
While working on CSS in a project, I noticed that old CSS would be shown after a hot reload when moving away from the page you're working on using Turbo. That is because the CSS hot reload adds a timestamp to get a fresh version in real time, but when moving away using Turbo, the timestamp is gone, and when navigating with Turbo [`Lucky::DevAssetCacheHandler`](https://github.com/luckyframework/lucky/blob/1399ce07f682aede964e0bfa91111024392a3c12/src/lucky/dev_asset_cache_handler.cr#L4) can't do its job. So we also need to add a timestamp to CSS files on the Lucky side in development.

Come to think of it, is lucky_cli adding the header for new apps as well?

```crystal
  def middleware : Array(HTTP::Handler)
    [
      Lucky::RequestIdHandler.new,
      Lucky::ForceSSLHandler.new,
      Lucky::HttpMethodOverrideHandler.new,
      Lucky::LogHandler.new,
      Lucky::ErrorHandler.new(action: Errors::Show),
      Raven::Lucky::ErrorHandler.new,
      Lucky::RemoteIpHandler.new,
      Lucky::RouteHandler.new,
      Lucky::DevAssetCacheHandler.new(enabled: LuckyEnv.development?), # <= this one
      Lucky::StaticCompressionHandler.new("./public", file_ext: "gz", content_encoding: "gzip"),
      Lucky::StaticFileHandler.new("./public", fallthrough: false, directory_listing: false),
      Lucky::RouteNotFoundHandler.new,
    ] of HTTP::Handler
  end
```

And I'm not sure about the `enabled: LuckyEnv.development?` part in the handler. Is that a good pattern? Or should we conditionally add it to the handlers list instead?

Also, I've added lucky_env as a dev dependency. Not sure it that's okay or not, but I needed it for testing the `bun_connect_reload_tag`. Or we could take out the guard and move it to layout head like lucky's reload tag.

## Checklist

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -53,6 +53,9 @@ dependencies:
     version: ~> 0.2.0
 
 development_dependencies:
+  lucky_env:
+    github: luckyframework/lucky_env
+    version: ~> 0.3.0
   ameba:
     github: crystal-ameba/ameba
     version: ~> 1.6.4
--- a/spec/lucky/specialty_tags_spec.cr
+++ b/spec/lucky/specialty_tags_spec.cr
@@ -26,14 +26,32 @@ describe Lucky::SpecialtyTags do
     HTML
   end
 
-  it "cache-busts local css links in development" do
+  it "cache-busts non-fingerprinted local css links" do
     html = view(&.css_link("/assets/css/app.css"))
 
     html.should contain "bust="
   end
 
+  it "does not cache-bust fingerprinted local css links" do
+    html = view(&.css_link("/assets/css/app-5e6f7a8b.css"))
+
+    html.should_not contain "bust="
+  end
+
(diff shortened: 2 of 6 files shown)


jwoertink (approved):

jwoertink:
> And I'm not sure about the enabled: LuckyEnv.development? part in the handler. Is that a good pattern?

I think so. I have several different middleware stacks in my app for different things, and this is what I used as opposed to trying to complicate building an array dynamically.

PR #2035, opened 2026-04-18
Add bun exclusion clauses and more flag options

wout:
## Purpose
We've been using this setup for a while in Rails and a few suggestions came up from our front-end dev. This PR includes those tweaks.

## Description
There are three improvements.

### Exclusions in glob patterns
Allows exclusion of sub-dirs or files from a CSS import with glob using the `not` clause:
```css
@import './components/**/*.css' not './components/admin/**';
```
Or with multiple:
```css
@import './components/**/*.css'
  not './components/admin/**'
  not './components/internal/**';
```
This also works for JS globs:
```js
import components from 'glob:./components/**/*.js not ./components/admin/**'
```

### More granular control for build flags
The only ~~two~~ three flags we had were:
- no flag: just build, no fingerprinting, no minification and inline sourcemap
- `--dev`: build and start dev server with watcher
- `--prod`: build with minification and fingerprinting, no soucemap
- `--debug`: more verbose WebSocket output

There were cases where we wanted to have fingerprinting, but no minification. And in hindsight, not having sourcemap in production was not a good idea. So we've added the following:
- `--fingerprint`: build without minification but with fingerprints
- `--minify`: build without fingerprints but with minification
- `--sourcemap=[inline, linked, external, none]`: this now hinges on the `--dev` flag where it’s `inline` and `linked` for builds

The `--prod` flag is now just a shortcut for `--fingerprint --minify`, so this is all backwards compatible.

### Retain scroll position on full reloads
CSS is hot reloaded, so changes are instant and don't affect the scroll position. With JS and other assets, a full page reload was triggered which inevitably reset the scroll position to `0`. Now the scroll position is stored in `sessionStorage` and restored after the reload.

This made me think, should we add that as well to Lucky's live reload tag? It's a really nice thing to have.

## Checklist

### DIFF
--- a/spec/bun/lucky.test.js
+++ b/spec/bun/lucky.test.js
@@ -1,6 +1,6 @@
 import {describe, test, expect, beforeEach, afterAll} from 'bun:test'
 import {mkdirSync, writeFileSync, rmSync, existsSync, readFileSync} from 'fs'
-import {join} from 'path'
+import {join, basename} from 'path'
 import LuckyBun from '../../src/bun/lucky.js'
 
 const TEST_DIR = join(process.cwd(), '.test-tmp')
@@ -14,6 +14,9 @@ beforeEach(() => {
   LuckyBun.debug = false
   LuckyBun.prod = false
   LuckyBun.dev = false
+  LuckyBun.fingerprint = false
+  LuckyBun.minify = false
+  LuckyBun.sourcemap = null
   LuckyBun.root = TEST_DIR
 })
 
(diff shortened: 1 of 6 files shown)


jwoertink:
I've never seen this `not` pattern before. Does that exist in anything else, or is that an exclusive feature to this bun setup? I tried looking it up, but not seeing anything. It''s definitely an interesting concept. I need to look over this code a bit more to really understand what's going on.

wout:
It's not something that existed before. It's specific to the `cssGlobs.js` and `jsGlobs.js` plugins. I found this easier to reason about than some complicated exclusion patterns. You can see what's going on in a glance.

wout:
Not unrelated, I created a plugins repo for this setup: https://codeberg.org/fluck/bun_bun_bundle-plugins

If you feel the `aliases`, `cssGlobs`, and `jsGlobs` plugins are too opinionated, we can always keep them in there. Then users can choose for themselves what to use and what not. Lucky provides a solid base, users extend where they need it.

jwoertink:
Sweet. I think it's neat, and definitely adds a new level of flexibility. I'm open to trying it out.

jwoertink (approved):

PR #2036, opened 2026-04-18
Add .subdomain chainable method for route helpers

### LINKED ISSUE #2033: Use a method for `subdomain` similar to `format` for routes
In https://github.com/luckyframework/lucky/pull/1974 a new arg to `Action.with` was added where you can pass in `subdomain: "xyz"` and then it will generate route helpers with that subdomain. 

However, in https://github.com/luckyframework/lucky/pull/2022 @wout added this `format()` method that you can chain to essentially slap something on the end of a URL, and I kind of like that better. In that case, it allows you to still use the word "format" in your query params or path params without messing anything up. The word `subdomain` is probably less likely to end up as a query or path param, but it's probably also a non-0% chance that it could....

So I'm proposing that we make this change so instead of `.with(subdomain: "admin")` it becomes a method like `.subdomain("admin")`. Then you could do something like

```crystal
# https://reporting.mysite.com/admin/reporting/4.csv?query=money
Admin::Reporting.with(id: 4, query: "money").format(:csv).subdomain("reporting").url
```

Since that `subdomain` param was never released, it would only be a "Breaking" change to anyone that's been just pulling against main directly 😬

grepsedawk:
## Purpose

Closes #2033.

Replaces the `subdomain:` keyword argument on `.with()` (added in #1974) with a chainable `.subdomain("admin")` method, matching the builder style introduced by the `.as_*` format methods in #2022.

The motivation from the issue: `subdomain` as a kwarg on `.with()` means no one can ever model a resource with a `subdomain` query or path param without collision. Pulling route modifiers out of the keyword bag avoids that, and reads more naturally alongside `.as_rss`/`.as_json`.

## Description

- Adds `Action.subdomain("admin")` and `Lucky::RouteHelper#subdomain("admin")` builder methods.
- Removes the `subdomain:` keyword from `self.with`, `self.route`, `self.path_without_query_params`, and `self.url_without_query_params`.
- `FormatBuilder` now carries subdomain state, so `.as_*` and `.subdomain(...)` compose in any order.

All four of these produce the same URL:

```crystal
Admin::Reporting.with(id: 4).as_json.subdomain("reporting").url
Admin::Reporting.with(id: 4).subdomain("reporting").as_json.url
Admin::Reporting.as_json.subdomain("reporting").with(id: 4).url
Admin::Reporting.subdomain("reporting").as_json.with(id: 4).url
```

### Breaking change

The `subdomain:` keyword argument is removed. Since it was merged in #1974 after v1.4.0 and has not appeared in a released version, this only affects users tracking `main`.

## Checklist

### DIFF
--- a/spec/lucky/route_helper_spec.cr
+++ b/spec/lucky/route_helper_spec.cr
@@ -16,6 +16,14 @@ class TestParamAction < TestAction
   end
 end
 
+class TestSubdomainFormatAction < TestAction
+  accepted_formats [:html, :rss, :json], default: :html
+
+  get "/subdomain_reports/:id" do
+    plain_text "report"
+  end
+end
+
 describe Lucky::RouteHelper do
   describe "url" do
     it "returns the host + path" do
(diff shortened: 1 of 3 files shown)


jwoertink (approved):
nice!

PR #2037, opened 2026-04-19
Add bun watcher patch

wout:
## Purpose
This adds a `watchExtensions` to allow watching file types other than the target file type.

## Description
Tailwind CSS setups require a rebuild of the CSS whenever a template file (Crystal, HTML, ECR, etc.) changes because Tailwind scans those templates for class candidates. Before the watcher only rebuilds CSS when a .css file itself changes, so template edits leave the stylesheet stale until the next CSS change.

This PR adds a `watchExtensions` option in config/bun.json that routes extra extensions into the CSS or JS rebuild path:                                                                                            

```json                                                                                                                                                                                                              
  {
    "watchDirs": ["src"],
    "watchExtensions": {
      "css": ["cr"],
      "js": []
    }
  }
```                                

With this config option a change to any .cr file under `watchDirs` triggers `buildCSS()`, so Tailwind re-scans the templates and the stylesheet updates.
     
## Checklist

### DIFF
--- a/spec/bun/lucky.test.js
+++ b/spec/bun/lucky.test.js
@@ -289,7 +289,9 @@ describe('buildAssets', () => {
     await setupProject({'src/js/app.js': 'console.log("bare")'})
     await LuckyBun.buildJS()
 
-    expect(existsSync(join(TEST_DIR, 'public/assets/js/app.js.map'))).toBe(false)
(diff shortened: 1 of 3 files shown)


wout:
Here's the Tailwind plugin btw: https://codeberg.org/fluck/bun_bun_bundle-plugins/src/branch/main/tailwind

jwoertink (approved):

PR #2038, opened 2026-04-20
Fix reload after rebuilds of the target extension

wout:
## Purpose
Fixed unnecessary reloads for watched files that are not the target build type.

## Description
A reload should not be triggered when watching Crystal files to do a CSS rebuild. Only changes to CSS files should trigger a CSS reload. This is precisely what we touched on before, Bun is responsible for reloads after front-end asset changes, Lucky is responsible for the backend. I came across this when working on the Bun setup with Tailwind for the Lucky website.

Sorry for all those tiny PR's BTW 😬. If this bothers you we could extract the JS part into a separate repo and register is as a git submodule.

## Checklist

### DIFF
--- a/src/bun/lucky.js
+++ b/src/bun/lucky.js
@@ -276,9 +276,11 @@ export default {
   },
 
   async watch() {
+    const cssBase = ['css']
+    const jsBase = ['js', 'ts', 'jsx', 'tsx']
     const extras = this.config.watchExtensions || {}
-    const cssExts = ['css', ...(extras.css || [])]
-    const jsExts = ['js', 'ts', 'jsx', 'tsx', ...(extras.js || [])]
+    const cssExts = [...cssBase, ...(extras.css || [])]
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
No worries. Since I haven't started the release process yet, it's fine. Might as well get it correct when it goes out! 

I did wonder about the volatility of front-end though... Having it locked into a whole Lucky release could make that hard... Though, on the other hand, if it's just patches and bug fixes, I'm ok doing patch release on the Lucky shard itself which doesn't require an entire ecosystem update. I say we just leave as-is for now and see how that goes before we get too wild with it.

Thanks!

wout:
Okay, great.

On volatility: I think we're mostly done. The few recent changes were needed to make the TailwindCSS setup work in the website repo. That's all done now. It works exactly as before again.

If anything comes up after the 1.5.0 release, I'll group the changes and create a PR when it's worthy of a patch release.

PR #2044, opened 2026-05-09
Override pipes

### LINKED ISSUE #2043: Re-adding skipped pipes
The current pipes implementation does a diff of added and skipped pipes, then call what's left. This means that after a pipe is skipped, re-adding it won't call the pipe:

```crystal
abstract class ApiAction < Lucky::Action
  # ...
  after :say_hello

  skip :say_hello # <= Somewhere along the line (could be in a module)

  def say_hello
    puts "Hello, World!"
    continue
  end
  # ...
end

class Posts::Index < ApiAction
  # ...
  before :say_hello # <= This won't work because the pipe is previously skipped
  # ...
end
```

There are pipes in [Shield](https://github.com/grottopress/shield) that are enabled by default in the base action. This means an app has to skip them in subclasses where it does not need them. This is a sane default for security, but requires *thinking in the reverse direction*.  I recently added features that makes that default unnecessary, and I cannot remove this default because *backwards compatibility*. So I imagine someone could disable this default by skipping them in their own base action, then re-add them in subclasses where needed. Currently, this is not possible, as skipping them disables them entirely.

Is the current implementation a design decision, or is a PR that allows re-adding a skipped pipe welcome?

akadusei:
## Purpose
Fixes #2043 

## Description
Allow re-adding skipped pipes. This also allows re-skipping a re-added pipe, and so on. Ultimately, the last rule wins.

If you consider the macro constants to be part of the public API, then this is a breaking change. They are marked as `:nodoc:` so I guess they are not public.

## Checklist

### DIFF
--- a/spec/lucky/action_pipes_spec.cr
+++ b/spec/lucky/action_pipes_spec.cr
@@ -38,6 +38,16 @@ class Pipes::Skipped < InheritablePipes
   end
 end
 
+class Pipes::Readded < Pipes::Skipped
+  before set_before_cookie
+  after overwrite_after_cookie
+
+  get "/readded_pipes" do
+    cookies.set("after", "This should be overwritten by the after pipe")
+    plain_text "Body"
+  end
+end
+
 class Pipes::Index < InheritablePipes
   before set_second_before_cookie
   after set_second_after_cookie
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Nice! Looks good.

PR #2045, opened 2026-06-20
Ameba 1.7

Sija:
## Purpose

This PR updates the Ameba linter to the latest (unreleased yet) v1.7.0-dev.


.github/workflows/ci.yml
```diff
@@ -4,25 +4,9 @@ on:
...
-    continue-on-error: false
-    steps:
-      - uses: actions/checkout@v6
-      - uses: crystal-lang/install-crystal@v1
-        with:
-          crystal: latest
-      - name: Install shards
-        run: shards install
-      - name: Format
-        run: crystal tool format --check
```
jwoertink, line 23: oh, does Ameba do the format check for us?

spec/lucky/form_helpers_spec.cr
```diff
@@ -73,34 +73,34 @@ describe Lucky::FormHelpers do
   it "renders a form tag" do
     without_csrf_protection do
       view(&.inferred_put_form).should contain <<-HTML
-      <form action="/form_helpers/fake_id" method="post"><input type="hidden" name="_method" value="put">foo</form>
-      HTML
+        <form action="/form_helpers/fake_id" method="post"><input type="hidden" name="_method" value="put">foo</form>
+        HTML
```
jwoertink, line 77: Indenting these heredocs is an interesting change... I would have thought having it indented like this would create extra space in the string 🤔

jwoertink (commented):
Thanks for putting this together! I guess Ameba takes care of the format check now and fails if the code isn't formatted fully, right? If so, then I think this will be ready to merge.

.github/workflows/ci.yml
Sija: Yep, `Lint/Formatting` rule takes care of that.

spec/lucky/form_helpers_spec.cr
Sija: Nope, heredocs infer the base indentation level from the ending marker position.

This is an example of a previously (erroneously, I reckon) indented string: https://github.com/luckyframework/lucky/pull/2045/changes#diff-e72a07d2e38a4bdf8b95d11bd03495b4e2619931642590249ee5853bb5fb0271R74-R82

Sija:
> Thanks for putting this together! I guess Ameba takes care of the format check now and fails if the code isn't formatted fully, right?

That's exactly the case 👍

jwoertink (approved):

PR #2050, opened 2026-08-10
Fix Cycle text helper thread safety

### LINKED ISSUE #2048: The Cycle text helper isn't thread-safe
This this class level hash that gets mutated to handle cycling between values. Generally might be used for something like styling every other row in a data table.

https://github.com/luckyframework/lucky/blob/8c0e50f4c051e6a10b6147fb187aa69e47668154/src/lucky/page_helpers/text_helpers.cr#L3

Since this gets shared across requests, there could be a race condition that would show up easily under multithreadding. This should just be moved into an instance level for the page you're on. And while we're at it, we can move that whole class to its own file.

Related: https://github.com/luckyframework/lucky/issues/1806

w3lld1:
## Purpose

Fixes #2048.

## Description

I moved cycle state from a module-level class variable to an instance variable on each page. This keeps `cycle`, `current_cycle`, and reset operations isolated between concurrently rendered pages.

I also moved `Lucky::TextHelpers::Cycle` into its own file and updated the specs so each example exercises one page instance. The new regression spec verifies that starting a cycle on one page does not advance the same named cycle on another page.

## Verification

- `crystal spec spec/lucky/text_helpers/cycle_spec.cr` — 14 examples, 0 failures
- `crystal spec` — 818 examples, 0 failures
- `crystal tool format --check src spec`
- Ameba — 230 files inspected, 0 failures
- `git diff --check`

The Docker wrapper was not run because the local Docker socket is unavailable; the equivalent Crystal suite and formatting checks above were run directly.

## Checklist

* [x] - Inline documentation has been added and/or updated where needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Changes in 1.5.0
 
+- Fixed: isolate text helper cycles between page instances. [#2048](https://github.com/luckyframework/lucky/issues/2048)
 - Updated: HTML Components will fail at compile-time for unused arguments. [#1978](https://github.com/luckyframework/lucky/pull/1978)
 - Added: format detection from URL extensions in actions. [#1979](https://github.com/luckyframework/lucky/pull/1979)
 - Added: Allow customizing the manifest retry timeout. [#1977](https://github.com/luckyframework/lucky/pull/1977)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ### Changes in 1.5.0
 
-- Fixed: isolate text helper cycles between page instances. [#2048](https://github.com/luckyframework/lucky/issues/2048)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### Changes in 1.5.0
 
+- Fixed: isolate text helper cycles between page instances. [#2048](https://github.com/luckyframework/lucky/issues/2048)
```
jwoertink: No need to add this. I'll add the changelog items when I'm ready to do a release.

jwoertink (changes requested):
Nice! Thanks for taking this on. It looks good. Just remove the line in the CHANGELOG and it should be good.

w3lld1:
Thanks — removed the changelog entry in `8318bc5`. `git diff --check` passes.

jwoertink (approved):

